PR #2, opened 2019-05-28
Add explicit list of use cases

vsekhar:
For discussion.


csharrison:
These are good general use-cases but I think it's too high level. Maybe it would make sense to make a separate document with more specific use-cases spelled out. In particular, we aren't just concerned with measuring campaign effectiveness, but also with e.g. training ML models. We aren't just concerned about VTC but also about measuring counterfactuals, etc.

We can also pull from https://github.com/w3c/web-advertising/blob/master/OnlineAdvertisingCapabilitiesHitList.md for inspiration.

vsekhar:
Definitely not complete. I'll take a look at the larger list. Updates to come.

The request from analytics folks was to make clear which use cases this proposal was meant to cover. I'm following up with Onil this afternoon.

Desiree6 (approved):

apasel422:
This has been open for over 4 years. Feel free to reopen this upon fixing conflicts and addressing the comments, if you're still interested in merging this.

PR #19, opened 2020-01-16
Clarify reporting windows

johnivdel:
Clarify in the reporting windows section that impression expiry can be used in place of the 7 day window.


csharrison:
Yeah this LGTM

PR #20, opened 2020-01-22
readme: fix iframe allow attribute

suzuken:
In the section of Publisher Controls for Impression Declaration, it seems `https://ad-tech.com` is missing in iframe.


csharrison:
Actually the current explainer just has a single bit flag allowing / disallowing impression registration at all. We don't need to have specific permission for ad-tech.com, just for the advertiser.com iframe document.

PR #28, opened 2020-03-09
Minor tweaks

samdutton:
Replaced curly quotes in code and removed unnecessary trailing slash in ``.

The `allow="conversion-measurement" ...` values look wrong, but I'm not sure what the correct syntax is.


apasel422:
Closing since this is quite out of date now.

Issue #24, opened 2020-03-03
Impossibility to model conversion flow


nchrys:
The proposal mentions the possibility to model the whole conversion flow (checkout, purchase, etc.). I would like to point that with the current level of noise proposed (5%) then if we use the 3bits metadata to encode steps in this flow it becomes actually impossible to learn models for each step since the level of noise (5%/8) will be higher than the typical conversion rate that is observed on advertiser sites. Actually it seems any level of noise would render this task very difficult or impossible. The only option with noise is that conversions would be registered only at the final step of the flow, and the 3bits metadata would only encode the value of conversion (such as for example a bucket of sales amount), which would severely limit the utility of this proposal.

johnivdel:
The 5% noise is applied only to the conversion metadata, not the conversion event itself. The API provides the ability to measure overall conversion rate with perfect accuracy. For example, if you have 1000 impressions and 20 conversions (conversion rate of 2%), then on average 1 of the 20 conversion reports will have randomized metadata (5% noise on metadata).

However, as you noted, if the different steps have significantly different conversion rates between them, then you will run into issues with utility with step-level modeling. E.g. 1000 impressions with 999 low value conversions, and 1 high value conversion. In reports, you are expected to see ~956 low value, and ~7(1 + .05/8*1000) high value conversions. 

Because the true conversion rate is known, with enough reports you can subtract the expected level of noise. For every 1000 conversions, you expect ~6 high value conversions solely due to noise.

Note that in [this speculative noise section](https://github.com/csharrison/conversion-measurement-api#speculative-adding-noise-to-the-conversion-event-itself), we do describe a possible extension to add noise to the conversion rate itself and note the potential utility hit.

nchrys:
I think we actually agree and that is my point: if you model different step in the flow then rare steps will be drowned in noise.
I was pointing out that the proposal states that it is possible to model the whole conversion flow, when in practice it is not really the case because of the rarity of the actual purchase step.
Consequently the only reasonable way to use this API is to register as conversion the actual purchase step, and use the 3bits metadata to qualify this purchase.

csharrison:
Even with rare events that occur with frequency <5% you can recover the true frequency with low error if you have enough samples via the noise subtraction mechanism John alluded to.

nchrys:
I rechecked the math, and there is indeed an unbiased estimator. I suspect the variance will be very high, which will lead to some difficulty for the models, but at least it is possible, thanks.

[...]

PR #31, opened 2020-03-12
Mention randomized response in the metadata section

csharrison:
Mention that the original frequencies can be recovered in aggregate (followup from issue #24 )


csharrison:
@johnivdel PTAL do you think there is a better reference here? I couldn't find anything simple.

johnivdel:
I think the wiki article is probably the best starting point, so lgtm

PR #33, opened 2020-03-23
Scaffolding spec with high level sections and algorithms

csharrison:
This PR creates the overall spec structures / sections and fills in a few structures and algorithms.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1,16 +1,130 @@
 <pre class='metadata'>
-Title: Your Spec Title
-Shortname: your-spec
+Title: Conversion Measurement
+Shortname: conversion-measurement
 Level: 1
-Status: ED
-Group: WGNAMEORWHATEVER
-URL: http://example.com/url-this-spec-will-live-at
-Editor: Your Name, Your Company http://example.com/your-company, your-email@example.com, http://example.com/your-personal-website
-Abstract: A short description of your spec, one or two sentences.
+Status: w3c/ED
+Group: TODO
+URL: http://csharrison.github.io/conversion-measurement-api
+Editor: Charlie Harrison, Google Inc. https://google.com, csharrison@chromium.org
+Abstract: A new API to measure and attribute cross-site conversions.
 </pre>
 
 Introduction {#intro}
 =====================
 
-Introduction here.
+<em>This section is non-normative</em>
 
+This specification describes how web browsers can provide a mechanism to the
+web that allows measuring and attributing conversions (e.g. purchases) to ads
+a user interacted with on another site. This mechanism should remove the need
+for cross site identifiers like third party cookies.
+
+## Overview ## {#overview}
+
+Anchor tags with `impressiondata` and `conversiondestination` attributes are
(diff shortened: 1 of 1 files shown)


csharrison:
@jyasskin can you take a look at this spec outline?

index.bs
```diff
@@ -1,16 +1,130 @@
...
+
+# Structures # {#structures}
+
+<h3 dfn-type=dfn>Impression</h3>
+
+An impression is a [=struct=] with the following items:
+
+<ul dfn-for="impression">
+* <dfn>impression source</dfn>: An [=url/origin=].
+* <dfn>impression data</dfn>: A {{DOMString}}.
```
jyasskin: For fields in structures, we usually use the data types in Infra, like https://infra.spec.whatwg.org/#strings, instead of WebIDL data types like `DOMString`. Then when it's not obvious, use prose to describe how the notional data type from Infra is converted to the programming type in WebIDL.

index.bs
```diff
@@ -1,16 +1,130 @@
...
+* <dfn>conversion destination</dfn>: An [=url/origin=].
+* <dfn>reporting endpoint</dfn>: An [=url/origin=].
+* <dfn>expiry</dfn>: A {{DOMTimeStamp}}.
```
jyasskin: Since a [DOMTimeStamp](https://heycam.github.io/webidl/#DOMTimeStamp) is just a number of milliseconds, this should say it's a point in time (if that's what it is). Then you convert to the number of ms in whatever API uses it.

See also https://github.com/whatwg/infra/issues/280.

jyasskin:
Oh, also, you should [join the WICG](https://www.w3.org/community/wicg/) and pair your W3C and github accounts as described in https://labs.w3.org/repo-manager/pr/id/WICG/conversion-measurement-api/33.

csharrison:
Thanks jyasskin! Sorry for the long delay on this.

index.bs
```diff
@@ -1,16 +1,130 @@
...
+</ul>
```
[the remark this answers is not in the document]
csharrison: I'll look into improving the bikeshed integration...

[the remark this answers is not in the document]
jyasskin: Here's another sample configuration: https://github.com/WICG/portals/blob/master/Makefile

index.bs
```diff
@@ -1,16 +1,165 @@
...
+and conversion registration requests.
+
+# Fetch monkeypatches # {#fetch-monkeypatches}
+
+TODO: Patch into fetch for cancelling requests redirected to the .well-known
+conversion domain.
+
+# HTML monkeypatches # {#html-monkeypatches}
+
+Rewrite the anchor element to accept the following attributes:
```
jyasskin, line 49: Note that the "attributes" in a WebIDL interface are a different sort of thing than DOM attributes. I think it's fine to check in what you have as a base, but we'll want to investigate the right way to say this.

csharrison: Ack

PR #36, opened 2020-04-04
s/attibution/attribution

davidvancleve:


davidvancleve:
It looks like the CI might be broken: https://travis-ci.com/github/WICG/conversion-measurement-api/builds/158409583.

csharrison:
Thanks, LGTM. Looks like the current CI integration doesn't support running from forks :/ I'll look into that later.

PR #58, opened 2020-08-06
Add a few more algorithms for impression / conversion construction

csharrison:
The PR adds a few algorithms for how to create impression / conversion structs.

***
<a href="https://pr-preview.s3.amazonaws.com/WICG/conversion-measurement-api/pull/58.html" title="Last updated on Oct 13, 2020, 10:40 PM UTC (aac21ab)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/conversion-measurement-api/58/50aeae6...aac21ab.html" title="Last updated on Oct 13, 2020, 10:40 PM UTC (aac21ab)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -44,7 +44,7 @@ and conversion registration requests.
 
 # Fetch monkeypatches # {#fetch-monkeypatches}
 
-TODO: Patch into fetch for cancelling requests redirected to the .well-known
+Issue: Patch into fetch for cancelling requests redirected to the .well-known
 conversion domain.
 
 # HTML monkeypatches # {#html-monkeypatches}
@@ -60,23 +60,23 @@ partial interface HTMLAnchorElement {
 };
 </pre>
 
-The <dfn for="a" dfn-type="element-attr">conversiondestination</dfn> is the
+The <dfn for="a" element-attr>conversiondestination</dfn> is the
 declared destination [=scheme-and-registrable-domain=] of the anchor for
 purposes of conversion measurement
 
-The <dfn for="a" dfn-type="element-attr">impressiondata</dfn> is a string
+The <dfn for="a" element-attr>impressiondata</dfn> is a string
 containing information about the `impression tag` and will be supplied in the
 `conversion report`.
 
 The <dfn for="a" element-attr>reportingorigin</dfn> declares the
 intended [=origin=] to send the `conversion report` for this impression.
 
-The <dfn for="a" dfn-type="element-attr">impressionexpiry</dfn> is the amount
+The <dfn for="a" element-attr>impressionexpiry</dfn> is the amount
 of time in milliseconds the impression should be considered for conversion
 measurement and reporting reporting.
 
 
-TODO: Need monkey patches passing impression data in navigation, and a mechanism
+Issue: Need monkey patches passing impression data in navigation, and a mechanism
 for validating the resulting document matches the conversiondestination.
 
 # Structures # {#structures}
@@ -131,37 +131,47 @@ A conversion report is a [=struct=] with the following items:
 
 # Algorithms # {#algorithms}
 
+<h3 algorithm id="parsing-conversion-destination">Parsing a conversion destination</h3>
+
+To <dfn>parse a conversion destination</dfn> from an <{a}> tag |anchor|,
+1. Let |url| be the result of running the [=URL parser=] on the value of
+    the |anchor|'s <{a/conversiondestination}>.
+1. Return the result of [=obtain a site|obtaining a site=] from |url|'s
+    [=url/origin=].
+
 <h3 algorithm id="creating-impression">Activating an impression</h3>
 
 To <dfn>activate an impression</dfn> from an <{a}> tag |anchor|,
+1. Let |currentTime| be the current time.
 1. Let |impression| be a new [=impression=] struct whose items are:
 
 : [=impression/impression source=]
 :: |anchor|'s [=relevant settings object=]'s [=environment/top-level origin=].
 : [=impression/impression data=]
-:: The result of applying [=parsing conversion data=] to |anchor|'s <{a/impressiondata}>
-    attribute.
+:: The result of applying [=parsing conversion data=] to |anchor|'s
+    <{a/impressiondata}> attribute.
 : [=impression/conversion destination=]
-:: The result of parsing <{a/conversiondestination}> as a site.
+:: The result of running [=parse a conversion destination=] on |anchor|.
 : [=impression/reporting endpoint=]
-:: The result of [=obtain a site|obtaining a site=] from the result of parsing
-    <{a/reportingorigin}> as an origin.
+:: The [=url/origin=] of the result of running the [=URL parser=] on the value
+    of |anchor|'s <{a/reportingorigin}> attribute.
 : [=impression/expiry=]
-:: The point in time <{a/impressionexpiry}> milliseconds in the future.
+:: |currentTime| +  <{a/impressionexpiry}> milliseconds.
 : [=impression/impression time=]
-:: The current time.
+:: |currentTime|.
 
-TODO: How to store the impression
+Issue: Need to spec how to store the impression.
 
 <h3 algorithm id="creating-a-conversion">Creating a conversion</h3>
 
-To <dfn>create a conversion</dfn> from a [=url=] |url|:
+To <dfn>create a conversion</dfn> from a [=url=] |url| and a
+    [=environment settings object=] |environment|.
 
 : [=conversion/conversion source=]
-:: The current top-level origin.
+:: |environment|'s [=environment/top-level origin=].
 : [=conversion/conversion data=]
 :: The result of applying [=parsing conversion data=] to the value associated with the
-    |conversion-data| field in the URL's query part.
+    |conversion-data| field of |url|'s query part.
 : [=conversion/conversion time=]
 :: Be the current time.
 
@@ -173,11 +183,18 @@ To <dfn>register a conversion</dfn> from a [=request=] |request|, run the follow
 
 1. If |request|'s [=request/current url's=] [=url/path=] is not `.well-known/register-conversion`,
     return.
-1. If the request's [=request/redirect count=] is less than 1, return.
+1. If |request|'s [=request/redirect count=] is less than 1, return.
+1. Let |previousUrl| be the second to last [=URL=] in |request|'s
+    [=request/URL list=].
+1. If |request|'s [=request/current url's=] [=url/origin=] is not [=same origin=] with
+    |previousUrl|'s [=url/origin=], return.
 1. Let |conversionToRegister| be the result of applying [=create a conversion=] from the
     request's [=request/current url=].
 
-TODO: How to store the conversion
+Note: the restriction to require a redirect is needed to ensure that the
+request's origin is aware and in control of the conversion registration.
+
+Issue: Need to spec how to store the conversion.
 
 <h3 algorithm id="parsing-data">Parsing data fields</h3>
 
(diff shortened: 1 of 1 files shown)


csharrison:
@jyasskin can you please review? There are a few TODOs that would be good to get feedback on:

1. Best way to integrate with storage
2. Should we use URLSearchParams directly to talk about parsing URL queries?
3. What is best practice for referencing definitions that are intentionally non-exported (e.g. obtain-a-site)

csharrison:
@jyasskin friendly ping

@johnivdel cc

index.bs
```diff
@@ -124,22 +131,61 @@ A conversion report is a [=struct=] with the following items:
...
+:: Be the current time.
+
+TODO: How to store the impression
+
+<h3 algorithm id="creating-a-conversion">Creating a conversion</h3>
+
+To <dfn>create a conversion</dfn> from a [=url=]:
+
+: [=conversion/conversion source=]
+:: The current top-level origin.
```
jyasskin, line 161: What's "current" here? You may need to pass in a settings object to look relative to.

index.bs
```diff
@@ -124,22 +131,61 @@ A conversion report is a [=struct=] with the following items:
...
+: [=conversion/conversion data=]
+:: The result of applying [=parsing conversion data=] to the value associated with the
+    |conversion-data| field in the URL's query part.
+: [=conversion/conversion time=]
+:: Be the current time.
+
+TODO: Formalize how to parse the query similar to URLSearchParams.
```
jyasskin, line 168: If you use

```suggestion
Issue: Formalize how to parse the query similar to URLSearchParams.
```

Bikeshed will format it louder.

index.bs
```diff
@@ -124,22 +131,61 @@ A conversion report is a [=struct=] with the following items:
...
 
 <h3 algorithm id="register-conversion">Register a conversion</h3>
-TODO
 
-<h3 algorithm id="parsing-the-metadata">Parsing the metadata</h3>
+To <dfn>register a conversion</dfn> from a fetch [=request=], run the following steps:
+
+1. If the requests [=request/current url's=] [=url/path=] is not `.well-known/register-conversion`,
```
jyasskin, line 174:
```suggestion
1. If |request|'s [=request/current url's=] [=url/path=] is not `.well-known/register-conversion`,
```

index.bs
```diff
@@ -124,22 +131,61 @@ A conversion report is a [=struct=] with the following items:
...
+: [=conversion/conversion time=]
+:: Be the current time.
+
+TODO: Formalize how to parse the query similar to URLSearchParams.
 
 <h3 algorithm id="register-conversion">Register a conversion</h3>
-TODO
 
-<h3 algorithm id="parsing-the-metadata">Parsing the metadata</h3>
+To <dfn>register a conversion</dfn> from a fetch [=request=], run the following steps:
```
jyasskin, line 172:
```suggestion
To <dfn>register a conversion</dfn> from a [=request=] |request|, run the following steps:
```

index.bs
```diff
@@ -124,22 +131,61 @@ A conversion report is a [=struct=] with the following items:
...
+
+1. If the requests [=request/current url's=] [=url/path=] is not `.well-known/register-conversion`,
+    return.
+1. If the request's [=request/redirect count=] is less than 1, return.
```
jyasskin, line 176: It might be useful to include a `Note: ` here saying why you have this restriction.

index.bs
```diff
@@ -150,7 +196,20 @@ the following steps:
 1. Return |encodedOutput|.
 
 <h3 algorithm id="delivery-time">Establishing report delivery time</h3>
-TODO
+To <dfn>establish report delivery time</dfn> given an [=impression=] and a
+[=conversion/conversion time=], run the following steps:
```
jyasskin, line 200: I might write

```suggestion
The <dfn>report delivery time</dfn> for an [=impression=] |impression| and a
[=conversion/conversion time=] |conversionTime| is the result of the following steps:
```
That phrasing tends to be easier to call. I wonder if "report delivery time" is going to be ambiguous with other reporting APIs, which would argue for "conversion report delivery time".

You might also have this take the conversion instead of just the conversion time, to make the calls a bit simpler and to give the arguments more semantic types.

index.bs
```diff
@@ -150,7 +196,20 @@ the following steps:
...
+1. Let |timeDelta| be the difference between the [=conversion/conversion time=] and [=impression/impression time=].
```
jyasskin, line 201: Maybe
```suggestion
1. Let |conversionTimeAfterImpression| be the difference between the [=conversion/conversion time=] and [=impression/impression time=].
```
would be easier to read inside the following algorithm.

index.bs
```diff
@@ -150,7 +196,20 @@ the following steps:
...
+2. Let |expiryDelta| be the difference between the [=impression/expiry=] and the [=impression/impression time=]
```
jyasskin, line 202: Do you use [=impression/expiry=] anywhere as an actual deadline? If not, it might be better to just store the `impressionexpiry` attribute value as a duration.

index.bs
```diff
@@ -150,7 +196,20 @@ the following steps:
...
+1. If |timeDelta| <= (2 days - 1 hour), return [=impression/impression time=] + 2 days
+1. If
+    - |expiryDelta| > (2 days - 1 hour) and
+    - |expiryDelta| < (7 days - 1 hour) and
+    - |timeDelta| <= |expiryDelta|
+
+
+    return the [=impression/expiry=] + 1 hour.
+1. If |timeDelta| <= (7 days - 1 hour), return [=impression/impression time=] + 7 days
```
jyasskin, line 211: I'm having trouble understanding the conditions where this happens. The expiry is outside of the 2-day-to-7-day range, or the conversion happened after expiration? But what's the expiry for if not to prevent too-long conversion delays?

index.bs
```diff
@@ -150,7 +196,20 @@ the following steps:
 1. Return |encodedOutput|.
 
 <h3 algorithm id="delivery-time">Establishing report delivery time</h3>
-TODO
+To <dfn>establish report delivery time</dfn> given an [=impression=] and a
+[=conversion/conversion time=], run the following steps:
+1. Let |timeDelta| be the difference between the [=conversion/conversion time=] and [=impression/impression time=].
+2. Let |expiryDelta| be the difference between the [=impression/expiry=] and the [=impression/impression time=]
+1. If |timeDelta| <= (2 days - 1 hour), return [=impression/impression time=] + 2 days
```
jyasskin, line 203: Consider using https://w3c.github.io/tr-design/src/readme.html#switch to improve the formatting for this block.

jyasskin (changes requested):
Re your questions:

### Best way to integrate with storage:

If I'm reading https://storage.spec.whatwg.org/#model right, you'll call [=obtain a local storage bottle map=] with an environment settings object and "conversion measurement", and it'll give back a key-value store that you can store things in. Once you have some wording drafted around this, it's probably a good idea to ask Anne to review it by sending a PR to add a row to https://storage.spec.whatwg.org/#registered-storage-endpoints.

### Should we use URLSearchParams directly to talk about parsing URL queries? 

Given https://github.com/whatwg/url/issues/491, probably not ... @domenic, do you already have a spec somewhere for parsing a key-value map from a URL's query?

If `application/x-www-form-urlencoded` does turn out to be the right thing to parse, use https://url.spec.whatwg.org/#urlencoded-parsing from a spec instead of going through the Javascript class.

### What is best practice for referencing definitions that are intentionally non-exported (e.g. obtain-a-site)

If they're accidentally non-exported, I either add an entry to the [`<pre class=anchors>` block](https://tabatkins.github.io/bikeshed/#custom-dfns) or use [`<a spec=spec-name>link</a>`](https://tabatkins.github.io/bikeshed/#linking-to-unexported-definitions) and then send a PR to the other spec to export them. If they're intentionally non-exported, talk to the other spec's editor about what they want you to use. Sometimes it's ok for a monkeypatch to use something that isn't meant for general use, since the monkeypatch will end up inside the other spec.

csharrison (commented):
Thanks so much for the detailed review, I tried to respond to most comments but I'll likely leave the storage stuff to another patch. It turned out my concern about unexported defs was just because I needed to run `bikeshed update`.

index.bs
```diff
@@ -124,22 +131,78 @@ A conversion report is a [=struct=] with the following items:
...
+    <{a/impressiondata}> attribute.
+: [=impression/conversion destination=]
+:: The result of running [=parse a conversion destination=] on |anchor|.
+: [=impression/reporting endpoint=]
+:: The [=url/origin=] of the result of running the [=URL parser=] on the value
+    of |anchor|'s <{a/reportingorigin}> attribute.
+: [=impression/expiry=]
+:: |currentTime| +  <{a/impressionexpiry}> milliseconds.
+: [=impression/impression time=]
+:: |currentTime|.
```
jyasskin, line 161: I think you can indent all of this by 4 spaces in order to make it part of the list item. That'll become important when you add subsequent list items to store the impression.

index.bs
```diff
@@ -124,22 +131,78 @@ A conversion report is a [=struct=] with the following items:
...
+
+Issue: Need to spec how to store the impression.
```
jyasskin, line 163: I'd make this a list item to be clear that it'll be part of the same algorithm.

```suggestion
1. Issue: Need to spec how to store the impression.
```

index.bs
```diff
@@ -124,22 +131,78 @@ A conversion report is a [=struct=] with the following items:
...
+
+<h3 algorithm id="creating-a-conversion">Creating a conversion</h3>
+
+To <dfn>create a conversion</dfn> from a [=url=] |url| and a
```
jyasskin, line 167:
```suggestion
To <dfn>create a conversion</dfn> from a [=url=] |url| and an
```

index.bs
```diff
@@ -124,22 +131,78 @@ A conversion report is a [=struct=] with the following items:
...
+    [=environment settings object=] |environment|.
```
jyasskin, line 168:
```suggestion
[=environment settings object=] |environment|, return a new [=conversion=] struct with the items:
```

index.bs
```diff
@@ -124,22 +131,78 @@ A conversion report is a [=struct=] with the following items:
...
+
+: [=conversion/conversion source=]
+:: |environment|'s [=environment/top-level origin=].
+: [=conversion/conversion data=]
+:: The result of applying [=parsing conversion data=] to the value associated with the
+    |conversion-data| field of |url|'s query part.
```
jyasskin, line 174: I'd probably use

```suggestion
    `"conversion-data"` field of |url|'s [=url/query=].
```

since a field name will be a [string](https://infra.spec.whatwg.org/#string).

index.bs
```diff
@@ -124,22 +131,78 @@ A conversion report is a [=struct=] with the following items:
...
+: [=conversion/conversion time=]
+:: Be the current time.
```
jyasskin, line 176:
```suggestion
:: The current time.
```

index.bs
```diff
@@ -124,22 +131,78 @@ A conversion report is a [=struct=] with the following items:
...
+    [=request/URL list=].
+1. If |request|'s [=request/current url's=] [=url/origin=] is not [=same origin=] with
+    |previousUrl|'s [=url/origin=], return.
+1. Let |conversionToRegister| be the result of applying [=create a conversion=] from the
+    request's [=request/current url=].
+
+Note: the restriction to require a redirect is needed to ensure that the
+request's origin is aware and in control of the conversion registration.
+
+Issue: Need to spec how to store the conversion.
```
jyasskin, line 197: I'd probably put this into a list item as well, again to show readers that the algorithm isn't complete.

index.bs
```diff
@@ -150,7 +213,31 @@ the following steps:
...
-TODO
+The <dfn>report delivery time</dfn> for an [=impression=] |impression| and a
+[=conversion/conversion time=] |conversionTime| is the result of the following steps:
+1. Let |conversionTimeAfterImpression| be the difference between the
+    [=conversion/conversion time=] and [=impression/impression time=].
+2. Let |expiryDelta| be the difference between the [=impression/expiry=] and
+    the [=impression/impression time=]
+
+<dl class="switch">
+<dt>If |conversionTimeAfterImpression| <= (2 days - 1 hour)</dt>
```
jyasskin, line 224: What you have is ok, but I'd usually put the `<dl>` inside a numbered step:

```suggestion
1. If:
    <dl class="switch">
    <dt>|conversionTimeAfterImpression| <= (2 days - 1 hour)</dt>
```
And similarly indented through the rest of the `<dl>`.

index.bs
```diff
@@ -150,7 +213,31 @@ the following steps:
...
+<dd>return [=impression/impression time=] + 2 days.</dd>
```
jyasskin, line 225: Is it intentional that the report delivery time is impression-time + 2 days even if conversion-time > expiry? That might deserve a note.

jyasskin (approved):
I'm sorry it took me so long to get back to this again. I left a bunch of nits, but I think it's in good shape to merge.

index.bs
csharrison: Done

index.bs
csharrison: Done

index.bs
csharrison: Done

index.bs
csharrison: Done

index.bs
csharrison: Done

index.bs
csharrison: Done

index.bs
csharrison: Done

index.bs
csharrison: Done

index.bs
csharrison: Done

index.bs
csharrison: Done

index.bs
csharrison: Ah good catch, this shouldn't be possible because after expiry the impression will be deleted. I'll add a note

csharrison:
Thanks so much for the review, I appreciate the thorough comments as always.

PR #59, opened 2020-08-10
Replaced "metadata" with "data"

### LINKED ISSUE #57: Attributes naming and terms
A few terms and attribute names in the API and explainer may (or may not) create confusion.

1. "Impression" (term) refers to both the tag and the event. 

2. `impression...` (attribute) may refer to both click and view in this spec. But in adtech lingo, "impression" usually means "view" as opposed to "click" (see the [definition from the IAB](https://www.iab.com/insights/glossary-of-terminology/#index-9): "Each time an advertisement loads onto a user’s screen, the ad server may count that loading as one impression" and [this one](https://clearcode.cc/glossary/impression/)). This means that `impression...` does not reflect the duality of "click or view" and that the API may be misunderstood as a view-through measurement API only.

3. Both the terms "data" and "metadata" are used; similarly, we have `impression-data` and `conversion-metadata` attributes. This may be hard to remember and confusing for API users; it's not necessarily clear why one qualifies as data while the other as metadata. **Note:** I think the API already reflects this, and this is just about updating the explainer.

**=> Ideas to solve these / proposals:**

1. In the explainer, clarify with "... tag" when referring to the tag - and stick to "..." when referring to the event.

2. Rename `impression` (@johnivdel suggests: "naming needs to be agnostic to click, since some API variants may support viewed impressions"). 
Ideas: 
- `impression` -> `event`, `conversion` -> `conversion` // may be confusing to differentiate between `eventdata` and `conversiondata`.
- `impression` -> `impact`, `conversion` -> `conversion`
- (by @johnivdel ) `measurement-data`
- (by @johnivdel ) `click-measurement-data` / `view-measurement-data`
- (by @johnivdel ) `link-measurement-data`
- `click-data` / `view-data`
- `click-event-data` / `view-event-data`
- `click-view-event-data`
- `click-view-data`
- `impression` -> `trigger`, `conversion` -> `conversion`. `trigger-data`, `conversion-data`.
- `impression` -> `trigger-event`, `conversion` -> `conversion`. `trigger-event-data`, `conversion-data`.

3. Unify: use data everywhere, both as terms and in the attributes.

maudnals:
Partly fixes #57.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,20 +16,24 @@ See the explainer on [aggregate measurement](AGGREGATE.md) for a potential exten
   - [Glossary](#glossary)
   - [Motivation](#motivation)
   - [Prior Art](#prior-art)
+- [Click Through Conversion Measurement Event-Level API Explainer](#click-through-conversion-measurement-event-level-api-explainer)
+  - [Glossary](#glossary)
+  - [Motivation](#motivation)
+  - [Prior Art](#prior-art)
 - [Overview](#overview)
   - [Impression Declaration](#impression-declaration)
     - [Publisher Controls for Impression Declaration](#publisher-controls-for-impression-declaration)
   - [Conversion Registration](#conversion-registration)
-    - [Metadata limits and noise](#metadata-limits-and-noise)
+    - [Data limits and noise](#data-limits-and-noise)
     - [Register a conversion algorithm](#register-a-conversion-algorithm)
     - [Multiple impressions for the same conversion (Multi-touch)](#multiple-impressions-for-the-same-conversion-multi-touch)
     - [Multiple conversions for the same impression](#multiple-conversions-for-the-same-impression)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -508,12 +512,12 @@ Speculative: Adding noise to the conversion event itself
 --------------------------------------------------------
 
 Another way to add privacy to this system is to add noise not only to
-the [reported conversion metadata value](https://github.com/csharrison/conversion-measurement-api#metadata-limits-and-noise),
+the [reported conversion data value](https://github.com/csharrison/conversion-measurement-api#metadata-limits-and-noise),
```
csharrison, line 515: s/metadata/data in the ref part of the URL (while you're at it you can change the URL from an absolute one to a relative one since this is wrong (we're in WICG now)

maudnals: Right - fixed, changed to relative URL.

PR #62, opened 2020-08-26
Clarify that feature policy changes are "publisher-side"

csharrison:
These controls are accessible to anyone in the publisher's trusted top level context


csharrison:
@maudnals PTAL? Do you think any wording needs to be changed or is this sufficient?

maudnals:
LGTM!

PR #64, opened 2020-08-28
Update default `impressionexpiry` to 30 days

johnivdel:
Chromium implements a 30 day default, and aligning max/default makes this a bit simpler to understand.


csharrison:
LGTM

PR #70, opened 2020-09-23
Add denoiser.py

csharrison:
Adds a simple python script that corrects noisy, aggregate bucket counts using an unbiased estimator.

Partially addresses the concerns in issue #24, to show that aggregate data can be estimated.
Follow-up work can outline a technique to reduce variance at the cost of adding bias.

### DIFF
--- a/denoiser.py
+++ b/denoiser.py
@@ -1,0 +1,66 @@
+import collections
+import random
+
+# This file provides a helper function to denoise reports coming
+# from the event level API (i.e. those where conversion metadata
+# is potentially randomly selected).
+
+# TODO(csharrison): This estimator is unbiased, but its variance
+# can be improved if we add bias. This is something we should
+# be able to smoothly trade off with another parameter.
+def corrected_buckets(buckets, noise_probability=.05):
+  """Returns a map of conversion bits --> corrected counts
+
+  buckets: A map from integer conversion metadata to conversion counts.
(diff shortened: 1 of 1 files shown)


csharrison:
John: PTAL?

denoiser.py
```diff
@@ -0,0 +1,66 @@
...
+    for r in range(count):
+      if random.random() <= noise_probability:
+        new_bucket = random.choice(buckets)
+        noisy_reports[new_bucket] += 1
+      else:
+        noisy_reports[bucket] += 1
+
+  corrected = corrected_buckets(noisy_reports, noise_probability)
+
+  print("Bucket\ttrue count\tnoisy count\tcorrected count")
```
johnivdel, line 64: aside: Could use f-strings if you wanted this table to be properly aligned but probably unnecessary

johnivdel (approved):
Looks good with minor comment

denoiser.py
```diff
@@ -0,0 +1,66 @@
+import collections
+import random
+
+# This file provides a helper function to denoise reports coming
```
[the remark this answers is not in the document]
csharrison: done

denoiser.py
csharrison: yes, I used fstrings which requires python3 so I also added a shebang line

PR #72, opened 2020-10-02
Register only viewed impression

jbpringuey:
There are currently third party vendors using thrid party cookie checking to make sure that impressions have been viewed before making the post view attribution. With the removal of third party cookie and as we are cleaning the ecosystem, we should make sure that only ads that have been seen can register impression.


jbpringuey:
Closing as I am creating an issue first https://github.com/WICG/conversion-measurement-api/issues/78

PR #82, opened 2020-11-13
Add private_histograms_mpc.md

csharrison:
This PR adds a draft for an MPC protocol that supports some basic aggregation computation with differential privacy.

### DIFF
--- a/private_histograms_mpc.md
+++ b/private_histograms_mpc.md
@@ -325,7 +325,7 @@ P(Geo(p) == k) = p(1-p)^k
 ```
 
 
-Consider an AggID count R and R’ that are one-sided-dp-neighbors. wlog there is only one such neighbor to consider, where R has count = n + 1 and R’ has count = n. That is, R’ is a one-sided DP neighbor of R because it consists of replacing a sensitive record with its absence.
+Consider an AggID count R and R’ that are one-sided-DP neighbors. Without loss of generality there is only one such neighbor to consider, where R has count = n + 1 and R’ has count = n. That is, R’ is a one-sided DP neighbor of R because it consists of replacing a sensitive record with its absence.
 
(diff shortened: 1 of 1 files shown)


johnivdel (approved):
lgtm

If we add more docs in it will probably make sense to reorganize the service + protocols into a separate directory from the web-platform specific explainers

csharrison:
> If we add more docs in it will probably make sense to reorganize the service + protocols into a separate directory from the web-platform specific explainers

Yep I was thinking the same thing. We can add a folder sometime later if we need to.

private_histograms_mpc.md
```diff
@@ -0,0 +1,368 @@
...
+See [this paper](https://arxiv.org/pdf/1712.05888.pdf) for more information on one-sided DP. In this formulation we consider the presence of a conversion to be sensitive, and the non-presence of a conversion to be not sensitive. This may be a reasonable relaxation of DP in the event of one malicious helper.
+
+The basic tool here will be adding fake records distributed according to a Geometric distribution.
+
+```
+P(Geo(p) == k) = p(1-p)^k
+```
+
+
+Consider an AggID count R and R’ that are one-sided-dp-neighbors. wlog there is only one such neighbor to consider, where R has count = n + 1 and R’ has count = n. That is, R’ is a one-sided DP neighbor of R because it consists of replacing a sensitive record with its absence.
```
johnivdel, line 328: be consistent with line below. Is wlog meant to be there?

csharrison: wlog = without loss of generality. I'll spell it out.

csharrison: I think the correct hyphen rules would say "one-sided-DP neighbors" and "one-sided DP"

PR #83, opened 2020-11-17
Amend framing on the per-record fake sampling technique

csharrison:
This PR changes the per-record sampling method to be listed as a "further refinement" rather than an alternative. This is because the technique can be composed with the per-output-domain sampling technique and isn't strictly an alternative.


csharrison:
John PTAQL?

private_histograms_mpc.md
```diff
@@ -347,24 +347,20 @@ The output of the refined mechanism can be analyzed based on the DP for sums of
...
+ The [attack](#attacks-against-this-protocol) in the original protocol should be fully mitigated by the internal DP on the counts achieved with this refinement. Although we don't achieve internal DP on presence/absence of a particular AggID, the original attacks against the protocol are all predicated on the attacker inserting known records to recover the AggID -> aggkey mapping. If the attacker does not have one of these "fingerprinted" records, then the PRF should provide sufficient privacy for presence/absence.
 
-More analysis on this idea is likely needed :)
+Note that this technique can be composed with pre-sharing an output domain. In that case, we would sample fake records from the raw records _and_ the output domain, and we would still drop records not included in the output domain at the end. This would allow us to continue avoiding the thresholding step at the end of the protocol.
```
johnivdel, line 362: Consider moving this paragraph below the following, as we jump from talking about alternative 2 -> altertnative 1 + 2 -> alternative 2

johnivdel (approved):
lgtm

private_histograms_mpc.md
csharrison: I chose not to do this because I didn't want this to be seen as an alternative, but primarily as an additive step. I reworded to make this more clear.

Issue #77, opened 2020-10-20
Impression declaration for links opened through Window.open


wilsonssun:
The [API](https://github.com/WICG/conversion-measurement-api#impression-declaration) currently only specifies implementation for links opened through HTML anchor tags, but we also have links that are programmatically navigated through [Window.open](https://developer.mozilla.org/en-US/docs/Web/API/Window/open). What, if any, is the appropriate way to do impression declaration for these?

johnivdel:
Supporting JS navigation APIs seems reasonable, especially in the context of #22 and privacycg/private-click-measurement#31, which discuss adding JS support for conversion registration.

A few directions we could go to support window.open navigations are:

####  Supply impression registration attributes directly to the window.open() call

Add a new param to window.open which takes an impression for the navigation.
```
Window? open(
  optional USVString url="",
  optional DOMString target = "_blank",
  optional [TreatNullAs=EmptyString] DOMString features = "",
  optional ImpressionInfo impression);
```

where ImpressionInfo is a new IDL dictionary:

```
dictionary ImpressionInfo {

  required DOMString impressionData;

  // This can be optional as we already have the |url| from the
  // window.open call, but this is still necessary if the
  // site is doing redirects.
  optional USVString conversionDestination;

  optional USVString reportingOrigin;
  optional unsigned long impressionExpiry;
}
```

One downside to this approach is that it does not work for other JS Navigation API's such as window.location.assign(), and params would need to be added to support those navigations as well.

#### Add a separate API to associate navigation urls with impression attributes
Add an API surface which associates an `ImpressionInfo` dictionary with a future navigation url.

```
[Exposed=Window] interface Measurement {

 // ForNavigation?
 void setImpressionForTarget( USVString url , ImpressionInfo info);
}
```

Example usage:
```
// This navigation has no impression.
window.open("https://advertiser.example/a.html", "_blank")

window.measurement.setImpressionForTarget("https://advertiser.example/a.html", {
  impressionData: "1234",
  reportingOrigin: "https://reporting.example" });

// This navigation has an impression!
window.open("https://advertiser.example/a.html", "_blank")

// This navigation doesn't!
window.open("https://advertiser.example/b.thml", "_blank")
```
This would allow impressions to be registered for any navigation, including window.open and navigations from anchor tags.

Impression registered in this manner would not make `impressiondata` visible to other scripts on the page. This could help with some issues discussed on #13 involving reporting fake conversions for impressions via scraping impression attributes.

[...]

PR #86, opened 2020-11-23
Add extension for window.open impressions

johnivdel:
Adds a section denoting how to register impressions for window.open navigations per #77

### DIFF
--- a/README.md
+++ b/README.md
@@ -147,7 +147,7 @@ dictionary ImpressionParams {
 }
 ```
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -115,9 +120,39 @@ Impression attributes:
...
+```
+dictionary ImpressionParams {
+  required DOMString impressionData;
+  required USVString conversionDestination;
+  optional USVString reportingOrigin;
+  optional unsigned long impressionExpiry;
+}
+```
+
+At the time window.open() is invoked, if the associated window must has a [transient activation](https://html.spec.whatwg.org/multipage/interaction.html#transient-activation), an impression event will be created and handled following [Handling an impression event](#handling-an-impression-event).
```
csharrison, line 150: delete "must" in "must has a"

csharrison (approved):

PR #97, opened 2021-01-07
fix query name of `attribution-credit` to `credit`

shigeki:
Chrome currently sends a report with a query parameter name of not `attribution-credit` but `credit`.
I'm not sure if this is a bug on Chrome, but it is better to follow the current implementation to avoid confusion.
If  `attribution-credit` is the right naming, we can rename it again with changes of other attribution names in the next version.


shigeki:
Closing this for fixed by #103.

PR #110, opened 2021-02-18
Make well-known endpoints scoped to attribution-reporting

johnivdel:
This aligns with https://github.com/privacycg/private-click-measurement/issues/59#issue-762861988, which indicates that we should use a path component which indicates these are being used for the Attribution Reporting API


csharrison (approved):

csharrison:
Thanks

PR #122, opened 2021-03-24
Use straight quotes inside of allow attribute value

johnivdel:
This is currently written using left and right quotes, which aren't handled by the Permissions Policy spec given the definition of the 'src' keyword, https://w3c.github.io/webappsec-permissions-policy/#allowlists


csharrison:
🤦 oops

csharrison (approved):

PR #103, opened 2021-01-29
Remove "impression"/"conversion" from API surfaces

johnivdel:
This updates the proposed API with the naming changes discussed on #57.

This includes updating the reporting format to a JSON body with the new names

PR #131, opened 2021-04-14
Update spec to current explainer

johnivdel:
This updates the draft spec to include all changes which were landed to the explainer since it was last updated.

This includes:
 - all naming changes on #103 and #121
 - removal of hexadecimal encoding for data fields
 - clarification on the value represented by a/attributiondestination

Algorithm definitions and variables are updated to reflect the new naming.

***
<a href="https://pr-preview.s3.amazonaws.com/WICG/conversion-measurement-api/pull/131.html" title="Last updated on Apr 30, 2021, 7:11 PM UTC (32fbcc1)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/conversion-measurement-api/131/6e542cc...32fbcc1.html" title="Last updated on Apr 30, 2021, 7:11 PM UTC (32fbcc1)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -66,7 +66,7 @@ which declares the intended final navigation url resulting from running
 
 Note: This origin does not need to match the exact final origin of the navigation. 
 However, this navigation and the final navigation url have to produce the same result when
-passed to [=obtain a site=]
+passed to [=obtain a site=].
 
 The <dfn for="a" element-attr>attributionsourceeventid</dfn> is a string
 containing information about the `attribution source` and will be supplied in the
@@ -78,13 +78,10 @@ intended [=origin=] to send the [=attribution report=] for this source.
 The <dfn for="a" element-attr>attributionexpiry</dfn> is the amount
 of time in milliseconds the attribution source should be considered for reporting.
 
-
-Issue: Need monkey patches passing impression data in navigation, and a mechanism
-for validating the resulting document matches the conversiondestination.
+Issue: Need monkey patches passing attribution source in navigation, and a mechanism
+for validating the resulting document matches the attributiondestination.
 
 # Structures # {#structures}
-
-<h3 dfn-type=dfn>Impression</h3>
 
 <h3 dfn-type=dfn>Attribution source</h3>
 
@@ -146,8 +143,6 @@ To <dfn>parse an attribution destination</dfn> from a string |str|:
 1. Return the result of [=obtain a site|obtaining a site=] from |url|'s
     [=url/origin=].
 
-<h3 algorithm id="creating-impression">Activating an impression</h3>
-
 <h3 algorithm id="obtaining-attribution-source-anchor">Obtaining an attribution source from an <code>a</code> element</h3>
 
 To <dfn>obtain an attribution source</dfn> from an <{a}> element |anchor|:
@@ -172,7 +167,7 @@ To <dfn>obtain an attribution source</dfn> from an <{a}> element |anchor|:
 
 <dfn>Max event id value</dfn> is a vendor specific integer which controls the meximum size value which can be used as an [=attribution source/event id=]
 
-1. Issue: Need to spec how to store the impression.
+1. Issue: Need to spec how to store the attribution source.
 
 <h3 algorithm id="attribution-trigger-creation">Creating an attribution trigger</h3>
 
@@ -236,7 +231,8 @@ return a point in time.
 1. Let |expiryDelta| be the difference between the |source|'s [=attribution source/expiry=] and
     the |source|'s [=attribution source/source time=]
 
-    Note: |timeToTrigger| is less than |expiryDelta| because it is not normally possible to convert an expired impression.
+    Note: |timeToTrigger| is less than |expiryDelta| because it is not normally possible to
+    convert an expired attribution source.
 
 1. If:
     <dl class="switch">
--- a/index.bs
+++ b/index.bs
@@ -49,7 +49,7 @@ conversion domain.
 
 # HTML monkeypatches # {#html-monkeypatches}
 
-Rewrite the <{a}> to accept the following attributes:
+Add the following content attributes to the <{a}> element:
 
 <pre class="idl">
 partial interface HTMLAnchorElement {
@@ -168,7 +168,7 @@ To <dfn>obtain an attribution source</dfn> from an <{a}> element |anchor|:
     :: |sourceOrigin|
     : [=attribution source/event id=]
     :: The result of running [=parse attribution data=] with |anchor|'s
-        <{a/attributionsourceeventid}> attribute and [=max event id value=].
+        <{a/attributionsourceeventid}> attribute modulo [=max event id value=].
     : [=attribution source/attribution destination=]
     :: |attributionDestination|
     : [=attribution source/reporting endpoint=]
@@ -182,11 +182,11 @@ To <dfn>obtain an attribution source</dfn> from an <{a}> element |anchor|:
 <dfn>Max event id value</dfn> is a vendor specific integer which controls 
 the maximum size value which can be used as an [=attribution source/event id=]
 
-1. Issue: Need to spec how to store the attribution source.
+Issue: Need to spec how to store the attribution source.
 
 <h3 algorithm id="attribution-trigger-creation">Creating an attribution trigger</h3>
 
-To <dfn>obtain an attribution trigger</dfn> given a [=url=] |url|, and an
+To <dfn>obtain an attribution trigger</dfn> given a [=url=] |url| and an
 [=environment settings object=] |environment|, return a [=attribution trigger=]
 with the items:
 
--- a/index.bs
+++ b/index.bs
@@ -1,13 +1,13 @@
 <pre class='metadata'>
-Title: Conversion Measurement
-Shortname: conversion-measurement
+Title: Attribution Reporting
+Shortname: attribution-reporting
 Level: 1
 Status: CG-DRAFT
 Group: wicg
 Repository: WICG/conversion-measurement-api
 URL: https://wicg.github.io/conversion-measurement-api
 Editor: Charlie Harrison, Google Inc. https://google.com, csharrison@chromium.org
-Abstract: A new API to measure and attribute cross-site conversions.
+Abstract: A new API to measure and attribute cross-site events.
 
 Markup Shorthands: markdown on
 Complain About: accidental-2119 on, missing-example-ids on
@@ -23,24 +23,24 @@ Introduction {#intro}
 <em>This section is non-normative</em>
 
 This specification describes how web browsers can provide a mechanism to the
-web that allows measuring and attributing conversions (e.g. purchases) to ads
+web that supports measuring and attributing conversions (e.g. purchases) to ads
 a user interacted with on another site. This mechanism should remove one need
 for cross site identifiers like third party cookies.
 
 ## Overview ## {#overview}
 
-An anchor tag with <{a/impressiondata}> and <{a/conversiondestination}> attributes is
-classified as an <dfn export>impression tag</dfn>. When impression tags are clicked, and the
-resulting navigation commits in a document matching the <{a/conversiondestination}>,
-then the impression is stored in UA storage.
+A page can register an [=attribution source=] on a site by providing
+ <{a/attributionsourceeventid}> and <{a/attributiondestination}> attributes on an <{a}> element.
+When such an <{a}> element is clicked, and the resulting navigation commits in a document matching
+the <{a/attributiondestination}>, the [=attribution source=] is stored in UA storage.
 
-At a later point, the <{a/conversiondestination}> site may fire an HTTP request to
-trigger conversion registration, which matches up with any previously
-stored impressions. If matching impressions exist, they are scheduled to be
+At a later point, the <{a/attributiondestination}> site may fire an HTTP request to
+trigger attribution, which matches an [=attribution trigger=] with any previously
+stored sources. If matching sources exist, they are scheduled to be
 reported at a later time, possibly multiple days in the future.
 
-Reports are sent to reporting endpoints that are configured in impression tags
-and conversion registration requests.
+Reports are sent to reporting endpoints that are configured in attribution sources
+and attribution trigger requests.
 
 # Fetch monkeypatches # {#fetch-monkeypatches}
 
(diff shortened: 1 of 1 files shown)


johnivdel:
@jyasskin could you take a look at this? Thanks!

cc @csharrison

index.bs
```diff
@@ -49,198 +49,207 @@ conversion domain.
 
 # HTML monkeypatches # {#html-monkeypatches}
 
-Rewrite the anchor element to accept the following attributes:
+Rewrite the <{a}> to accept the following attributes:
 
 <pre class="idl">
 partial interface HTMLAnchorElement {
```
jyasskin, line 55: Not new in this PR, so better to make another PR to fix it, but this only adds to the `HTMLAnchorElement` interface, and not to the allowed content attributes.

index.bs
```diff
@@ -49,198 +49,207 @@ conversion domain.
...
+    [CEReactions, Reflect] attribute DOMString attributionsourceeventid;
+    [CEReactions, Reflect] attribute DOMString attributionreportto;
+    [CEReactions, Reflect] attribute unsigned long long attributionexpiry;
 };
 </pre>
 
-The <dfn for="a" element-attr>conversiondestination</dfn> is the
-declared destination [=scheme-and-registrable-domain=] of the anchor for
-purposes of conversion measurement
+The <dfn for="a" element-attr>attributiondestination</dfn> is an [=url/origin=]
```
jyasskin, line 63: Do you define anywhere what should happen if the user sets something that isn't an origin (or isn't even a URL) into this string?

index.bs
```diff
@@ -49,198 +49,207 @@ conversion domain.
...
+which declares the intended final navigation url resulting from running
```
jyasskin, line 64: An origin won't declare the exact final navigation URL. And we can incorporate the note below into this paragraph. Perhaps

```suggestion
that is intended to be [=same site=] with the origin of the final navigation url resulting from running
```

index.bs
```diff
@@ -49,198 +49,207 @@ conversion domain.
...
+Issue: Need monkey patches passing attribution source in navigation, and a mechanism
+for validating the resulting document matches the attributiondestination.
 
 # Structures # {#structures}
 
-<h3 dfn-type=dfn>Impression</h3>
+<h3 dfn-type=dfn>Attribution source</h3>
 
-An impression is a [=struct=] with the following items:
+An attribution source is a [=struct=] with the following items:
```
jyasskin, line 88: For a followup, it'd be nice for readers if you include a sentence for each of these structs saying what they're intended to mean. Those sentences are non-normative, so you can wrap them in `<span class=note>` if you're feeling formal, but most spec authors don't bother.

index.bs
```diff
@@ -49,198 +49,207 @@ conversion domain.
...
 
-<dl dfn-for="impression">
-: <dfn>impression source</dfn>
+<dl dfn-for="attribution source">
+: <dfn>source origin</dfn>
 :: An [=url/origin=].
```
jyasskin, line 92: For a followup, it'd be good to include a statement of what each of the fields is intended to mean, or where it comes from. E.g. this one could be "copied from the clicked <{a}> element's [=Node/node document=]'s [=Document/origin=]."

index.bs
```diff
@@ -49,198 +49,207 @@ conversion domain.
...
+    : [=attribution source/expiry=]
+    :: |currentTime| +  <{a/attributionexpiry}> milliseconds.
+    : [=attribution source/source time=]
     :: |currentTime|.
+1. Return |source|
+
+<dfn>Max event id value</dfn> is a vendor specific integer which controls the meximum size value which can be used as an [=attribution source/event id=]
 
-1. Issue: Need to spec how to store the impression.
+1. Issue: Need to spec how to store the attribution source.
```
jyasskin, line 170:
```suggestion
Issue: Need to spec how to store the attribution source.
```

index.bs
```diff
@@ -49,198 +49,207 @@ conversion domain.
...
+    : [=attribution source/reporting endpoint=]
+    :: The [=url/origin=] of the result of running the [=URL parser=] with
+        |anchor|'s <{a/attributionreportto}> attribute.
+    : [=attribution source/expiry=]
+    :: |currentTime| +  <{a/attributionexpiry}> milliseconds.
+    : [=attribution source/source time=]
     :: |currentTime|.
+1. Return |source|
+
+<dfn>Max event id value</dfn> is a vendor specific integer which controls the meximum size value which can be used as an [=attribution source/event id=]
```
jyasskin, line 168:
```suggestion
<dfn>Max event id value</dfn> is a vendor specific integer which controls the maximum size value which can be used as an [=attribution source/event id=]
```

index.bs
```diff
@@ -49,198 +49,207 @@ conversion domain.
...
-[=string=] |input| and a unsigned long long |maxData|.
+[=attribution source/event id=] and [=attribution trigger/trigger data=].
 
-<dfn>Parsing conversion data</dfn> from |input| with |maxData| returns the result of
-the following steps:
+To <dfn>parse attribution data</dfn> given a [=string=] |input|, and an unsigned long long
+|maxData| perform the  following steps. They return an unsigned long long:
 
-1. Let |decodedInput| be the result of decoding |input| as a base-16 integer.
+1. Let |decodedInput| be the result of decoding |input| as a base-10 integer.
```
jyasskin, line 219: This needs to define what happens if the value isn't a base-10 integer. Probably call https://html.spec.whatwg.org/multipage/common-microsyntaxes.html#rules-for-parsing-non-negative-integers and handle the resulting errors.

index.bs
```diff
@@ -49,198 +49,207 @@ conversion domain.
...
+An attribution source is a [=struct=] with the following items:
 
-<dl dfn-for="impression">
-: <dfn>impression source</dfn>
+<dl dfn-for="attribution source">
+: <dfn>source origin</dfn>
 :: An [=url/origin=].
-: <dfn>impression data</dfn>
+: <dfn>event id</dfn>
 :: A [=string=].
```
jyasskin, line 94: You're setting this to an integer in [=obtain an attribution source=].

index.bs
```diff
@@ -49,198 +49,207 @@ conversion domain.
...
-1. Issue: Need to spec how to store the impression.
+1. Issue: Need to spec how to store the attribution source.
 
-<h3 algorithm id="creating-a-conversion">Creating a conversion</h3>
+<h3 algorithm id="attribution-trigger-creation">Creating an attribution trigger</h3>
 
-To <dfn>create a conversion</dfn> from a [=url=] |url| and an
-[=environment settings object=] |environment|, return a new [=conversion=]
-struct with the items:
+To <dfn>obtain an attribution trigger</dfn> given a [=url=] |url|, and an
```
jyasskin, line 174:
```suggestion
To <dfn>obtain an attribution trigger</dfn> given a [=url=] |url| and an
```

index.bs
```diff
@@ -49,198 +49,207 @@ conversion domain.
...
-    :: The result of running [=parse a conversion destination=] on |anchor|.
-    : [=impression/reporting endpoint=]
-    :: The [=url/origin=] of the result of running the [=URL parser=] on the value
-        of |anchor|'s <{a/reportingorigin}> attribute.
-    : [=impression/expiry=]
-    :: |currentTime| +  <{a/impressionexpiry}> milliseconds.
-    : [=impression/impression time=]
+    : [=attribution source/event id=]
+    :: The result of running [=parse attribution data=] with |anchor|'s
+        <{a/attributionsourceeventid}> attribute and [=max event id value=].
```
jyasskin, line 156: Here I'd probably phrase it 
```suggestion
        <{a/attributionsourceeventid}> attribute modulo [=max event id value=].
```

and match that in the definition. It doesn't quite match the advice in https://infra.spec.whatwg.org/#algorithm-params, but I think it winds up more readable.

index.bs
```diff
@@ -49,198 +49,207 @@ conversion domain.
...
+    : [=attribution source/attribution destination=]
+    :: The result of running [=parse an attribution destination=] with |anchor|'s <{a/attributiondestination}> attribute.
+    : [=attribution source/reporting endpoint=]
+    :: The [=url/origin=] of the result of running the [=URL parser=] with
```
jyasskin, line 160: The URL parser can return an error, and you need to say what happens in that case.

index.bs
```diff
@@ -49,198 +49,207 @@ conversion domain.
...
-To <dfn>create a conversion</dfn> from a [=url=] |url| and an
-[=environment settings object=] |environment|, return a new [=conversion=]
-struct with the items:
+To <dfn>obtain an attribution trigger</dfn> given a [=url=] |url|, and an
+[=environment settings object=] |environment|, return a [=attribution trigger=]
+with the items:
 
-    : [=conversion/conversion source=]
+    : [=attribution trigger/trigger origin=]
     :: |environment|'s [=environment/top-level origin=].
```
jyasskin, line 179: Why is this the top-level origin instead of just the origin? Can attribution triggers happen in cross-origin iframes? Have you discussed somewhere the security implications of writing a different origin into the attribution trigger, and what it means for specs that need to trust or distrust that value?

index.bs
```diff
@@ -49,198 +49,207 @@ conversion domain.
...
+1. If |request|'s [=request/current url's=] [=url/path=] is not `.well-known/attribution-reporting/trigger-attribution`,
     return.
 1. If |request|'s [=request/redirect count=] is less than 1, return.
 1. Let |previousUrl| be the second to last [=URL=] in |request|'s
     [=request/URL list=].
 1. If |request|'s [=request/current url's=] [=url/origin=] is not [=same origin=] with
     |previousUrl|'s [=url/origin=], return.
-1. Let |conversionToRegister| be the result of applying [=create a conversion=] from the
-    request's [=request/current url=].
+1. Let |trigger| be the result of running [=obtain an attribution trigger=] with
```
jyasskin, line 203: This doesn't pass the environment settings object you expect above.

jyasskin (changes requested):
I haven't gotten all the way through the spec, but here's a bunch of comments to start on.

index.bs
johnivdel: Ack.

index.bs
johnivdel: Ack.

index.bs
johnivdel: Fixed. I noted this is a 64-bit integer, but I couldn't find any precedence in the html spec for this.

index.bs
```diff
@@ -49,198 +49,207 @@ conversion domain.
...
 <pre class="idl">
 partial interface HTMLAnchorElement {
-    [CEReactions, Reflect] attribute DOMString conversiondestination;
-    [CEReactions, Reflect] attribute DOMString impressiondata;
-    [CEReactions, Reflect] attribute DOMString reportingorigin;
-    [CEReactions, Reflect] attribute unsigned long long impressionexpiry;
+    [CEReactions, Reflect] attribute DOMString attributiondestination;
+    [CEReactions, Reflect] attribute DOMString attributionsourceeventid;
+    [CEReactions, Reflect] attribute DOMString attributionreportto;
+    [CEReactions, Reflect] attribute unsigned long long attributionexpiry;
```
johnivdel, line 59: unsigned long long is not a type defined in https://html.spec.whatwg.org/#reflecting-content-attributes-in-idl-attributes

This either needs to be an unsigned long or a string. I can address this in a separate PR.

index.bs
johnivdel: Ack.

index.bs
```diff
@@ -1,13 +1,13 @@
 <pre class='metadata'>
-Title: Conversion Measurement
-Shortname: conversion-measurement
+Title: Attribution Reporting
+Shortname: attribution-reporting
```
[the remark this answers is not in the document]
johnivdel: Hopefully one change is enough :)

johnivdel (commented):
Thanks for the great comments Jeffrey!

index.bs
```diff
@@ -49,198 +49,223 @@ conversion domain.
...
+1. Let |sourceOrigin| be |anchor|'s [=relevant settings object=]'s
+    [=environment/top-level origin=].
+1. Let |reportingOrigin| be |sourceOrigin|.
+1. If |anchor| has an <{a/attributionreportto}> attribute, then:
+    1. Let |reportingUrl| be the result of running the
+        [=URL parser=] with |anchor|'s <{a/attributionreportto}> value
+    1. If |reportingUrl| is failure or null, return null.
+    1. Set |reportingOrigin| to |reportingUrl|'s [=url/origin=].
+1. If |anchor| has an <{a/attributionexpiry}> attribute, let |expiry| be that 
+    attribute's value in milliseconds.
```
jyasskin, line 162: Use HTML's integer parsing rules here too.

index.bs
```diff
@@ -49,198 +49,223 @@ conversion domain.
...
+    :: |sourceOrigin|
+    : [=attribution source/event id=]
+    :: The result of running [=parse attribution data=] with |anchor|'s
+        <{a/attributionsourceeventid}> attribute modulo [=max event id value=].
+    : [=attribution source/attribution destination=]
+    :: |attributionDestination|
+    : [=attribution source/reporting endpoint=]
+    :: |reportingOrigin|
+    : [=attribution source/expiry=]
+    :: |currentTime| +  |expiry|
```
jyasskin, line 177:
```suggestion
    :: |currentTime| + |expiry|
```

index.bs
jyasskin: I was worried about a hostile iframe being able to create somehow-hostile attributions to the surrounding page, but I'm not sure what attacks that would actually be useful for. It helps that the user has to click in order to make the attribution source happen. Maybe hostile triggers are more worrying, since they might cause payment from the trigger to the source? In any case, this sort of attribution (no pun intended) of one origin's actions to a different origin is worth talking about in the security considerations.

jyasskin (approved):

index.bs
johnivdel: Yes that would cause issues. Sources are also problematic: a hostile frame logging additional sources could cause credits to be assigned incorrectly, or cause the browser to delete already existing sources in some cases.

The main mitigation we have is [requiring a permissions policy](https://github.com/WICG/conversion-measurement-api#publisher-side-controls-for-attribution-source-declaration) to invoke the API inside of iframes, which is disabled by default for cross-origin iframes.

I will make a note to address this in Security Considerations when the Permissions Policy is added to this draft spec.

Issue #130, opened 2021-04-13
Use a different overload to handle window.open navigations


johnivdel:
As pointed out in https://crbug.com/1164959, there is some existing web usage of a fourth parameter to window.open, "replace", as documented in this old draft spec https://www.w3.org/TR/2009/WD-html5-20090423/browsers.html#dom-open.

This raises concerns that modifying the pre-existing window.open API makes it hard to detect browser support for the feature. There is no way to determine whether a native function accepts a specific type of argument in a spot.

This could be avoided by exposing a separate window associated function like `openWithAttributionSource`, which avoids the legacy usage collision, and provides explicit feature detection.

johnivdel:
cc @domenic for thoughts

domenic:
Hmm.

Introducing a new method like this seems tricky because `window.open()` is full of legacy design mistakes. So, it'd be a shame to introduce a whole new method that carries them over and then just adds a single piece of functionality.

My instinct is it'd be more palatable to design a new method from scratch. Something like:

```js
window.openWindow(url, { allowOpenerAccess, referrerPolicy, impressionParams });
window.openPopup(url, { left, top, width, height, allowOpenerAccess, referrerPolicy, impressionParams });
```

In particular such a clean slate would:

- Not encode left/top/width/height/noopener/noreferrer into strings
- Not also have a second use where you pass a window name and it navigates that window
- Not infer "popupness" (as opposed to new-tab-ness) from the width/height/left/top, but instead have it be explicit
- Allow any referrer policy, not just `no-referrer` (via today's `noreferrer` string option)
- Flip the default so usually you don't get opener access, and you have to opt in to it
- Not parse the URL relative to the entry settings object (https://github.com/whatwg/html/issues/1431) but instead use the relevant settings object like other modern APIs.
- Not special case the empty string or `about:blank` URLs

/cc @annevk as he might also be enthusiastic about fixing some of these legacy mistakes. I'd be happy to help with the spec for such a method if this path sounds interesting to you.

However, note that feature detection for such a design would require the same contortions (https://github.com/heycam/webidl/issues/107) as your current design, of adding to `window.open()`, does:

```js
let supportsImpressionParams = false;
try {
  window.openPopup("about:blank", {
    get impressionParams() {
      supportsImpressionParams = true;
    },
    get width() {
      // Since we don't actually want to open a popup, throw in an
      // alphabetically-later dictionary member
      throw new Error();
    }
} catch {}

[...]

PR #138, opened 2021-04-28
Update window.open interface to use features parameter

johnivdel:
This updates the window.open overload to supply attribution reporting attributes in the features parameter, rather than in a new window.open overload per #130


csharrison (approved):
LGTM. A bit unfortunate the we lose the AttributionSourceParams for window.open and now only have it for event source types, but seems like we have general consensus.

Issue #133, opened 2021-04-21
Unspecified [Reflect] extended attribute used for <a> attributes


johnivdel:
As noted in this [comment](https://github.com/WICG/conversion-measurement-api/pull/131/files#r614457070) on #131, [Reflect] is not defined by https://heycam.github.io/webidl/#es-extended-attributes. 

Instead this should use language like https://html.spec.whatwg.org/multipage/common-dom-interfaces.html#reflect

csharrison:
I think this is fixed?

Issue #123, opened 2021-03-25
Need notes for 64bit integer of attributionsourceeventid and source_event_id


shigeki:
The current description of `attributionsourceeventid=[unsigned long long]` leads to an implementation pitfall for JavaScript developers who missed `Number.MAX_SAFE_INTEGER`. 
It should add a note that `attributionsourceeventid` can be accepted as `DOMString` and need not be handled in  Integer or BigInt.

The description of `"source_event_id: 64 bit event id set on the attribution source"` in the JSON report would also cause interoperability issues as described in https://tools.ietf.org/html/rfc8259#section-6. It should be a string type.

csharrison:
Thanks Shigeki this is great feedback. Agree with you on both counts.

@johnivdel FYI

johnivdel:
Resolved in the merges above.

PR #140, opened 2021-05-04
Update <a> element monkey patch with content attributes

johnivdel:
Includes the list of content attributes to add, and specifies that they must be reflected from the idl attributes as discussed on #133

***
<a href="https://pr-preview.s3.amazonaws.com/WICG/conversion-measurement-api/pull/140.html" title="Last updated on May 19, 2021, 5:26 PM UTC (079db6f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/conversion-measurement-api/140/5c456e6...079db6f.html" title="Last updated on May 19, 2021, 5:26 PM UTC (079db6f)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -49,6 +49,28 @@ conversion domain.
 
 # HTML monkeypatches # {#html-monkeypatches}
 
+<h3 id="longlong-reflection"> long long reflection </h3>
+
+Add the following rules for <a spec=html>reflecting</a> <a spec=html>content attributes</a>:
+
+If a reflecting IDL attribute has a signed integer type ({{long long}}) then, on getting, the content attribute must be
+parsed according to the <a spec="html">rules for parsing integers</a>, and if that is successful, and the value is in the
+range of the IDL attribute's type, the resulting value must be returned. If, on the other hand, it fails or returns
+an out of range value, or if the attribute is absent, then the default value must be returned instead, or 0 if there
+is no default value. On setting, the given value must be converted to the shortest possible string representing the
+number as a valid integer and then that string must be used as the new content attribute value.
+
+If a reflecting IDL attribute has a signed integer type ({{long long}}) that is <dfn>limited to only non-negative numbers</dfn> then,
+on getting, the content attribute must be parsed according to the <a spec="html">rules for parsing non-negative integers</a>, and if
+that is successful, and the value is in the range of the IDL attribute's type, the resulting value must be returned.
+If, on the other hand, it fails or returns an out of range value, or if the attribute is absent, the default value
+must be returned instead, or −1 if there is no default value. On setting, if the value is negative, the user agent
+must throw an {{"IndexSizeError"}} {{DOMException}}. Otherwise, the given value must be converted to the shortest possible
+string representing the number as a valid non-negative integer and then that string must be used as the new content
+attribute value.
+
+<h3 id="monkeypatch-anchor">&lt;a&gt; element</h3>
+
 Add the following <a spec=html>content attributes</a> to the <{a}> element:
 
 : <{a/attributionsourceeventid}>
@@ -72,8 +94,11 @@ partial interface HTMLAnchorElement {
 </pre>
 
 The IDL attributes {{HTMLAnchorElement/attributionDestination}}, {{HTMLAnchorElement/attributionSourceEventId}}, 
-{{HTMLAnchorElement/attributionReportTo}}, {{HTMLAnchorElement/attributionExpiry}}
-must <a spec=html>reflect</a> the respective content attributes of the same name.
+{{HTMLAnchorElement/attributionReportTo}} must <a spec=html>reflect</a> the respective content 
+attributes of the same name.
+
+The IDL attribute {{HTMLAnchorElement/attributionExpiry}} must reflect the <{a/attributionexpiry}>
+content attribute, [=limited to only non-negative numbers=].
 
 The <dfn for="a" element-attr>attributiondestination</dfn> attribute is a string
 representing an [=url/origin=] that is intended to be [=same site=] with the origin
--- a/index.bs
+++ b/index.bs
@@ -64,15 +64,15 @@ Extend the <{a}> element's <a spec=html>DOM interface</a> to include the followi
 
 <pre class="idl">
 partial interface HTMLAnchorElement {
-    [CEReactions] attribute DOMString attributiondestination;
-    [CEReactions] attribute DOMString attributionsourceeventid;
-    [CEReactions] attribute DOMString attributionreportto;
-    [CEReactions] attribute DOMString attributionexpiry;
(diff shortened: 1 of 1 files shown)


johnivdel:
@jyasskin could you take a look?

index.bs
```diff
@@ -49,29 +49,45 @@ conversion domain.
...
+: <{a/attributionsourceeventid}>
+:: Identifies the declared attribution source
+: <{a/attributiondestination}>
+:: Site which can attribute an event to the declared attribution source
+: <{a/attributionreportto}>
+:: [=url/origin=] to receive attribution reports
+: <{a/attributionexpiry}>
+:: Length of time the attribution souce is valid
+
+Extend <{a}> element <a spec=html>DOM interface</a> to include the following interface:
```
jyasskin, line 63:
```suggestion
Extend the <{a}> element's <a spec=html>DOM interface</a> to include the following interface:
```

index.bs
```diff
@@ -49,29 +49,45 @@ conversion domain.
 
 # HTML monkeypatches # {#html-monkeypatches}
 
-Add the following content attributes to the <{a}> element:
+Add the following <a spec=html>content attributes</a> to the <{a}> element:
```
jyasskin, line 52: We should file a bug against HTML to check if there's a better way to do this than referring to a bunch of non-exported `dfn`s. E.g. HTML could export these definitions.

I believe Anne also likes us to file an HTML issue saying that this monkeypatch exists, like https://github.com/whatwg/fetch/issues/784.

index.bs
```diff
@@ -49,29 +49,45 @@ conversion domain.
...
 <pre class="idl">
 partial interface HTMLAnchorElement {
-    [CEReactions, Reflect] attribute DOMString attributiondestination;
-    [CEReactions, Reflect] attribute DOMString attributionsourceeventid;
-    [CEReactions, Reflect] attribute DOMString attributionreportto;
-    [CEReactions, Reflect] attribute unsigned long long attributionexpiry;
+    [CEReactions] attribute DOMString attributiondestination;
+    [CEReactions] attribute DOMString attributionsourceeventid;
+    [CEReactions] attribute DOMString attributionreportto;
+    [CEReactions] attribute DOMString attributionexpiry;
```
jyasskin, line 70: Glancing over the HTML spec, it does look like numeric content attributes get reflected into numeric IDL types, so you should probably either make this an `unsigned long` or add a definition for `unsigned long long` reflection into HTML. 32-bit numbers of milliseconds give us 49 days of expiration, which cuts it a little close to the current maximum expiration times, so I lean toward sending a change to HTML.

@domenic, before I make @johnivdel do extra work, does this match your sense of the right thing to do?

Assuming this is the right thing to do, I would commit this as an `unsigned long long` before you get the change into HTML, even though that's formally broken, because it's unlikely to cause implementation divergence.

jyasskin (approved):

index.bs
johnivdel: I had changed this to a DOMString to align with `attributionsourceeventid`, which also represents a 64 bit unsigned integer. 

Using unsigned long long runs into `Number.MAX_SAFE_INTEGER` as noted on #123, and so we decided to accept the value as a string for simplicity. Although perhaps I am misunderstanding the relationship between JavaScript & WebIDL here.

index.bs
domenic: In general monkeypatch specs need a lot of similar references to non-exported definitions; that's probably the best way.

index.bs
domenic: Interesting. I tend to agree that normally things reflect as numbers, but if there is a legitimate case where the number would be outside the max-safe-integer range, then either strings or bigints would be correct. Bigints would be unprecedented but perhaps better.

On the other hand, maybe it's OK for the number to stay inside the 2^53 range, in which case creating unsigned long long reflection would make sense. Hmm.

jyasskin: 2<sup>53</sup> milliseconds is 285,427 years, so it's probably fine for an expiration time. It's not as good for an event ID, but those are IDs rather than things the Javascript might do math on, which is a plausible reason to keep them different.

This could be a good argument for giving the reflection machinery a knob similar to ["limited to only non-negative numbers"](https://html.spec.whatwg.org/multipage/common-dom-interfaces.html#limited-to-only-non-negative-numbers) to control whether the reflection throws if it's given a non-safe number.

johnivdel: Agree w/ @jyasskin that 2^53 should be fine. 

In this case, both an `unsigned long long` and a `long long` "limited to only non-negative integers" would support the same range of values given the 2^53 limit. Is using the unsigned type still preferable in that case?

jyasskin: I don't understand why we have "`long` limited to only non-negative integers", so I don't really have a preference between you using `long long` limited… or `unsigned long long`.

domenic: As to @jyasskin's question, I suspect that most HTML interfaces were developed before Web IDL really existed, and so a lot of things used `long` semantics in the implementation, with custom checks.

I think defining "`long long` limited only to non-negative numbers" would probably be best, both for consistency and for web developers. In particular, if we had `unsigned long long` reflection, then

```js
el.attributionExpiry = -1;
```

would behave the same as `el.attributionExpiry = 2**64 - 1`. Whereas if you did "`long long` limited only to non-negative numbers", it would throw an "`IndexSizeError`" `DOMException`.

index.bs
```diff
@@ -49,29 +49,45 @@ conversion domain.
...
+
+Extend the <{a}> element's <a spec=html>DOM interface</a> to include the following interface:
 
 <pre class="idl">
 partial interface HTMLAnchorElement {
-    [CEReactions, Reflect] attribute DOMString attributiondestination;
-    [CEReactions, Reflect] attribute DOMString attributionsourceeventid;
-    [CEReactions, Reflect] attribute DOMString attributionreportto;
-    [CEReactions, Reflect] attribute unsigned long long attributionexpiry;
+    [CEReactions] attribute DOMString attributiondestination;
```
domenic, line 67: IDL attributes use camel case, so e.g. `attributionDestination`, `attributionSourceEventId`, `attributionReportTo`, `attributionExpiry`.

jyasskin: Oops, thanks! For John, the rules are in https://w3ctag.github.io/design-principles/#casing-rules.

index.bs
johnivdel: That makes sense. I've updated the IDL to long long.

What's the best way to move forward on spec'ing the long long reflection? Given that it would be unused does it make more sense to add it to this spec? Or just file a PR against HTML that would be necessary if these monkeypatches were resolved?

domenic: IMO adding it to this spec, with a note that you intend to upstream it to HTML along with the rest of the monkeypatches, makes sense.

PR #142, opened 2021-05-07
Update AGGREGATE.md with more details

csharrison:
Adds more details about 
- Integrating the worklet design
- Forming aggregation keys
- Privacy budgeting
- High level details about MPC protocol and example query model

### DIFF
--- a/AGGREGATE.md
+++ b/AGGREGATE.md
@@ -46,7 +46,7 @@ This API builds on the same mechanisms described in the [Multi-browser aggregati
 
 In particular, we introduce experimental mechanisms including:
 
-*   **Bring your own aggregation servers:** anyone can receive reports for aggregation during an initial experimentation perio.
+*   **Bring your own aggregation servers:** anyone can receive reports for aggregation during an initial experimentation period.
 *   **Multi-party computation (MPC) optional:** Long-term,  we expect servers will need to process data exclusively with [secure multi-party computation](https://en.wikipedia.org/wiki/Secure_multi-party_computation) (or technology that achieves similar goals), but for an initial experimentation period this is optional.
 
 ## Goals
--- a/AGGREGATE.md
+++ b/AGGREGATE.md
@@ -341,7 +341,7 @@ In general, it is undesirable to require JavaScript on the attribution trigger r
 
 ### Custom attribution models
 
-The worklet based scheme possibly allows for more flexible attribution options, including specifying partial “credit” for multiple previous impressions that would provide value to advertisers that are interested in attribution models other than last-touch.
+The worklet based scheme possibly allows for more flexible attribution options, including specifying partial “credit” for multiple previous attribution sources that would provide value to advertisers that are interested in attribution models other than last-touch.
 
 ### “Count” vs. “value” histograms
 
--- a/AGGREGATE.md
+++ b/AGGREGATE.md
@@ -1,180 +1,359 @@
-# Conversion Measurement with Aggregation Explainer
-
-# Introduction
-
-This document is an explainer for extensions to our existing [event-level conversion measurement API](https://github.com/csharrison/conversion-measurement-api) explainer. The intention is to provide mechanisms for information about conversions to be reported in a way that reporting endpoints can only learn _aggregate_ data, not any data associated with a particular click or user.
-This is designed to work in along side our event-level conversion measurement API as they address different use cases.
-
-
-In this way, we can satisfy use cases for which an event-level API would reveal too much private information.
-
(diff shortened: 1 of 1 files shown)


AGGREGATE.md
```diff
@@ -1,180 +1,359 @@
...
+This API builds on the same mechanisms described in the [Multi-browser aggregation service](https://github.com/WICG/conversion-measurement-api/blob/main/SERVICE.md) explainer.
 
-### Better accuracy
-The aggregate conversion measurement API should give high-fidelity measurement of conversions. The reports it generates should produce more faithful results than, for example, the noisy conversion data in the event-level API.
+In particular, we introduce experimental mechanisms including:
 
-In the event-level API, impressions are also limited to converting a small number of times. The aggregate conversion measurement API should allow advertisers to get more accurate counts of how many conversions there were for a campaign.
+*   **Bring your own aggregation servers:** anyone can receive reports for aggregation during an initial experimentation perio.
```
davidvancleve, line 49: driveby: "period"

AGGREGATE.md
```diff
@@ -1,180 +1,359 @@
...
+Additionally, there are some privacy benefits of smaller domains for aggregation, in that it is helpful if it is difficult to map individual users (or user IDs crossed with other information) to individual buckets, especially in higher epsilon DP regimes.
+
+
+### Declarative / HTTP-based aggregation triggering
+
+In general, it is undesirable to require JavaScript on the attribution trigger registration. Many existing conversion use-cases use HTTP to register conversions. It is challenging to densely encode an aggregation key space (e.g. in 32 bits) with two disjoint identifiers (source and destination), but it would be good to support this case if possible.
+
+### Custom attribution models
+
+The worklet based scheme possibly allows for more flexible attribution options, including specifying partial “credit” for multiple previous impressions that would provide value to advertisers that are interested in attribution models other than last-touch.
```
johnivdel, line 344:
```suggestion
The worklet based scheme possibly allows for more flexible attribution options, including specifying partial “credit” for multiple previous attribution sources that would provide value to advertisers that are interested in attribution models other than last-touch.
```

AGGREGATE.md
```diff
@@ -1,180 +1,359 @@
...
-agg-multi-touch-model=<linear,last-clicked,first-clicked,...>
+```json
+{
+  "source_site": "https://publisher.com",
+  "attribution_destination": "https://advertiser.com",
+  "scheduled_report_time": <timestamp in msec>,
+  "aggregation_service_payloads": [
+    {
+      "origin": "https://helper1.com",
+      "payload": <base64 encoded encrypted data>
```
johnivdel, line 169: Should this have quotes? Presumably this is encoded as a string? 

ditto below

AGGREGATE.md
csharrison: done

AGGREGATE.md
```diff
@@ -1,180 +1,359 @@
...
+  // Don't expand a node if it doesn't reach the threshold.
+  expansion_threshold_per_prefix: [5, 5, 5]
+}
 ```
-local-dedupe-key=<key>
+
+The output (after merging results from two servers) could look like this, where the aggregate data is returned for all queried prefixes.
+
+
+```json
```
johnivdel, line 279: We should fix the syntax highlighting here and use strings as the keys.

Note sure what to do about the comment

AGGREGATE.md
```diff
@@ -1,180 +1,359 @@
...
+3-bit prefix counts: 99   0    1   2      200   200   1  1
 ```
 
-Once an impression / conversion is registered, the browser should interact with the signing endpoint, which should provide authentication tokens that will attach to conversion reports and be verified by the helper servers. Helper servers can (optionally) discard reports that are not signed. This is explored in more detail [here](SERVICE.md#authenticating-inputs).
 
-### Many-per-click / Once-per-click
+We could have a query configuration asking for counts at every prefix level, with some strategy for iterating down and pruning the tree.
+
+
+```json
```
johnivdel, line 263: ditto to below on highlighting

Could also consider dropping json for this one

johnivdel (approved):
lgtm % comments

AGGREGATE.md
csharrison: done

AGGREGATE.md
csharrison: done, removed json syntax highlighting

AGGREGATE.md
csharrison: done

PR #143, opened 2021-05-13
Add privacy_budget key to histogram result

csharrison:
Even though the example does not use noise, we should reveal the noise share added to each hierarchical result


csharrison:
Hm maybe this isn't the right naming we should be going for. What about `privacy_budget_consumed`?

johnivdel:
This would just mirror the query configuration correct? Or could these be different in the output?

csharrison:
Not relevant anymore

PR #149, opened 2021-05-20
Return a promise from registerAttributionSource method

### LINKED ISSUE #148: Improved error reporting for registerAttributionSource method
https://github.com/WICG/conversion-measurement-api/blob/main/event_attribution_reporting.md#registering-attribution-sources-with-javascript currently defines the `bool registerAttributionSource(AttributionSourceParams params)` method, which it says returns `false` if the parameters are invalid.

It would probably be more useful to throw a specific exception instead of returning a `bool`.

We could also consider changing this to a promise-based API, to cover the scenario in which the browser doesn't know whether a source will be logged when the API is invoked, e.g. because of storage limits.

apasel422:
Fixes #148.

### DIFF
--- a/event_attribution_reporting.md
+++ b/event_attribution_reporting.md
@@ -50,7 +50,7 @@ dictionary AttributionSourceParams {
 }
 ```
 
-When invoked the browser will directly add the specified source to storage, and resolve or reject the promise based on whether the browser was successful in parsing the params and storing the source.
+When invoked the browser will directly add the specified source to storage, and resolve or reject the promise based on whether the browser was successful in parsing the params and storing the source. Due to privacy reasons, some failure states (e.g. those that depend on cross-site information) may not show up as rejections.
 
 
 ### Different Classes of Attribution Sources
--- a/event_attribution_reporting.md
+++ b/event_attribution_reporting.md
@@ -35,7 +35,7 @@ The browser will expose a new interface:
 
 ```
 [Exposed=Window] interface AttributionReporting {
-  bool registerAttributionSource(AttributionSourceParams params);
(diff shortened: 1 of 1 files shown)


event_attribution_reporting.md
```diff
@@ -35,7 +35,7 @@ The browser will expose a new interface:
 
 ```
 [Exposed=Window] interface AttributionReporting {
-  bool registerAttributionSource(AttributionSourceParams params);
+  Promise<void> registerAttributionSource(AttributionSourceParams params);
```
apasel422, line 38: I'm not sure if this should be spelled `Promise<void>` or `Promise<undefined>`.

csharrison: I think void has been replaced by undefined in WebIDL from what I can see
https://heycam.github.io/webidl/#idl-undefined

apasel422: Done.

event_attribution_reporting.md
```diff
@@ -50,7 +50,7 @@ dictionary AttributionSourceParams {
 }
 ```
 
-When invoked the browser will directly add the specified source to storage, and return whether the browser was successful in parsing the params.
+When invoked the browser will directly add the specified source to storage, and resolve or reject the promise based on whether the browser was successful in parsing the params and storing the source.
```
apasel422, line 53: We should think about whether failure to store the source leaks unintended information back to the caller.

csharrison: We should mention in the explainer that, due to privacy reasons, some failure states (those that depend on cross site information) may not show up as rejections.

apasel422: Done.

csharrison (approved):

PR #152, opened 2021-05-25
Spec attribution caches

johnivdel:
This is a copy of #141 which was reverted on #151

***
<a href="https://pr-preview.s3.amazonaws.com/WICG/conversion-measurement-api/pull/152.html" title="Last updated on May 28, 2021, 12:00 AM UTC (00d9689)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/conversion-measurement-api/152/5c456e6...00d9689.html" title="Last updated on May 28, 2021, 12:00 AM UTC (00d9689)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -131,15 +131,16 @@ An attribution report is a [=struct=] with the following items:
 
 # Storage # {#storage}
 
-A user agent MUST provide access to a set of <dfn>attribution caches</dfn> which are shared among all
-[=environment settings objects=] and consist of:
+A user agent holds an <dfn>attribution source cache</dfn>, which is an [=ordered set=] of [=attribution sources=].
 
-1. an <dfn>attribution source cache</dfn>, which is an [=ordered set=] of [=attribution sources=].
-1. an <dfn>attribution report cache</dfn>, which is an [=ordered set=] of [=attribution reports=].
+A user agent holds an <dfn>attribution report cache</dfn>, which is an [=ordered set=] of [=attribution reports=].
+
+The above caches are collectively known as the <dfn>attribution caches</dfn>. The [=attribution caches=] are
+shared among all [=environment settings objects=].
 
 User agents SHOULD place limits on the maximum [=list/size=] of the [=attribution caches=].
 
-Issue: This would ideally use <a spec=storage>storage bottles</a> to provide access to the attribution caches.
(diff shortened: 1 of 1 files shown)


johnivdel:
@jyasskin Could you please take a look? (The spec fell off my radar for a bit)

index.bs
```diff
@@ -280,3 +293,8 @@ TODO
 
 # Privacy consideration # {#privacy-considerations}
 TODO
+
+<h3 id="clearing-attribution-storage">Clearing attribution storage</h3>
+
+A user agent's [=attribution caches=] contain data about a user's web activity. If a user agent gives users the ability to clear 
+their browsing history and browsing related storage, the user agent MUST also remove the related items within the [=attribution caches=].
```
jyasskin, line 300: I think we talked about being more precise about this clearing. e.g.

```suggestion
A user agent's [=attribution caches=] contain data about a user's web activity. When a user agent clears an origin's storage, it MUST also remove entries in the [=attribution caches=] whose [=source origin=], [=attribution destination=], [=reporting endpoint=], or [=trigger origin=] is the [=same origin/same=] as the cleared origin.
```

johnivdel: Do you think it's worth differentiating between origin storage and browsing history here?

For example, in the Chrome implementation when a user deletes a site from the browsing history (but leaves its storage along), the browser also clears the entries for that sites.

jyasskin: I don't know of anywhere in the rest of the platform that acknowledges that browsers store history. It won't hurt to include it here. Do you clear the attribution cache entries any time any entry with a given origin is removed from history, or just when an entire site is removed?

johnivdel: The Chrome implementation does both. If history is deleted within a time range, the attribution cache entries matching origins in that time range get deleted.

If a site is completely removed from history (meaning there are no longer any occurrences of it), all of the entries which match the origin are deleted.

This approach is relatively aggressive, and I could imagine implementations would want to use slightly different strategies. The thinking is that this does not result in any breakage for users, so we can be more permissive in what we delete. Maybe there is a more flexible way to word this, at least dropping the MUST for the browsing history part?

jyasskin: Or a "MAY drop entries at other times", mentioning browsing history as an example of the other times?

jyasskin (approved):

PR #156, opened 2021-06-03
Add monkeypatches for navigation logic

johnivdel:
This PR adds logic for:
 - passing an attribution source on a navigation originating from a link click
 - validating that the attribution source attached to a navigation declares the correct destination
 - logic for adding a source to the attribution source cache

***
<a href="https://pr-preview.s3.amazonaws.com/WICG/conversion-measurement-api/pull/156.html" title="Last updated on Jun 9, 2021, 8:47 PM UTC (f912bd2)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/conversion-measurement-api/156/0cda822...f912bd2.html" title="Last updated on Jun 9, 2021, 8:47 PM UTC (f912bd2)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -236,7 +236,7 @@ An attribution source is a [=struct=] with the following items:
 : <dfn>event id</dfn>
 :: A non-negative 64-bit integer.
 : <dfn>attribution destination</dfn>
-:: An [=url/origin=].
+:: A [=site=].
 : <dfn>reporting endpoint</dfn>
 :: An [=url/origin=].
 : <dfn>expiry</dfn>
@@ -358,8 +358,8 @@ To <dfn>process an attribution source</dfn> given an [=attribution source=] |sou
 
 1. Let |cache| be the user agent's [=attribution source cache=].
 1. [=list/Remove=] all entries in |cache| where all of the following are true:
-    * the entry's [=attribution source/attribution destination=] matches |source|'s [=attribution source/attribution destination=].
-    * the entry's [=attribution source/reporting endpoint=] matches |source|'s [=attribution source/reporting endpoint=].
+    * the entry's [=attribution source/attribution destination=] and |source|'s [=attribution source/attribution destination=] are equal.
+    * the entry's [=attribution source/reporting endpoint=] is [=same origin=] with |source|'s [=attribution source/reporting endpoint=].
     * the entry's [=attribution source/number of reports=] value is greater than 0 .
 
     Note: This causes the user agent to favor triggering newer [=attribution sources=] over sources that have already been triggered.
--- a/index.bs
+++ b/index.bs
@@ -151,9 +151,23 @@ to set the [=navigation params/attribution source=] of |navigationParams| to |at
 
 In the case where
 
+> If resource is a request whose url's scheme is "javascript"
+
+modify the substep 
+
+> 4. Let navigationParams be a new navigation params whose request is resource, 
+>     ...
+
+to set the [=navigation params/attribution source=] of |navigationParams| to |attributionSource|.
+
+In the case where
+
 > If resource is a request whose url's scheme is a fetch scheme
 
 modify the substep to pass |attributionSource| to the <a spec="html">process a navigate fetch</a> algorithm.
+
+Note: The final case, were the request is not a javascript or fetch scheme, does not need to be handled
+ as it will not result in the navigation of a top-level browsing context.
 
 <h4 id="monkeypatch-navigate-fetch">Process a navigate fetch</h4>
 
@@ -348,13 +362,14 @@ To <dfn>maybe process an attribution source</dfn> given a <a spec="HTML">navigat
 To <dfn>process an attribution source</dfn> given an [=attribution source=] |source|:
 
 1. Let |cache| be the user agent's [=attribution source cache=].
-1. [=list/Remove=] all entries in |cache| where any of:
+1. [=list/Remove=] all entries in |cache| where all of the following are true:
     1. the entry's [=attribution source/attribution destination=] matches |source|'s [=attribution source/attribution destination=].
     1. the entry's [=attribution source/reporting endpoint=] matches |source|'s [=attribution source/reporting endpoint=].
-    1. the entry's [=attribution source/num reports=] value is greater than 0 .
+    1. the entry's [=attribution source/number of reports=] value is greater than 0 .
 
     Note: This causes the user agent to favor triggering newer [=attribution sources=] over sources that have already been triggered.
-1. [=set/Append=] |source| to |cache|.
+1. [=list/Remove=] all entries in |cache| where the entry's [=attribution source/expiry=] value is less than the current time.
+1. If the [=list/size=] is less than an implementation-defined limit, [=set/Append=] |source| to |cache|.
 
 <h3 algorithm id="attribution-trigger-creation">Creating an attribution trigger</h3>
 
--- a/index.bs
+++ b/index.bs
@@ -118,7 +118,7 @@ of time in milliseconds the attribution source should be considered for reportin
 <h3 id="monkeypatch-navigation">Navigation</h3>
 
 This section ensures that an [=attribution source=] associated with a navigation
-results in a top-level navigation whose final url is [=same site=]
+results in a top-level navigation whose final url is [=same site=] with the
 [=attribution source/attribution destination=].
 
 <h4 id="monkeypatch-navigation-params">Navigation Params</h4>
@@ -200,7 +200,7 @@ In <a spec="html">follow the hyperlink</a> after
 add the following steps:
 
 1. Let <var>attributionSource</var> be null
-1. If |subject| is an <{a}> element, set attributionSource to the result of running [=obtain an attribution source=] with |subject|.
+1. If |subject| is an <{a}> element, set |attributionSource| to the result of running [=obtain an attribution source=] with |subject|.
 
 Modify the step:
 
--- a/index.bs
+++ b/index.bs
@@ -358,9 +358,9 @@ To <dfn>process an attribution source</dfn> given an [=attribution source=] |sou
 
 1. Let |cache| be the user agent's [=attribution source cache=].
 1. [=list/Remove=] all entries in |cache| where all of the following are true:
-    1. the entry's [=attribution source/attribution destination=] matches |source|'s [=attribution source/attribution destination=].
-    1. the entry's [=attribution source/reporting endpoint=] matches |source|'s [=attribution source/reporting endpoint=].
-    1. the entry's [=attribution source/number of reports=] value is greater than 0 .
+    * the entry's [=attribution source/attribution destination=] matches |source|'s [=attribution source/attribution destination=].
+    * the entry's [=attribution source/reporting endpoint=] matches |source|'s [=attribution source/reporting endpoint=].
+    * the entry's [=attribution source/number of reports=] value is greater than 0 .
 
     Note: This causes the user agent to favor triggering newer [=attribution sources=] over sources that have already been triggered.
 1. [=list/Remove=] all entries in |cache| where the entry's [=attribution source/expiry=] value is less than the current time.
--- a/index.bs
+++ b/index.bs
@@ -117,7 +117,7 @@ of time in milliseconds the attribution source should be considered for reportin
 
 <h3 id="monkeypatch-navigation">Navigation</h3>
 
-The user agent must verify that an [=attribution source=] associated with a navigation
+This section ensures that an [=attribution source=] associated with a navigation
 results in a top-level navigation whose final url is [=same site=]
 [=attribution source/attribution destination=].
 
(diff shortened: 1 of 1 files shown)


johnivdel:
@jyasskin Please take a look thanks!

index.bs
```diff
@@ -181,6 +275,9 @@ shared among all [=environment settings objects=].
 
 User agents SHOULD place limits on the maximum [=list/size=] of the [=attribution caches=].
 
+User agents SHOULD periodically remove entries from the [=attribution source cache=] where the entry's
+[=attribution source/expiry=] is in the past.
```
jyasskin, line 279: Good catch, but this makes me realize that some of this actually needs to be expressed in the algorithms that add entries to the attribution caches. So, below at the "Add |source| to cache." step, you should say something like "If the size of |cache| is less than an implementation-defined limit ..."

You could have it remove expired entries in that algorithm too, or this "periodically" is probably ok (with a different meaning).

index.bs
```diff
@@ -237,7 +334,28 @@ To <dfn>obtain an attribution source</dfn> from an <{a}> element |anchor|:
...
 
-Issue: Need to spec how to store the attribution source.
+<h3 id="processing-an-attribution-source">Processing an attribution source</h3>
+
+To <dfn>maybe process an attribution source</dfn> given a <a spec="HTML">navigation params</a>
+|navigationParams| and [=browsing context=] |browsingContext|, run the following steps:
+1. If |browsingContext| is not a <a spec="html">top-level browsing context</a>, return.
+1. Let <var>attributionSource</var> be |navigationParams|'s <a lt="navigation params attribution source">attribution source</a>.
+1. If |attributionSource| is null, return.
+1. Let |attributionSource|'s [=attribution source/attribution destination=] is not [=same site=] to |navigationParams|'s
```
jyasskin, line 344:
```suggestion
1. If |attributionSource|'s [=attribution source/attribution destination=] is not [=same site=] to |navigationParams|'s 
```

index.bs
```diff
@@ -237,7 +334,28 @@ To <dfn>obtain an attribution source</dfn> from an <{a}> element |anchor|:
...
+
+1. Let |cache| be the user agent's [=attribution source cache=].
+1. Let |sources| be the all entries in |cache| where:
+    1. the entry's [=attribution source/attribution destination=] matches |source|'s [=attribution source/attribution destination=].
+    1. the entry's [=attribution source/reporting endpoint=] matches |source|'s [=attribution source/reporting endpoint=].
+    1. the entry's [=attribution source/num reports=] value is greater than 0 .
+1. Remove |sources| from |cache|.
+
+    Note: This causes the user agent to favor triggering newer [=attribution sources=] over sources that have already been triggered.
+1. Add |source| to cache.
```
jyasskin, line 358:
```suggestion
1. [=set/Append=] |source| to |cache|.
```

Or prepend if you prefer.

index.bs
```diff
@@ -237,7 +334,28 @@ To <dfn>obtain an attribution source</dfn> from an <{a}> element |anchor|:
...
+1. [=Queue a task=] to [=process an attribution source=] with |attributionSource|.
+
+To <dfn>process an attribution source</dfn> given an [=attribution source=] |source|:
+
+1. Let |cache| be the user agent's [=attribution source cache=].
+1. Let |sources| be the all entries in |cache| where:
+    1. the entry's [=attribution source/attribution destination=] matches |source|'s [=attribution source/attribution destination=].
+    1. the entry's [=attribution source/reporting endpoint=] matches |source|'s [=attribution source/reporting endpoint=].
+    1. the entry's [=attribution source/num reports=] value is greater than 0 .
+1. Remove |sources| from |cache|.
```
jyasskin, line 355: Since you don't use |sources| later, I'd use something like:

```suggestion
1. [=list/Remove=] all entries in |cache| where any of:
    1. the entry's [=attribution source/attribution destination=] matches |source|'s [=attribution source/attribution destination=].
    1. the entry's [=attribution source/reporting endpoint=] matches |source|'s [=attribution source/reporting endpoint=].
    1. the entry's [=attribution source/num reports=] value is greater than 0 .
```

index.bs
```diff
@@ -115,8 +115,100 @@ The <dfn for="a" element-attr>attributionreportto</dfn> attribute optionally dec
 The <dfn for="a" element-attr>attributionexpiry</dfn> attribute optionally defines the amount
 of time in milliseconds the attribution source should be considered for reporting.
 
-Issue: Need monkey patches passing attribution source in navigation, and a mechanism
-for validating the resulting document matches the attributiondestination.
+<h3 id="monkeypatch-navigation">Navigation</h3>
+
+The user agent must verify that an [=attribution source=] associated with a navigation
```
jyasskin, line 120: Don't use "must" in introductory paragraphs like this, since that introduces a normative requirement. You could say

```suggestion
This section ensures that an [=attribution source=] associated with a navigation
```

index.bs
```diff
@@ -115,8 +115,100 @@ The <dfn for="a" element-attr>attributionreportto</dfn> attribute optionally dec
...
+<h4 id="monkeypatch-navigation-params">Navigation Params</h4>
+
+A <a spec="HTML">navigation params</a> struct has an item:
+
+: <dfn lt="navigation params attribution source" id="navigation-params-attribution-source">attribution source</dfn>
+:: an optional [=attribution source=] declared when initiating a navigation
+
+<h4 id="monkeypatch-navigate-algorithm">Navigate algorithm</h4>
+
+Modify the <a spec="html">navigate</a> algorithm to accept a new parameter
```
jyasskin, line 133:
```suggestion
Modify the <a spec="html">navigate</a> algorithm to accept a new optional parameter
```

? (and then "defaulting to null" if that's what you do for the navigation params item.)

index.bs
```diff
@@ -115,8 +115,100 @@ The <dfn for="a" element-attr>attributionreportto</dfn> attribute optionally dec
...
+The user agent must verify that an [=attribution source=] associated with a navigation
+results in a top-level navigation whose final url is [=same site=]
+[=attribution source/attribution destination=].
+
+<h4 id="monkeypatch-navigation-params">Navigation Params</h4>
+
+A <a spec="HTML">navigation params</a> struct has an item:
+
+: <dfn lt="navigation params attribution source" id="navigation-params-attribution-source">attribution source</dfn>
+:: an optional [=attribution source=] declared when initiating a navigation
```
jyasskin, line 129: It looks like the other items are "null or X" instead of "an optional X", so try to follow that here.

index.bs
```diff
@@ -115,8 +115,100 @@ The <dfn for="a" element-attr>attributionreportto</dfn> attribute optionally dec
...
+
+The user agent must verify that an [=attribution source=] associated with a navigation
+results in a top-level navigation whose final url is [=same site=]
+[=attribution source/attribution destination=].
+
+<h4 id="monkeypatch-navigation-params">Navigation Params</h4>
+
+A <a spec="HTML">navigation params</a> struct has an item:
+
+: <dfn lt="navigation params attribution source" id="navigation-params-attribution-source">attribution source</dfn>
```
jyasskin, line 128:
```suggestion
: <dfn for="navigation params">attribution source</dfn>
```

I think.

jyasskin (commented):
I didn't quite get all the way through. Will try again tomorrow.

index.bs
```diff
@@ -115,8 +115,100 @@ The <dfn for="a" element-attr>attributionreportto</dfn> attribute optionally dec
...
+
+Modify the <a spec="html">navigate</a> algorithm to accept a new optional parameter
+<var>attributionSource</var> of type [=attribution source=] defaulting to null.
+
+In <a spec="html">navigate</a>, within step
+
+> 19. This is the step that attempts to obtain resource, if necessary. Jump to the first appropriate substep:
+>     ...
+
+in the case where
```
jyasskin, line 141: Can you write something about why it's correct to drop the attribution source for the other cases?

index.bs
```diff
@@ -115,8 +115,100 @@ The <dfn for="a" element-attr>attributionreportto</dfn> attribute optionally dec
 The <dfn for="a" element-attr>attributionexpiry</dfn> attribute optionally defines the amount
 of time in milliseconds the attribution source should be considered for reporting.
 
-Issue: Need monkey patches passing attribution source in navigation, and a mechanism
-for validating the resulting document matches the attributiondestination.
+<h3 id="monkeypatch-navigation">Navigation</h3>
+
+This section ensures that an [=attribution source=] associated with a navigation
+results in a top-level navigation whose final url is [=same site=]
+[=attribution source/attribution destination=].
```
jyasskin, line 122:
```suggestion
results in a top-level navigation whose final url is [=same site=] with the
[=attribution source/attribution destination=].
```
I think.

index.bs
```diff
@@ -115,8 +115,100 @@ The <dfn for="a" element-attr>attributionreportto</dfn> attribute optionally dec
...
+<a spec="html">navigate</a> algorithm. 
+
+In <a spec="html">follow the hyperlink</a> after
+
+> 14. Let historyHandling be "replace" if windowType is not "existing or none"; otherwise, "default".
+
+add the following steps:
+
+1. Let <var>attributionSource</var> be null
+1. If |subject| is an <{a}> element, set attributionSource to the result of running [=obtain an attribution source=] with |subject|.
```
jyasskin, line 203:
```suggestion
1. If |subject| is an <{a}> element, set |attributionSource| to the result of running [=obtain an attribution source=] with |subject|.
```

index.bs
```diff
@@ -237,7 +334,27 @@ To <dfn>obtain an attribution source</dfn> from an <{a}> element |anchor|:
...
+To <dfn>process an attribution source</dfn> given an [=attribution source=] |source|:
+
+1. Let |cache| be the user agent's [=attribution source cache=].
+1. [=list/Remove=] all entries in |cache| where any of:
+    1. the entry's [=attribution source/attribution destination=] matches |source|'s [=attribution source/attribution destination=].
+    1. the entry's [=attribution source/reporting endpoint=] matches |source|'s [=attribution source/reporting endpoint=].
+    1. the entry's [=attribution source/num reports=] value is greater than 0 .
+
+    Note: This causes the user agent to favor triggering newer [=attribution sources=] over sources that have already been triggered.
+1. [=set/Append=] |source| to |cache|.
```
jyasskin, line 357: Remember to implement the cache size limit here.

jyasskin (approved):
Just a couple nits left. Thanks!

index.bs
johnivdel: Done. Upon looking at this further I don't think there is a good reason to not handle javascript schemes, so I added that and a note on the other case.

index.bs
```diff
@@ -348,13 +357,14 @@ To <dfn>maybe process an attribution source</dfn> given a <a spec="HTML">navigat
 To <dfn>process an attribution source</dfn> given an [=attribution source=] |source|:
 
 1. Let |cache| be the user agent's [=attribution source cache=].
-1. [=list/Remove=] all entries in |cache| where any of:
+1. [=list/Remove=] all entries in |cache| where all of the following are true:
     1. the entry's [=attribution source/attribution destination=] matches |source|'s [=attribution source/attribution destination=].
     1. the entry's [=attribution source/reporting endpoint=] matches |source|'s [=attribution source/reporting endpoint=].
```
jyasskin: Sorry for missing this earlier: "matches" is a bad word to use in a spec unless it's linked to a definition, because it's unclear what equivalence relation it implies. You probably mean https://html.spec.whatwg.org/multipage/origin.html#same-origin here?

index.bs
```diff
@@ -237,7 +343,28 @@ To <dfn>obtain an attribution source</dfn> from an <{a}> element |anchor|:
...
+    <a href="https://html.spec.whatwg.org/multipage/browsing-the-web.html#navigation-params-origin">origin</a>, return.
+1. [=Queue a task=] to [=process an attribution source=] with |attributionSource|.
+
+To <dfn>process an attribution source</dfn> given an [=attribution source=] |source|:
+
+1. Let |cache| be the user agent's [=attribution source cache=].
+1. [=list/Remove=] all entries in |cache| where all of the following are true:
+    1. the entry's [=attribution source/attribution destination=] matches |source|'s [=attribution source/attribution destination=].
+    1. the entry's [=attribution source/reporting endpoint=] matches |source|'s [=attribution source/reporting endpoint=].
+    1. the entry's [=attribution source/number of reports=] value is greater than 0 .
```
jyasskin, line 363: Oh, and these aren't ordered so

```suggestion
    * the entry's [=attribution source/attribution destination=] matches |source|'s [=attribution source/attribution destination=].
    * the entry's [=attribution source/reporting endpoint=] matches |source|'s [=attribution source/reporting endpoint=].
    * the entry's [=attribution source/number of reports=] value is greater than 0 .
```

index.bs
johnivdel: Hmmm, for reportingendpoint same origin is correct.

However, attribution destination is actually a site (https://html.spec.whatwg.org/multipage/origin.html#site). The attribution source struct currently declares this as an origin which is wrong.

Would "the entries ... is equivalent to ..." work?

jyasskin: No, "equivalent to" also has multiple definitions. For sites, I think you can use "are the same" or "are equal". (You'd use https://html.spec.whatwg.org/multipage/origin.html#same-site if it were an origin, but I don't see a defined term for sites, and their structure is simple enough that equality is right. (And hosts, the only complicated part, have https://url.spec.whatwg.org/#concept-host-equals that just uses "A is B".)

johnivdel: Got it, thanks for the explanation!

PR #159, opened 2021-06-11
Spec declaration of attributionsourcepriority

johnivdel:
This priority is necessary to spec algorithms for handling conversion events, and the browser attribution logic

***
<a href="https://pr-preview.s3.amazonaws.com/WICG/conversion-measurement-api/pull/159.html" title="Last updated on Jun 23, 2021, 4:38 PM UTC (bb20ccd)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/conversion-measurement-api/159/77abb8d...bb20ccd.html" title="Last updated on Jun 23, 2021, 4:38 PM UTC (bb20ccd)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -82,7 +82,7 @@ Add the following <a spec=html>content attributes</a> to the <{a}> element:
 : <{a/attributionexpiry}>
 :: Length of time the attribution souce is valid
 : <{a/attributionsourcepriority}>
-:: The priority of this source relative to other sources for attribution triggers
+:: The priority of this source relative to other sources when triggering attribution
 
 Extend the <{a}> element's <a spec=html>DOM interface</a> to include the following interface:
 
--- a/index.bs
+++ b/index.bs
@@ -81,6 +81,8 @@ Add the following <a spec=html>content attributes</a> to the <{a}> element:
 :: [=url/origin=] to receive attribution reports
 : <{a/attributionexpiry}>
 :: Length of time the attribution souce is valid
+: <{a/attributionsourcepriority}>
+:: The priority of this source relative to other sources for attribution triggers
 
 Extend the <{a}> element's <a spec=html>DOM interface</a> to include the following interface:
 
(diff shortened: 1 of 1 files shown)


johnivdel:
@jyasskin please take a look thanks!

index.bs
```diff
@@ -241,6 +247,8 @@ An attribution source is a [=struct=] with the following items:
 :: An [=url/origin=].
 : <dfn>expiry</dfn>
 :: A point in time.
+: <dfn>priority</dfn>
+:: A 64-bit integer.
```
jyasskin, line 251: Where is this going to be used? In a future PR for this spec?

index.bs
```diff
@@ -115,6 +118,9 @@ The <dfn for="a" element-attr>attributionreportto</dfn> attribute optionally dec
 The <dfn for="a" element-attr>attributionexpiry</dfn> attribute optionally defines the amount
 of time in milliseconds the attribution source should be considered for reporting.
 
+The <dfn for="a" element-attr>attributionsourcepriority</dfn> attribute optionally defines the
+priority of a source relative to other sources when triggering attribution.
```
jyasskin, line 122: We should say something here about the meanings of the possible values. Will an attribution with a higher priority value be picked instead of one with a lower value? What's the default? Do different mutually-trusting parties need to agree at all about their prioritization scheme (including different libraries used by the same company), and if so, what kind of agreement do we suggest?

jyasskin (changes requested):

index.bs
johnivdel: Yes, this is a pre-requisite for spec'ing the conversion/attribution algorithms

index.bs
johnivdel: Added a bit more color. The actual behavior is defined by the attribution triggering algorithm which is why I left out more detail here, but having more detail next to the definition seems valuable.

Priorities are interpreted within <reporting endpoint, attribution destination> scopes, so generally third parties will not need to agree on this. If there are libraries using the same reporting endpoint, they will need to agree.

jyasskin (approved):

PR #189, opened 2021-07-21
Remove obsolete mention of multiple reports per trigger from click explainer

apasel422:


event_attribution_reporting_clicks.md
```diff
@@ -229,10 +229,12 @@ from 0-5 (~2.6 bits of information)
 
 When the browser receives a attribution trigger redirect on a URL matching
 the `attributiondestination` eTLD+1, it looks up all sources in storage that
-match <`attributionreportto`, `attributiondestination`>.
+match <`attributionreportto`, `attributiondestination`> and picks the one with
+the greatest `attributionsourcepriority`. If multiple sources have the greatest
+`attributionsourcepriority`, the browser picks the one that was stored at the
```
johnivdel, line 234: I believe this should be most recent time (the default should be last touch)

event_attribution_reporting_clicks.md
```diff
@@ -246,17 +248,6 @@ the browser will compare the priority of the new report with the priorities of t
 If the new report has the lowest priority, it will be ignored. Otherwise, the browser will
 delete the scheduled report with the lowest priority and schedule the new report.
 
-### Multiple sources for the same trigger (Multi-touch)
-
-If multiple sources were clicked and associated with a single attribution trigger, send reports for the one with the highest priority.
-If no priority is specified, the browser performs last-touch.
-
-There are many possible alternatives to this,
```
johnivdel, line 254: I think we should still keep this section. It's useful to frame that the chosen model: priority based attribution is one possibility, and other models could potentially be considered.

johnivdel (approved):

PR #193, opened 2021-07-23
Spec trigger priority

apasel422:
***
<a href="https://pr-preview.s3.amazonaws.com/apasel422/conversion-measurement-api/pull/193.html" title="Last updated on Aug 6, 2021, 8:00 PM UTC (29f0eb9)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/conversion-measurement-api/193/1f4084a...apasel422:29f0eb9.html" title="Last updated on Aug 6, 2021, 8:00 PM UTC (29f0eb9)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -252,6 +252,8 @@ This specification defines a [=policy-controlled feature=] identified by the str
 An attribution source is a [=struct=] with the following items:
 
 <dl dfn-for="attribution source">
+: <dfn>source identifier</dfn>
+:: A unique opaque string.
 : <dfn>source origin</dfn>
 :: An [=url/origin=].
 : <dfn>event id</dfn>
@@ -312,8 +314,8 @@ An attribution report is a [=struct=] with the following items:
 :: A 64-bit integer.
 : <dfn>trigger time</dfn>
 :: A point in time.
-: <dfn>source</dfn>
-:: An [=attribution source=].
+: <dfn>source identifier</dfn>
+:: An opaque string.
 
 </dl>
 
@@ -358,6 +360,7 @@ To <dfn>parse an attribution destination</dfn> from a string |str|:
 <h3 algorithm id="obtaining-attribution-source-anchor">Obtaining an attribution source from an <code>a</code> element</h3>
 
 To <dfn>obtain an attribution source</dfn> from an <{a}> element |anchor|:
+1. Let |sourceIdentifier| be a new unique opaque string.
 1. Let |currentTime| be the current time.
 1. If |anchor| does not have both an <{a/attributiondestination}> attribute and
     an <{a/attributionsourceeventid}> attribute, return null.
@@ -385,6 +388,8 @@ To <dfn>obtain an attribution source</dfn> from an <{a}> element |anchor|:
     results in an integer, then set |priority| to that value.
 1. Let |source| be a new [=attribution source=] struct whose items are:
 
+    : [=attribution source/source identifier=]
+    :: |sourceIdentifier|
     : [=attribution source/source origin=]
     :: |sourceOrigin|
     : [=attribution source/event id=]
@@ -516,7 +521,7 @@ To <dfn>trigger attribution</dfn> given an [=attribution trigger=] |trigger| run
     user agent's [=max reports per source=] value, then:
     1. Let |matchingReports| be all entries in the [=attribution report cache=] where all of the following are true:
          * entry's [=attribution report/report time=] and |report|'s [=attribution report/report time=] are equal.
-         * entry's [=attribution report/source=] and |report|'s [=attribution report/source=] are the same.
+         * entry's [=attribution report/source identifier=] and |report|'s [=attribution report/source identifier=] are equal.
     1. If |matchingReports| is empty, then [=list/remove=] |sourceToAttribute| from the [=attribution source cache=] and return.
     1. Set |matchingReports| to the result of [=list/sort in ascending order|sorting=] |matchingReports|
         in ascending order, with |a| being less than |b| if any of the following are true:
@@ -591,8 +596,8 @@ To <dfn>obtain a report</dfn> given an [=attribution source=] |source| and an [=
     :: |trigger|'s [=attribution trigger/trigger priority=].
     : [=attribution report/trigger time=]
     :: |trigger|'s [=attribution trigger/trigger time=].
-    : [=attribution report/source=]
-    :: |source|
+    : [=attribution report/source identifier=]
+    :: |source|'s [=attribution source/source identifier=].
 1. Return |report|.
 
 # Security consideration # {#security-considerations}
--- a/index.bs
+++ b/index.bs
@@ -288,6 +288,8 @@ An attribution trigger is a [=struct=] with the following items:
 :: An [=url/origin=].
 : <dfn>dedup key</dfn>
 :: Null or a 64-bit integer.
+: <dfn>trigger priority</dfn>
+:: A 64-bit integer.
 
 </dl>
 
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -492,22 +504,30 @@ To <dfn>trigger attribution</dfn> given an [=attribution trigger=] |trigger| run
...
 1. For each |item| of |matchingSources|:
     1. [=list/Remove=] |item| from the [=attribution source cache=].
-1. Let |report| be the result of running [=obtain a report=] with |sourceToAttribute| and |trigger|.
 1. If the [=list/size=] of the [=attribution report cache=] is greater than an implementation-defined limit, return.
 1. Add |report| to the [=attribution report cache=].
 1. Increment |sourceToAttribute|'s [=attribution source/number of reports=] value by 1.
 1. If |trigger|'s [=attribution trigger/dedup key=] is not null, [=list/append=] it to |sourceToAttribute|'s
     [=attribution source/dedup keys=].
-1. If |sourceToAttribute|'s [=attribution source/number of reports=] value is greater than or equal to the
-    user agent's [=max reports per source=] value, [=list/remove=] |sourceToAttribute| from the [=attribution source cache=].
```
apasel422, line 510: I'm not sure whether this should be incorporated into the replacement section above.

index.bs
```diff
@@ -492,22 +504,30 @@ To <dfn>trigger attribution</dfn> given an [=attribution trigger=] |trigger| run
...
     in descending order, with |a| being less than |b| if any of the following are true:
       * |a|'s [=attribution source/priority=] is less than |b|'s [=attribution source/priority=].
-      * |a|'s [=attribution source/priority=] is equal to |b|'s [=attribution source/priority=] and |a|'s 
+      * |a|'s [=attribution source/priority=] is equal to |b|'s [=attribution source/priority=] and |a|'s
          [=attribution source/source time=] is less than |b|'s [=attribution source/source time=].
 1. Let |sourceToAttribute| be the first item in |matchingSources|.
 1. If |trigger|'s [=attribution trigger/dedup key=] is not null and |sourceToAttribute|'s
     [=attribution source/dedup keys=] [=list/contains=] it, return.
+1. Let |report| be the result of running [=obtain a report=] with |sourceToAttribute| and |trigger|.
+1. If |sourceToAttribute|'s [=attribution source/number of reports=] value is greater than or equal to the
```
johnivdel, line 513: Number of reports should never be greater than max reports per source, we might want to use just equals so that is clear.

index.bs
```diff
@@ -492,22 +504,30 @@ To <dfn>trigger attribution</dfn> given an [=attribution trigger=] |trigger| run
...
+    user agent's [=max reports per source=] value, then:
+    1. Let |matchingReports| be all entries in the [=attribution report cache=] where all of the following are true:
+         * entry's [=attribution report/report time=] and |report|'s [=attribution report/report time=] are equal.
+         * entry's [=attribution report/source=] and |report|'s [=attribution report/source=] are the same.
+    1. Set |matchingReports| to the result of [=list/sort in ascending order|sorting=] |matchingReports|
+        in ascending order, with |a| being less than |b| if |a|'s [=attribution report/trigger priority=] is less than |b|'s [=attribution report/trigger priority=].
```
johnivdel, line 519: How do we do tie breaks here? We could rely on stable sorting, but I think it is more well-defined to favor deleting the most recent conversion in this case.

index.bs
```diff
@@ -492,22 +504,30 @@ To <dfn>trigger attribution</dfn> given an [=attribution trigger=] |trigger| run
...
+    1. Let |lowestPriorityReport| be the first item in |matchingReports|.
+    1. If |report|'s [=attribution report/trigger priority=] is less than or equal to |lowestPriorityReport|'s [=attribution report/trigger priority=], return.
```
johnivdel, line 521: I think we still want to delete the rest of sources in matching sources in this case. The browser properly ran the attribution logic, it just chose not to send the report.

index.bs
johnivdel: I think it should be:

If matchingReports is empty we cannot do prioritization and should delete the source

index.bs
```diff
@@ -492,22 +504,30 @@ To <dfn>trigger attribution</dfn> given an [=attribution trigger=] |trigger| run
...
     [=attribution source/dedup keys=] [=list/contains=] it, return.
+1. Let |report| be the result of running [=obtain a report=] with |sourceToAttribute| and |trigger|.
+1. If |sourceToAttribute|'s [=attribution source/number of reports=] value is greater than or equal to the
+    user agent's [=max reports per source=] value, then:
+    1. Let |matchingReports| be all entries in the [=attribution report cache=] where all of the following are true:
+         * entry's [=attribution report/report time=] and |report|'s [=attribution report/report time=] are equal.
+         * entry's [=attribution report/source=] and |report|'s [=attribution report/source=] are the same.
+    1. Set |matchingReports| to the result of [=list/sort in ascending order|sorting=] |matchingReports|
+        in ascending order, with |a| being less than |b| if |a|'s [=attribution report/trigger priority=] is less than |b|'s [=attribution report/trigger priority=].
+    1. Let |lowestPriorityReport| be the first item in |matchingReports|.
```
johnivdel, line 520: We need to handle matchingReports being empty

index.bs
apasel422: Aren't the report times inherently equal for `matchingReports`?

apasel422: Oh, you probably mean trigger time, which isn't currently associated with the report structure here. I'll update that.

index.bs
apasel422: Is that different from what's currently implemented?

https://source.chromium.org/chromium/chromium/src/+/master:content/browser/conversions/conversion_storage_sql.cc;l=521-526;drc=09b4249fe83da3613c09b66cdab1d44dd40d21fd

johnivdel: Yes it is. Looking further, I think we are actually guaranteed matchingSource only contains sourceToAttribute at this point:

 - other sources would have gotten deleted on the first trigger
 - new sources afterwards would have deleted sourceToAttribute due to the store an attribution source logic

Does that seem right to you?

In any case think we can leave this as is.  Might be worth adding a note here to explain that.

apasel422: Yes, that seems right. What's the proper way to add a note like that?

johnivdel: I'm not really sure. One concern is that it could get out of date if algorithms change. I think what you have is fine.

apasel422: Ack.

index.bs
```diff
@@ -492,22 +511,34 @@ To <dfn>trigger attribution</dfn> given an [=attribution trigger=] |trigger| run
...
-      * |a|'s [=attribution source/priority=] is equal to |b|'s [=attribution source/priority=] and |a|'s 
+      * |a|'s [=attribution source/priority=] is equal to |b|'s [=attribution source/priority=] and |a|'s
          [=attribution source/source time=] is less than |b|'s [=attribution source/source time=].
 1. Let |sourceToAttribute| be the first item in |matchingSources|.
 1. If |trigger|'s [=attribution trigger/dedup key=] is not null and |sourceToAttribute|'s
     [=attribution source/dedup keys=] [=list/contains=] it, return.
+1. Let |report| be the result of running [=obtain a report=] with |sourceToAttribute| and |trigger|.
+1. If |sourceToAttribute|'s [=attribution source/number of reports=] value is equal to the
+    user agent's [=max reports per source=] value, then:
+    1. Let |matchingReports| be all entries in the [=attribution report cache=] where all of the following are true:
```
johnivdel, line 522: Do you think it's worthwhile to split this into a different algorithm? 

I'm thinking, if reports is equal to max reports, let removed be the the result of running "maybe remove lowest priority report"...

johnivdel (commented):
lgtm with comments

index.bs
apasel422: Seems unnecessary for the time being, though if there's any additional complexity we can do so.

johnivdel (approved):

PR #203, opened 2021-08-31
Specify payload format

### LINKED ISSUE #180: Formally specify aggregate report payload
We should resolve the [aggregate explainer](https://github.com/WICG/conversion-measurement-api/blob/main/AGGREGATE.md#aggregate-attribution-reports)'s TODO to formally specify the aggregate report payload.

alexmturner:
Resolves #180. Also makes minor updates to report format


alexmturner:

AGGREGATE.md
```diff
@@ -187,8 +189,22 @@ The `payload` will need to contain all the information needed for the aggregatio
...
+  "operation": "hierarchical-histogram",  // Allows for the service to support other operations in the future
+  "privacy_budget_key": "<field for server to do privacy budgeting>",
+  "scheduled_report_time": <timestamp in msec>,
+  "reporting_origin": "https://reporter.example",
+  "dpf_key": <binary serialization of the DPF key>,
+}
+```
+Note that the `dpf_key` value would differ in the two payloads.
 
+For the single-server design, the `dpf_key` field would be replaced with `"data": { "bucket": <bucket>, "value": <value> }`. If two `aggregationServices` are specified, one payload (chosen randomly) would contain that data and the other would instead contain `"data": {}`.
```
johnivdel, line 207:
```suggestion
Payloads generated with the `single-server` processingType will directly encode the bucket and value of the histogram contribution instead of a DPF key. The `dpf_key` field will be replaced with `"data": { "bucket": <bucket>, "value": <value> }`. 

If two `aggregationServices` are specified, one payload (chosen randomly) would contain that data and the other would instead contain `"data": {}`.
```

AGGREGATE.md
```diff
@@ -187,8 +189,22 @@ The `payload` will need to contain all the information needed for the aggregatio
 
 The payload should be encrypted via [HPKE](https://datatracker.ietf.org/doc/draft-irtf-cfrg-hpke/), to public keys specified by the processing origins at some well-known address  `/.well-known/aggregation-service/keys.json` that the browser can fetch. Note that we are avoiding using the `attribution-reporting` namespace because many APIs may want to use this infrastructure beyond attribution reporting.
 
-**TODO**: more formally specify `payload`.
+For the MPC protocol, we propose the contents of this base64-encoded encrypted payload will be a [CBOR](https://cbor.io) map with the following structure:
```
johnivdel, line 192:
```suggestion
The contents of this base64-encoded encrypted payload will be a [CBOR](https://cbor.io) map.

Payloads generated with the `two-party` processingType will have the following structure:
```

AGGREGATE.md
```diff
@@ -187,8 +189,22 @@ The `payload` will need to contain all the information needed for the aggregatio
 
 The payload should be encrypted via [HPKE](https://datatracker.ietf.org/doc/draft-irtf-cfrg-hpke/), to public keys specified by the processing origins at some well-known address  `/.well-known/aggregation-service/keys.json` that the browser can fetch. Note that we are avoiding using the `attribution-reporting` namespace because many APIs may want to use this infrastructure beyond attribution reporting.
```
johnivdel, line 190:
```suggestion
The payload should be encrypted via [HPKE](https://datatracker.ietf.org/doc/draft-irtf-cfrg-hpke/), to public keys specified by the processing origins at some well-known address  `/.well-known/aggregation-service/keys.json` that the browser can fetch and then base64 encoded. Note that we are avoiding using the `attribution-reporting` namespace because many APIs may want to use this infrastructure beyond attribution reporting.
```

AGGREGATE.md
```diff
@@ -187,8 +189,22 @@ The `payload` will need to contain all the information needed for the aggregatio
...
 
-**TODO**: more formally specify `payload`.
+For the MPC protocol, we propose the contents of this base64-encoded encrypted payload will be a [CBOR](https://cbor.io) map with the following structure:
+
+```
+// CBOR
+{
+  "version": "<api version>",
+  "operation": "hierarchical-histogram",  // Allows for the service to support other operations in the future
+  "privacy_budget_key": "<field for server to do privacy budgeting>",
```
johnivdel, line 199: Should we keep a TODO to more formally specify privacy_budget_key?

alexmturner: Done

alexmturner:
@johnivdel thanks! ready for another round

AGGREGATE.md
```diff
@@ -185,7 +187,9 @@ The `payload` will need to contain all the information needed for the aggregatio
...
 
-The payload should be encrypted via [HPKE](https://datatracker.ietf.org/doc/draft-irtf-cfrg-hpke/), to public keys specified by the processing origins at some well-known address  `/.well-known/aggregation-service/keys.json` that the browser can fetch. Note that we are avoiding using the `attribution-reporting` namespace because many APIs may want to use this infrastructure beyond attribution reporting.
+**TODO:** more formally specify the `privacy_budget_key`.
+
+The payload should be encrypted via [HPKE](https://datatracker.ietf.org/doc/draft-irtf-cfrg-hpke/) and then base64 encoded. The encryption will use public keys specified by the processing origins at some well-known address  `/.well-known/aggregation-service/keys.json` that the browser can fetch. Note that we are avoiding using the `attribution-reporting` namespace because many APIs may want to use this infrastructure beyond attribution reporting.
```
johnivdel, line 192:
```suggestion
The `payload` should be encrypted via [HPKE](https://datatracker.ietf.org/doc/draft-irtf-cfrg-hpke/) and then base64 encoded. The encryption will use public keys specified by the processing origins at some well-known address  `/.well-known/aggregation-service/keys.json` that the browser can fetch. Note that we are avoiding using the `attribution-reporting` namespace because many APIs may want to use this infrastructure beyond attribution reporting.
```
wdyt? here and below

AGGREGATE.md
```diff
@@ -209,8 +213,26 @@ The browser will encrypt payloads just before the report is sent by requesting p
...
+  "privacy_budget_key": "<field for server to do privacy budgeting>",
+  "scheduled_report_time": <timestamp in msec>,
+  "reporting_origin": "https://reporter.example",
+  "dpf_key": <binary serialization of the DPF key>,
+}
+```
+Note that the `dpf_key` value would differ in the two payloads.
 
+Payloads generated with the `single-server` processing type will directly encode the bucket and value of the histogram contribution instead of a DPF key.
+The `dpf_key` field will be replaced with `"data": { "bucket": <bucket>, "value": <value> }`.
```
johnivdel, line 234: I wonder if it makes sense to encode bucket as a string here, it's possible we may want the schema to be robust to output domain size.

johnivdel:
lgtm % comments + question

johnivdel (approved):

AGGREGATE.md
alexmturner: The payload is currently CBOR, which supports integers in [-2^64, 2^64). Do we expect bucket sizes might go beyond 64-bit unsigned/65-bit signed? I'm leaning towards leaving it as is for efficiency and because supporting >64-bit integers will already require some non-trivial changes to the parsing/processing code (so changing to a string then won't be too bad). Wdyt?

AGGREGATE.md
alexmturner: I'm a little on the fence about having every reference use code style (it gets a little distracting). I went with code styling the references to the encrypted, encoded payload, and leaving the references to the unencrypted, unencoded payload unstyled.

But, happy to change to whatever if you'd prefer

johnivdel: Ack

AGGREGATE.md
johnivdel: Ack. Given the explainer uses 32 bits as the example, this seems fine. I can file a separate issue to track this as it probably affects other parts as well.

PR #205, opened 2021-09-01
Add the new key format for Aggregation Service

linnan-github:


AGGREGATE.md
```diff
@@ -187,6 +187,26 @@ The `payload` will need to contain all the information needed for the aggregatio
 
 The payload should be encrypted via [HPKE](https://datatracker.ietf.org/doc/draft-irtf-cfrg-hpke/), to public keys specified by the processing origins at some well-known address  `/.well-known/aggregation-service/keys.json` that the browser can fetch. Note that we are avoiding using the `attribution-reporting` namespace because many APIs may want to use this infrastructure beyond attribution reporting.
 
+The  processing origin will respond with a set of keys and a cache control header which dictates how long to store the keys for. The browser will use these keys for the max-age as specified by the header. The scheme of the JSON encoded public keys is as follows:
```
alexmturner, line 190: optional: s/and a cache control header which dictates/which will be stored according to standard HTTP caching rules, i.e. using Cache-Control headers to dictate/
optional: s/max-age/freshness lifetime/

Do we want to mention that the browser might want to enforce minimum/maximum freshness lifetimes?

AGGREGATE.md
```diff
@@ -187,6 +187,26 @@ The `payload` will need to contain all the information needed for the aggregatio
...
+
+
+```
+{
+  "version": "<format version>",
+  "keys": [
+    {
+      "id": "<arbitrary string identifying the key>",
```
alexmturner, line 198: can we specify a maximum string length here, e.g. up to 128 chars?

linnan-github: Fixed. Thanks!

AGGREGATE.md
linnan-github: Updated based on suggestions. Thanks both.

AGGREGATE.md
```diff
@@ -187,6 +187,26 @@ The `payload` will need to contain all the information needed for the aggregatio
 
 The payload should be encrypted via [HPKE](https://datatracker.ietf.org/doc/draft-irtf-cfrg-hpke/), to public keys specified by the processing origins at some well-known address  `/.well-known/aggregation-service/keys.json` that the browser can fetch. Note that we are avoiding using the `attribution-reporting` namespace because many APIs may want to use this infrastructure beyond attribution reporting.
 
+The processing origin will respond with a set of keys which will be stored according to standard HTTP caching rules, i.e. using Cache-Control headers to dictate how long to store the keys for (e.g. following the [freshness lifetime](https://datatracker.ietf.org/doc/html/rfc7234#section-4.2)). The browser could enforce maximum/minimum lifetimes of stored keys to encourage faster key rotation and/or mitigate bandwidth usage. The scheme of the JSON encoded public keys is as follows:
```
johnivdel, line 190: Could we add a bit more color about how these keys gets used to the start of this paragraph.

The browser will encrypt payloads just before the report is sent by requesting public keys for the processing origin. `keys.json` will be fetched with an un-credentialed request.

Let's also add a note below the schema like:

Note: The browser may need some mechanism to ensure that the same set of keys are delivered to different users.

johnivdel:
lgtm % comment

AGGREGATE.md
linnan-github: Thanks John. Updated per your suggestion.

johnivdel (approved):

alexmturner:
lgtm too

PR #209, opened 2021-09-08
Disallow repeated key IDs

### LINKED ISSUE #208: Disallow repeated key IDs
Currently, there is no explicit guidance on whether an origin can reuse a key ID. We should disallow this, at least within a single keyset

alexmturner:
Resolves #208

### DIFF
--- a/AGGREGATE.md
+++ b/AGGREGATE.md
@@ -206,7 +206,7 @@ The browser will encrypt payloads just before the report is sent by requesting p
 }
 ```
 A processing origin should not reuse an ID string for a different key.
-In particular, the origin must not repeat the same ID in a keyset (i.e. a single fetch of `keys.json`).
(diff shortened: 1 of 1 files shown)


alexmturner:
@johnivdel could you PTAL?

AGGREGATE.md
```diff
@@ -205,7 +205,8 @@ The browser will encrypt payloads just before the report is sent by requesting p
   ]
 }
 ```
-
+A processing origin should not reuse an ID string for a different key.
+In particular, the origin must not repeat the same ID in a keyset (i.e. a single fetch of `keys.json`).
```
johnivdel, line 209:
```suggestion
In particular, IDs must be unique within a `keys.json` to be valid.
```

johnivdel (approved):
lgtm with comment

PR #213, opened 2021-09-09
Spec registerAttributionSource API

apasel422:


index.bs
```diff
@@ -440,6 +472,58 @@ To <dfn>maybe obtain an event attribution source</dfn> given an |anchor|, run th
 1. Round |source|'s [=attribution source/expiry=] away from zero to the nearest day.
 1. Return |source|.
 
+<h3 algorithm id="obtaining-attribution-source-params">Obtaining an attribution source from {{AttributionSourceParams}}</h3>
```
johnivdel, line 475: Since having a shared algorithm seems desirable, we could probably just pass each data field as a parameter which seems fairly common.

The other option is to use AttributionSourceParams as the intermediate struct, which makes sense conceptually as the parameters should match across JS/HTML

index.bs
```diff
@@ -376,42 +408,75 @@ To <dfn>parse an attribution destination</dfn> from a string |str|:
 
 <h3 algorithm id="obtaining-attribution-source-anchor">Obtaining an attribution source from an <code>a</code> element</h3>
 
-To <dfn>obtain an attribution source</dfn> from an <{a}> element |anchor|:
+To <dfn>obtain an attribution source from an anchor</dfn> given an <{a}> element |anchor|:
+
+1. Let |params| be an {{AttributionSourceParams}} with the following key/value pairs:
```
johnivdel, line 413: This could follow example 33 https://heycam.github.io/webidl/#idl-dictionaries if we wanted and just send an ordered map in rather than directly doing this construction. I'm having a hard time finding examples of constructing IDL dictionaries as input to a shared function

@jyasskin is there an accepted practice here?

johnivdel (commented):
Overall looks good to me, adding @jyasskin as well

index.bs
```diff
@@ -376,42 +408,75 @@ To <dfn>parse an attribution destination</dfn> from a string |str|:
...
-1. If |anchor|'s [=relevant settings object=]'s [=environment settings object/responsible document=] is
+1. If |params| does not have both an {{AttributionSourceParams/attributionDestination}} field and
+    an {{AttributionSourceParams/attributionSourceEventId}} field, return null.
+1. If |params|'s [=relevant settings object=]'s [=environment settings object/responsible document=] is
     not [=allowed to use=] the [=attribution-reporting=] [=policy-controlled feature=], return null.
 1. Let |attributionDestination| be the result of running
-    [=parse an attribution destination=] with |anchor|'s
-    <{a/attributiondestination}> attribute.
+    [=parse an attribution destination=] with |params|'s
+    {{AttributionSourceParams/attributionDestination}} field.
```
jyasskin, line 455: You can spell this

```suggestion
    [=parse an attribution destination=] with |params|["{{AttributionSourceParams/attributionDestination}}"].
```

Per https://infra.spec.whatwg.org/#map-get and https://heycam.github.io/webidl/#idl-dictionaries

index.bs
```diff
@@ -376,42 +408,75 @@ To <dfn>parse an attribution destination</dfn> from a string |str|:
...
 1. If |attributionDestination| is null, return null.
-1. Let |sourceOrigin| be |anchor|'s [=relevant settings object=]'s
+1. Let |sourceOrigin| be |params|'s [=relevant settings object=]'s
     [=environment/top-level origin=].
 1. Let |reportingOrigin| be |sourceOrigin|.
-1. If |anchor| has an <{a/attributionreportto}> attribute, then:
+1. If |params| has an {{AttributionSourceParams/attributionReportTo}} field, then:
```
jyasskin, line 460: This can be

```suggestion
1. If |params|["{{AttributionSourceParams/attributionReportTo}}"] [=map/exists=], then:
```

Like the example at https://infra.spec.whatwg.org/#example-map-get

index.bs
```diff
@@ -376,42 +408,75 @@ To <dfn>parse an attribution destination</dfn> from a string |str|:
...
+
+To <dfn>obtain an attribution source from params</dfn> given an {{AttributionSourceParams}} |params|, run the following steps:
+
 1. Let |sourceIdentifier| be a new unique opaque string.
 1. Let |currentTime| be the current time.
-1. If |anchor| does not have both an <{a/attributiondestination}> attribute and
-    an <{a/attributionsourceeventid}> attribute, return null.
-1. If |anchor|'s [=relevant settings object=]'s [=environment settings object/responsible document=] is
+1. If |params| does not have both an {{AttributionSourceParams/attributionDestination}} field and
+    an {{AttributionSourceParams/attributionSourceEventId}} field, return null.
```
jyasskin, line 450: These entries are `required`, so you'll have to deal with the possibility that they're missing from the anchor earlier in the flow.

index.bs
```diff
@@ -376,42 +408,75 @@ To <dfn>parse an attribution destination</dfn> from a string |str|:
...
+1. Let |source| be the result of running [=obtain an attribution source from an anchor=] with |anchor|.
+1. If |source| is null, return null.
+1. Return the result of running [=obtain an event attribution source from a source=] with |source|.
+
+<h3 algorithm id="obtaining-event-attribution-source-source">Obtaining an event attribution source from a source</h3>
+
+To <dfn>obtain an event attribution source from a source</dfn> given an [=attribution source=] |source|, run the following steps:
+
+1. Set |source|'s [=attribution source/source type=] to "<code>event</code>".
+1. Round |source|'s [=attribution source/expiry=] away from zero to the nearest day.
```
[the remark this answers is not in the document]
apasel422: Filed #236 for this.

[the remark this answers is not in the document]
apasel422: Actually, `expiry` is a duration, not a date, so I don't think there's any ambiguity in terms of time zone or leap seconds. Neither should apply, correct?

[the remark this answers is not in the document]
jyasskin: https://pr-preview.s3.amazonaws.com/apasel422/conversion-measurement-api/pull/213.html#attribution-source-expiry says it's a point in time, rather than a duration. I think the `<a>` attribute is a duration, which would remove the timezone ambiguity if you rounded that value before converting it to an attribution source. There's still ambiguity around leap seconds, so you could say "to the nearest day (86400 seconds)".

apasel422:
@jyasskin Please take another look, thanks!

index.bs
```diff
@@ -254,6 +254,39 @@ add the following check to the list:
...
+  readonly attribute AttributionReporting attributionReporting;
+};
+</pre>
+
+The
+<dfn method for=AttributionReporting><code>registerAttributionSource(<var>params</var>)</code></dfn>
+method steps are:
+
+1. Let |source| be the result of running [=obtain an attribution source from params=] with |params|
+    and the [=relevant settings object=].
```
jyasskin, line 284:
```suggestion
    and the [=relevant settings object=] of [=this=].
```

jyasskin (approved):
Some tweaks, but otherwise looks good. Thanks!

apasel422:
@johnivdel PTAL

PR #232, opened 2021-09-23
Spec trigger data noising

apasel422:


index.bs
```diff
@@ -370,6 +370,18 @@ To <dfn>parse attribution data</dfn> given a [=string=] |input| modulo an intege
...
 
+<h3 algorithm id="noising-data-fields">Noising data fields</h3>
+
+This section defines how to noise [=attribution trigger/trigger data=] and
+[=attribution trigger/event source trigger data=].
+
+To <dfn>noise attribution data</dfn> given an integer |input| between 0 (inclusive) and |maxData|
+(exclusive) and a probability |noiseRate|:
+
+1. Let |shouldNoise| be a randomly selected boolean value such that true is selected with probability |noiseRate|.
```
johnivdel, line 381: Is there any precedence for going the probability approach rather than e.g. a double and comparing against a random double between 0 and 1?

The double approach seems like it leaves less to the imagination

index.bs
```diff
@@ -536,6 +536,19 @@ To <dfn>process an attribution source</dfn> given an [=attribution source=] |sou
...
 # Triggering Algorithms # {#trigger-algorithms}
 
+<h3 algorithm id="noising-data-fields">Noising trigger data</h3>
+
+This section defines how to noise [=attribution trigger/trigger data=] and
+[=attribution trigger/event source trigger data=].
+
+To <dfn>noise trigger data</dfn> given an integer |input|, integer |maxData|, and double |noiseRate|,
+where |input| must be between 0 (inclusive) and |maxData| (exclusive) and |noiseRate| must be
+between 0 and 1 (both inclusive), run the following steps:
```
johnivdel, line 546: wdyt about moving input arg requirements into asserts, similar to: https://fetch.spec.whatwg.org/#concept-request-add-range-header

PR #251, opened 2021-10-13
Spec false report attribution mode

apasel422:
When a source's attribution mode is "falsely", the current implementation immediately deactivates it, preventing the source from subsequently being selected during attribution, but since there is no concept of active vs. inactive in the spec, the spec instead skips adding such sources to the source cache entirely.

***
<a href="https://pr-preview.s3.amazonaws.com/apasel422/conversion-measurement-api/pull/251.html" title="Last updated on Nov 8, 2021, 7:38 PM UTC (05f60fe)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/conversion-measurement-api/251/004ca47...apasel422:05f60fe.html" title="Last updated on Nov 8, 2021, 7:38 PM UTC (05f60fe)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -423,21 +423,6 @@ To <dfn>parse an attribution destination</dfn> from a string |str|:
 1. Return the result of [=obtain a site|obtaining a site=] from |url|'s
     [=url/origin=].
 
-<h3 algorithm id="noising-attribution">Noising attribution</h3>
-
-To <dfn>select a random attribution mode</dfn> given doubles |neverRate| and |falselyRate|, run the
-following steps:
-
-1. Assert: |neverRate| is between 0 and 1 (both inclusive).
-1. Assert: |falselyRate| is between 0 and 1 (both inclusive).
-1. Let |noiseRate| be the sum of |neverRate| and |falselyRate|.
-1. Assert: |noiseRate| <= 1.
-1. Let |r| be a random double between 0 (inclusive) and 1 (exclusive).
-1. If |r| is greater than or equal to |noiseRate|, return "<code>truthfully</code>".
-1. Let |t| be a random double between 0 (inclusive) and 1 (exclusive).
-1. If |t| is greater than or equal to |falselyRate|, return "<code>never</code>".
-1. Return "<code>falsely</code>".
-
 <h3 algorithm id="obtaining-attribution-source-anchor">Obtaining an attribution source from an <code>a</code> element</h3>
 
 To <dfn>obtain an attribution source from an anchor</dfn> given an <{a}> element |anchor|:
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -426,12 +424,16 @@ To <dfn>parse an attribution destination</dfn> from a string |str|:
...
-To <dfn>select a random attribution mode</dfn> given a double |noiseRate|, run the following steps:
+To <dfn>select a random attribution mode</dfn> given doubles |noiseRate| and |falselyRate|, run the
+following steps:
 
 1. Assert: |noiseRate| is between 0 and 1 (both inclusive).
+1. Assert: |falselyRate| is between 0 and 1 (both inclusive).
 1. Let |r| be a random double between 0 (inclusive) and 1 (exclusive).
 1. If |r| is greater than or equal to |noiseRate|, return "<code>truthfully</code>".
-1. Return "<code>never</code>".
+1. Let |t| be a random double between 0 (inclusive) and 1 (exclusive).
```
johnivdel, line 434: I don't think this will work as expected (let me know if my understanding is wrong).

The choice of falsely is conditioned on 1 - noiseRate, and so it will happen with (1-noise rate) * falsely rate probability.

index.bs
```diff
@@ -556,8 +563,34 @@ To <dfn>process an attribution source</dfn> given an [=attribution source=] |sou
 1. [=list/Remove=] all entries in |cache| where the entry's [=attribution source/expiry time=] is less than the current time.
 1. If the [=list/size=] of |cache| is greater than or equal to the user agent's
      [=max source cache size=], return.
+1. If |source|'s [=attribution source/attribution mode=] is "<code>falsely</code>", then:
+    1. Assert: |source|'s [=attribution source/source type=] is "<code>event</code>".
+    1. Let |fakeReport| be a new [=attribution report=] with the items:
```
johnivdel, line 568: Should this pass in a generated trigger to obtain a report instead?

index.bs
```diff
@@ -556,8 +563,34 @@ To <dfn>process an attribution source</dfn> given an [=attribution source=] |sou
...
+        :: 0
+        : [=attribution report/trigger time=]
+        :: |source|'s [=attribution source/source time=]
+        : [=attribution report/source identifier=]
+        :: |source|'s [=attribution source/source identifier=]
+    1. Add |fakeReport| to the [=attribution report cache=].
+    1. Return.
 1. [=set/Append=] |source| to |cache|.
 
+Issue: Should fake reports respect the user agent's [=max report cache size=] and/or
```
johnivdel, line 591: Good call out! I don't see why this should not respect the max report cache size.

As for max reports per attribution destination, that seems a bit more nuanced and leaving as an issue sgtm.

johnivdel (commented):
generally lgtm % comments

index.bs
apasel422: One issue is that a trigger requires a trigger origin, but we only have an attribution destination (site) from the source. I guess we could change the trigger structure to store a site instead?

index.bs
apasel422: Is that different from [what the implementation does now](https://source.chromium.org/chromium/chromium/src/+/master:content/browser/attribution_reporting/attribution_policy.cc;l=121-141;drc=a67b4875d7d7f06ebbd559b591198c3eec181968)?

johnivdel: Ah, I misunderstood the meaning of falselyRate, this does work fine.

I think it might be a bit more intuitive to change the approach here to specify two rate:

% of sources which get never
% of sources which get falsely

If a UA wanted to vary one of the rate, it would require them to do a bit more math to get the constants. wdyt?

index.bs
johnivdel: Looks like trigger origin is used only used an origin for data clearing... which is erroneous anyway because triggers are not stored in any of the caches. Assuming that behavior was fixed to pass the origin to the report, we would actually still require a trigger origin here.

Perhaps we could add an attribution destination to triggers (in addition to the origin), and set the trigger origin as an opaque origin which should not break any assertions in this spec. wdyt?

index.bs
```diff
@@ -427,12 +425,18 @@ To <dfn>parse an attribution destination</dfn> from a string |str|:
...
-1. Assert: |noiseRate| is between 0 and 1 (both inclusive).
+1. Assert: |neverRate| is between 0 and 1 (both inclusive).
+1. Assert: |falselyRate| is between 0 and 1 (both inclusive).
+1. Let |noiseRate| be the sum of |neverRate| and |falselyRate|.
+1. Assert: |noiseRate| <= 1.
 1. Let |r| be a random double between 0 (inclusive) and 1 (exclusive).
 1. If |r| is greater than or equal to |noiseRate|, return "<code>truthfully</code>".
-1. Return "<code>never</code>".
+1. Let |t| be a random double between 0 (inclusive) and 1 (exclusive).
+1. If |t| is greater than or equal to |falselyRate|, return "<code>never</code>".
```
johnivdel, line 438: Should t be a random number between 0 and noiseRate here?

johnivdel (approved):

index.bs
apasel422: Would it be less confusing to just specify this algorithm as:

> Return one of "`truthfully`", "`falsely`", or "`never`", with implementation-defined probability.

?

johnivdel: Yes I am supportive of that simplification (seeing this version of the algorithm).

PR #252, opened 2021-10-13
Spec report ID for deduplication in the event of retries on network errors

### LINKED ISSUE #228: Consider adding a nonce to attribution reports
As described in https://privacycg.github.io/private-click-measurement/#receving-reports,

> We may have to add a nonce to the HTTP POST request to prohibit double counting in cases where the user agent decides to retry the request.

There are potentially many edge cases that can cause the reporting endpoint and the browser to have a different view of what happened to a request. 

For example, the report request made it to the reporting endpoint server but there was a transient network issue before response headers were received by the client. Without adding a nonce, any retry logic has the potential to introduce counting issues.

This would be beneficial for both aggregate and event-level reports.

apasel422:
Fixes #228

***
<a href="https://pr-preview.s3.amazonaws.com/apasel422/conversion-measurement-api/pull/252.html" title="Last updated on Nov 11, 2021, 6:54 PM UTC (d00ade6)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/conversion-measurement-api/252/3c7fb7b...apasel422:d00ade6.html" title="Last updated on Nov 11, 2021, 6:54 PM UTC (d00ade6)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -18,7 +18,7 @@ Assume Explicit For: on
 <pre class=link-defaults>
 spec:html; type:element; text:a
 spec: uuid; type: dfn; urlPrefix: https://wicg.github.io/uuid/
-    text: generating a random UUID; url: #dfn-generate-a-random-uuid
+    text: generate a random UUID; url: #dfn-generate-a-random-uuid
 </pre>
 
 Introduction {#intro}
--- a/index.bs
+++ b/index.bs
@@ -847,7 +847,7 @@ To <dfn>obtain a report</dfn> given an [=attribution source=] |source| and an [=
     : [=attribution report/source identifier=]
     :: |source|'s [=attribution source/source identifier=].
     : [=attribution report/report id=]
-    :: A random string.
+    :: The result of [=generating a random UUID=].
 1. Return |report|.
 
 # Report delivery # {#report-delivery}
--- a/index.bs
+++ b/index.bs
@@ -17,6 +17,8 @@ Assume Explicit For: on
 </pre>
 <pre class=link-defaults>
 spec:html; type:element; text:a
+</pre>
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -844,6 +846,8 @@ To <dfn>obtain a report</dfn> given an [=attribution source=] |source| and an [=
     :: |trigger|'s [=attribution trigger/trigger time=].
     : [=attribution report/source identifier=]
     :: |source|'s [=attribution source/source identifier=].
+    : [=attribution report/report id=]
+    :: A random string.
```
johnivdel, line 850: Do we want to enforce this is a random number encoded as a string or something?  E.g. like a random 64 bit int.

johnivdel (approved):

index.bs
apasel422: I wasn't sure if that was overly restrictive, but that would be fine with me. CC @csharrison.

johnivdel: I was thinking to align with eventid, but I agree it's not necessary. E.g. a UUID seems like a good fit here.

We may want to make some statement on uniqueness, e.g.:

A new unique random string.

apasel422: Are we sure we want it to be unique? The client would have to scan the report cache to ensure that. And then we have the question of whether we need to retain a cache of already-used IDs for all time, or if it only needs to be unique among reports that haven't been sent yet.

johnivdel: Good point. I guess functionally what we want is:

"A random string with high enough entropy that they should be effectively unique"

Maybe that is enough of an argument to just use a UUID?

apasel422: Changed to UUID.

johnivdel (approved):

PR #257, opened 2021-10-19
Include attribution destination in report instead of referrer header

apasel422:
This brings the spec in line with PCM: https://privacycg.github.io/private-click-measurement/#report-format


apasel422:
> Can we include this in the explainer as well?

Done.

index.bs
```diff
@@ -800,6 +800,8 @@ To <dfn>serialize an [=attribution report=]</dfn> |report|, run the following st
 
 1. Let |data| be a [=map=] of the following key/value pairs:
 
+    : "`attribution_destination`"
+    :: |report|'s [=attribution report/attribution destination=]
```
johnivdel, line 804: We need to pass this in as a string given the requirements of https://infra.spec.whatwg.org/#convert-an-infra-value-to-a-json-compatible-javascript-value

event_attribution_reporting_clicks.md
```diff
@@ -336,12 +336,13 @@ https://attributionreportto/.well-known/attribution-reporting/report-attribution
 
 The report data is included in the request body as a JSON object with the following keys:
 
+-   `attribution_destination`: the eTLD+1 that triggered attribution
```
johnivdel, line 339: Could also just say the attribution_destination set on the attribution source to match the description of event id

index.bs
```diff
@@ -798,8 +798,12 @@ To <dfn>queue a report for delivery</dfn> given an [=attribution report=] |repor
 
 To <dfn>serialize an [=attribution report=]</dfn> |report|, run the following steps:
 
+1. Let |destination| be |report|'s [=attribution report/attribution destination=].
+1. Assert: |destination| is not an [=opaque origin=].
 1. Let |data| be a [=map=] of the following key/value pairs:
 
+    : "`attribution_destination`"
+    :: The [=string/concatenate|concatenation=] of «|destination|'s scheme, "`://`", |destination|'s host».
```
johnivdel, line 806: I think this should |destination|'s host, serialized to match the URL serialization algo: https://url.spec.whatwg.org/#url-serializing

However it seems like that definition is not exported, so it may be necessary to create an origin and serialize that directly.

johnivdel (commented):
generally lgtm % question on serializing the site

PR #266, opened 2021-11-03
Store event id and trigger data as integers, not strings

apasel422:
***
<a href="https://pr-preview.s3.amazonaws.com/apasel422/conversion-measurement-api/pull/266.html" title="Last updated on Nov 5, 2021, 7:00 PM UTC (21a61bd)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/conversion-measurement-api/266/cabacc3...apasel422:21a61bd.html" title="Last updated on Nov 5, 2021, 7:00 PM UTC (21a61bd)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -794,6 +794,12 @@ To <dfn>queue a report for delivery</dfn> given an [=attribution report=] |repor
     Note: This is intended to allow user agents to optimize device resource usage.
 1. Run [=attempt to deliver a report=] with |report|.
 
+<h3 id="serialize-integer">Serialize an integer</h3>
+
+To <dfn>serialize an integer</dfn>, represent it as a string of the shortest possible decimal number.
+
+Issue: This will be replaced by a more descriptive algorithm in Infra. See infra/201.
+
 <h3 id="serialize-report-body">Serialize attribution report body</h3>
 
 To <dfn>serialize an [=attribution report=]</dfn> |report|, run the following steps:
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -813,6 +813,10 @@ To <dfn>serialize an [=attribution report=]</dfn> |report|, run the following st
 
 1. Return the [=byte sequence=] resulting from executing [=serialize an infra value to JSON bytes=] on |data|.
 
+Issue: How can we force [=attribution report/event id=] and [=attribution report/trigger data=] to
```
johnivdel, line 816: Could we directly convert these to strings before placing them in the map?

apasel422: I couldn't find a way to convert a 64-bit integer (or any integer, necessarily) into a string, without relying on a JS function.

johnivdel: I think we can copy what fetch does for now: https://fetch.spec.whatwg.org/#serialize-an-integer

index.bs
```diff
@@ -794,6 +794,13 @@ To <dfn>queue a report for delivery</dfn> given an [=attribution report=] |repor
     Note: This is intended to allow user agents to optimize device resource usage.
 1. Run [=attempt to deliver a report=] with |report|.
 
+<h3 id="serialize-integer">Serialize an integer</h3>
+
+To <dfn>serialize an integer</dfn>, represent it as a string of the shortest possible decimal number.
+
+Issue: This will be replaced by a more descriptive algorithm in Infra. See
```
johnivdel, line 801: Given that issue has been open a long time, maybe something like "This would ideally use a..."

johnivdel (approved):

Issue #234, opened 2021-09-24
Consider removing keys.json format version string


alexmturner:
Instead, we could update the URL whenever backwards incompatible changes are made. This would allow aggregation origins to support multiple incompatible versions simultaneously without requiring all the relevant information to be packed into one JSON response during the transition.

One downside is that the client may need to make multiple requests to find the highest mutually supported version. This may not be a major drawback if we expect helpers will be kept up-to-date with stable browser releases’ newest supported versions. This could also be mitigated by separately serving the supported version set, e.g. as part of a browser's allowlist.

Note that, even if the version string is removed, the option to add a version string later is still available (with a missing version string indicating the first version).

PR #267, opened 2021-11-11
Remove format version field from keys.json

alexmturner:
Resolves issue #234.

### DIFF
--- a/AGGREGATE.md
+++ b/AGGREGATE.md
@@ -210,7 +210,7 @@ The browser will encrypt payloads just before the report is sent by requesting p
 ```
 In the case of backwards incompatible changes to this scheme (e.g. in future versions), the hosting address (i.e. URL) should also change.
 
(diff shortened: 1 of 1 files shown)


alexmturner:
@johnivdel could you PTAL?

AGGREGATE.md
```diff
@@ -209,7 +208,7 @@ The browser will encrypt payloads just before the report is sent by requesting p
   ]
 }
 ```
-
+In the case of backwards incompatible changes to this scheme (e.g. in future versions), the hosting address (i.e. URL) should also change.
```
johnivdel, line 211:
```suggestion
In the case of backwards incompatible changes to this scheme (e.g. in future versions of the API), the .well-known endpoint should also change.
```

johnivdel (approved):
lgtm % comment

Issue #92, opened 2021-01-06
How do you deal with deduplication?


iab-france:
Hi,

I am writing on behalf of the dedicated IAB France taskforce working on sandbox. On the conversion measurement API, we are having an issue when it is set up more than one report recipients. 

Main use of attribution is to be able to manage deduplication, so if you use a single touch attribution models as last clic and you have more than  media touch point to the conversion, just one media platform receive the credit for the conversion, how do you manage that here?

johnivdel:
Could you clarify what you mean by media platform? Referring to this [glossary](https://github.com/WICG/conversion-measurement-api#glossary), would this be different publishers? Or would these platforms be different reporting origins?

If these are different publishers:
Multiple events leading to a single conversion are handled by the `attribution-credit` reporting parameter. If two ads were clicked, and a conversion was registered, the last clicked ads impression data would have a report with attribution credit 100. The other report would have a credit of 0.

Credit assignment is specified [here](https://github.com/WICG/conversion-measurement-api#register-a-conversion-algorithm) and credits are defined [here](https://github.com/WICG/conversion-measurement-api#multiple-impressions-for-the-same-conversion-multi-touch).

If these are different reporting origins:
The credit assignment is done in the domain of <reporting origin, conversiondestination>. So if two ads with different reporting origins and the same destination are clicked, and a conversion is registered for each of the reporting origins, both reports will have a credit of 100.

iab-france:
It would be advertiser i guess.

Can you illustrate please you last paragraph, let's say you have advertiser A that is running campaign on Google and Facebook. Let's say Google was the first media touch point, Facebook the second one, then the conversion. The attribution mode used is last clic and I want just Facebook to receive 100% of the credit for the conversion, is that possible?

apasel422:
Deduplication keys are supported for both event-level and aggregatable attribution. Feel free to reopen if there are further questions here.

PR #268, opened 2021-11-15
Rename report_id field to idempotency_key

apasel422:
To align with https://tools.ietf.org/id/draft-idempotency-header-01.html.

#92

***
<a href="https://pr-preview.s3.amazonaws.com/apasel422/conversion-measurement-api/pull/268.html" title="Last updated on Nov 15, 2021, 9:05 PM UTC (02ace12)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/conversion-measurement-api/268/da1fafe...apasel422:02ace12.html" title="Last updated on Nov 15, 2021, 9:05 PM UTC (02ace12)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -381,7 +381,7 @@ An attribution report is a [=struct=] with the following items:
 :: A string.
 : <dfn>delivered</dfn> (default false)
 :: A boolean.
-: <dfn>report id</dfn>
+: <dfn>idempotency key</dfn>
 :: A string.
 
 </dl>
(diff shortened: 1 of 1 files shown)


johnivdel:
`idempotency_key` seems to better reflect the usage of this field so I am supportive of this. @csharrison for a second opinion.

csharrison:
I slightly prefer using a more common word than idempotency per https://w3ctag.github.io/design-principles/#naming-common-words, but I agree it explains its own purpose much better than `report_id` does, and I do not have an alternative at the moment other than something that clashes with dedupe key. Although, to be fair, `idempotency_key` also kind of clashes with `dedup_key` somewhat.

Possibly these could be reconciled with name changes like `local_dedup_key` and `remote_dedup_key`.

PR #270, opened 2021-11-19
Rename operation to "histogram"

alexmturner:
This is more succinct and allows for more flexibility in the data processing design


alexmturner:
@csharrison could you PTAL?

csharrison (approved):

PR #274, opened 2021-11-30
Add noise probabilities to report body

### LINKED ISSUE #273: Event-level reports should indicate how noise has been applied
We should consider adding fields like:

`metadata_noise_probability` or `randomized_response_probability`

This will make it easier to handle multiple implementations with different probabilities (or different experiments in the same implementation). It also signals to developers that values may be noisy.

cc @maudnals @johnivdel

apasel422:
Fixes #273.

- `randomized_trigger_data_rate`
- `randomized_trigger_rate`

***
<a href="https://pr-preview.s3.amazonaws.com/apasel422/conversion-measurement-api/pull/274.html" title="Last updated on Jan 6, 2022, 8:39 PM UTC (8686a04)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/conversion-measurement-api/274/1ed40d1...apasel422:8686a04.html" title="Last updated on Jan 6, 2022, 8:39 PM UTC (8686a04)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -337,6 +337,8 @@ An attribution source is a [=struct=] with the following items:
 :: [=ordered set=] of [=attribution trigger/dedup keys=] associated with this [=attribution source=].
 : <dfn>attribution mode</dfn>
 :: An [=attribution mode=].
+: <dfn>randomized trigger rate</dfn>
+:: A number between 0 and 1 (both inclusive).
 
 </dl>
 
@@ -351,6 +353,10 @@ An attribution trigger is a [=struct=] with the following items:
 :: A non-negative 64-bit integer.
 : <dfn>event source trigger data</dfn>
 :: A non-negative 64-bit integer.
+: <dfn>randomized trigger data rate</dfn>
+:: A number between 0 and 1 (both inclusive).
+: <dfn>randomized event-source trigger data rate</dfn>
+:: A number between 0 and 1 (both inclusive).
 : <dfn>trigger time</dfn>
 :: A point in time.
 : <dfn>reporting endpoint</dfn>
@@ -373,6 +379,10 @@ An attribution report is a [=struct=] with the following items:
 :: Either "<code>navigation</code>" or "<code>event</code>".
 : <dfn>trigger data</dfn>
 :: A non-negative 64-bit integer.
+: <dfn>randomized trigger data rate</dfn>
+:: A number between 0 and 1 (both inclusive).
+: <dfn>randomized trigger rate</dfn>
+:: A number between 0 and 1 (both inclusive).
 : <dfn>reporting endpoint</dfn>
 :: An [=url/origin=].
 : <dfn>attribution destination</dfn>
@@ -576,6 +586,8 @@ and an [=environment settings object=] |settings|, run the following steps:
     :: "<code>navigation</code>"
     : [=attribution source/attribution mode=]
     :: ("`truthfully`", null)
+    : [=attribution source/randomized trigger rate=]
+    :: 0
 1. Return |source|.
 
 <dfn>Max event id value</dfn> is a vendor-specific integer which controls
@@ -615,6 +627,10 @@ To <dfn>process an attribution source</dfn> given an [=attribution source=] |sou
         :: 0
         : [=attribution trigger/event source trigger data=]
         :: |source|'s [=attribution source/attribution mode=][1]
+        : [=attribution trigger/randomized trigger data rate=]
+        :: The user agent's [=randomized trigger data rate=].
+        : [=attribution trigger/randomized event-source trigger data rate=]
+        :: The user agent's [=randomized event-source trigger data rate=].
         : [=attribution trigger/trigger time=]
         :: |source|'s [=attribution source/source time=]
         : [=attribution trigger/reporting endpoint=]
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -885,16 +894,28 @@ To <dfn>serialize an [=attribution report=]</dfn> |report|, run the following st
...
     : "`source_type`"
     :: |report|'s [=attribution report/source type=]
     : "`source_event_id`"
     :: |report|'s [=attribution report/event id=], [=serialize an integer|serialized=]
+    : "`source_event_id_noise_probability`"
+    :: 0
     : "`trigger_data`"
     :: |report|'s [=attribution report/trigger data=], [=serialize an integer|serialized=]
+    : "`trigger_data_noise_probability`"
+    :: |triggerDataNoiseProbability|
```
apasel422, line 918: One caveat with this approach is that the probabilities may be wrong for reports in storage that haven't been sent yet, e.g. if the browser is updated between the time a source is stored and the time the report is sent.

We could add fields representing the probabilities to the `attribution source` and `attribution trigger` structures, but it would incur storage overhead. Thoughts?

index.bs
```diff
@@ -483,11 +493,22 @@ To <dfn>obtain an event attribution source from a source</dfn> given an [=attrib
 1. Set |resultSource|'s [=attribution source/source identifier=] to a new unique opaque string.
 1. Set |resultSource|'s [=attribution source/source type=] to "<code>event</code>".
 1. Round |resultSource|'s [=attribution source/expiry=] away from zero to the nearest day (86400 seconds).
-1. Set |resultSource|'s [=attribution source/attribution mode=] to one of
-    "<code>truthfully</code>", "<code>never</code>", or "<code>falsely</code>", with
-    [=implementation-defined=] probability.
+1. Let |never| be a random boolean, with true having an [=implementation-defined=] probability.
```
csharrison, line 496: I don't think this is right. We want `event source falsely mode rate` to include the probability that `never` is selected. There should only be a single probability that the impression will be reported non-truthfully.

index.bs
```diff
@@ -831,13 +862,20 @@ return a point in time.
 To <dfn>obtain a report</dfn> given an [=attribution source=] |source| and an [=attribution trigger=] |trigger|:
 
 1. Let |triggerData| be |trigger|'s [=attribution trigger/trigger data=].
-1. If |source|'s [=attribution source/source type=] is "<code>event</code>", set |triggerData| to |trigger|'s [=attribution trigger/event source trigger data=].
+1. Let |triggerDataNoiseProbability| be |trigger|'s [=attribution trigger/trigger data noise probability=].
+1. If |source|'s [=attribution source/source type=] is "<code>event</code>":
```
csharrison, line 866: Rather than branching on the type, can we look at the actual value of the sources noise probability and omit if it is zero? This would allow us to add noise to clicks without changing this logic at all.

csharrison (commented):
Two high level comments

index.bs
apasel422: Just checking my logic for this: If the source's `attribution mode` was chosen to be `never`, then a report would never have been generated for it. So given that the report exists (by virtue of the reporting origin receiving the report), then we know that the source's `attribution mode` wasn't never. So shouldn't the probability only be `truthfully` vs `falsely`?

index.bs
apasel422: I don't think this logic will have to be changed if we add noise to clicks. This is just ensuring that the correct per-source-type trigger data and trigger data noise probability are being propagated to the report properly.

apasel422: Or are you saying that once we've determined that the source won't be attributed truthfully, that never and fake should be equally likely?

csharrison: Ah I see, I was missing some logic. We have separate trigger data / probabilities for event and nav sources in the data model. Was there a strong reason to separate those out that way?

I think thinking that this logic could be simplified if we just had single fields in `attribution trigger`s for the trigger data noise probability and the trigger data itself. These fields would be set by branching on the source type, but in report generation we could just process them as-is.

index.bs
csharrison: The easiest way for me to think about the flip mechanism is like this:

1. Flip a coin with heads probability `p`
2. If the coin is heads --> report the event truthfully
3. If the coin is tails --> pick a random output with uniform probability

The choice of `never` is included in (3).

The mechanism as described here is:
1. Flip a coin with heads probability `p`
2. If heads --> Set the output to `never`
3. Else
    1. Flip another coin with heads probability `q`
    2. If heads --> report truthfully
    3. Else pick a random output (other than `never`) with uniform probability

The thing that is confusing here is that the initial `p` probability is very constrained by 3.3 since we want `never` to be picked just as likely as any of the other output states.

index.bs
apasel422: We can't have single fields for trigger data noise probability and trigger data in the `attribution trigger` struct, because we don't know what the source type is until we've looked up matching sources, which requires the trigger itself.

csharrison: Ah you're right, that's what I was missing, thanks.

index.bs
apasel422: > If the coin is tails --> pick a random output with uniform probability

Is `truthfully` included in those potential outputs, or does it include only the two non-`truthfully` outputs of `never` and `falsely`?

I think it makes sense for me to reword the spec in a separate PR before continuing with this one.

csharrison (approved):
As discussed offline, let's add "navigation" to "randomized trigger data rate" to align with the source events in a follow-up.

PR #279, opened 2021-12-15
Reformulate noising as randomized responses

apasel422:
***

### :boom: Error: 500 Internal Server Error :boom: ###

[PR Preview](https://github.com/tobie/pr-preview#pr-preview) failed to build. _(Last tried on Jan 6, 2022, 8:10 PM UTC)_.

<details>
<summary>More</summary>

PR Preview relies on a number of web services to run. There seems to be an issue with the following one:

:rotating_light: [CSS Spec Preprocessor](https://api.csswg.org/bikeshed/) - CSS Spec Preprocessor is the web service used to build Bikeshed specs.

:link: [Related URL](https://api.csswg.org/bikeshed/?url=https%3A%2F%2Fraw.githubusercontent.com%2Fapasel422%2Fconversion-measurement-api%2Fd5960ff3fadb15c7f51c63d1950557cfd45699dd%2Findex.bs&force=1&md-warning=not%20ready)

```
<!DOCTYPE HTML PUBLIC "-//IETF//DTD HTML 2.0//EN">
<html><head>
<title>500 Internal Server Error</title>
</head><body>
<h1>Internal Server Error</h1>
<p>The server encountered an internal error or
misconfiguration and was unable to complete
your request.</p>
<p>Please contact the server administrator at 
 [no address given] to inform them of the time this error occurred,
 and the actions you performed just before this error.</p>
<p>More information about this error may be available
in the server error log.</p>
<hr>
<address>Apache/2.4.10 (Debian) Server at api.csswg.org Port 443</address>
</body></html>

```

_If you don't have enough information above to solve the error by yourself (or to understand to which web service the error is related to, if any), please [file an issue](https://github.com/tobie/pr-preview/issues/new?title=Error%20not%20surfaced%20properly&body=See%20WICG/conversion-measurement-api%23279.)._
</details>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -499,10 +499,10 @@ To <dfn>obtain an event attribution source from a source</dfn> given an [=attrib
 1. Set |resultSource|'s [=attribution source/attribution mode=] to the result of
     [=obtaining a randomized response=] with "`truthfully`",
     «"`truthfully`", "`never`", "`falsely`"», and the user agent's
-    [=event source false trigger probability=].
+    [=event-source false trigger probability=].
 1. Return |resultSource|.
 
-<dfn>Event source false trigger probability</dfn> is a vendor-specific double between 0 and 1 (both
+<dfn>Event-source false trigger probability</dfn> is a vendor-specific double between 0 and 1 (both
 inclusive) that controls the randomized response probability for the
 [=attribution source/attribution mode=] of an [=attribution source=] whose
 [=attribution source/source type=] is "`event`".
--- a/index.bs
+++ b/index.bs
@@ -516,10 +516,10 @@ To <dfn>obtain an event attribution source from a source</dfn> given an [=attrib
 1. Set |resultSource|'s [=attribution source/attribution mode=] to the result of
     [=obtaining a random attribution mode=] with the user agent's
     [=event-source trigger data cardinality=] and the user agent's
-    [=event-source randomized trigger probability=].
+    [=randomized event-source trigger rate=].
 1. Return |resultSource|.
 
-<dfn>Event-source randomized trigger probability</dfn> is a vendor-specific double between 0 and 1 (both
+<dfn>Randomized event-source trigger rate</dfn> is a vendor-specific double between 0 and 1 (both
 inclusive) that controls the randomized response probability for the
 [=attribution source/attribution mode=] of an [=attribution source=] whose
 [=attribution source/source type=] is "`event`".
--- a/index.bs
+++ b/index.bs
@@ -470,7 +470,7 @@ To <dfn>parse an attribution destination</dfn> from a string |str|:
 <h3 algorithm id="obtaining-random-attribution-mode">Obtaining a random attribution mode</h3>
 
 To <dfn>obtain a random attribution mode</dfn> given a positive integer |triggerDataCardinality|
-and a double |randomizedPickRate|:
+and a double |randomPickRate|:
 
 1. Let |possibleValues| be the [=set=] of [=attribution modes=]
     « (`"truthfully"`, null), (`"never"`, null) ».
--- a/index.bs
+++ b/index.bs
@@ -678,7 +678,7 @@ To <dfn>obtain an attribution trigger</dfn> given a [=URL=] |url| and an
 [=attribution trigger/trigger data=]:
 0 <= [=attribution trigger/trigger data=] < [=trigger data cardinality=].
 
-<dfn>Event-source trigger cardinality</dfn> is a vendor-specific integer that controls the valid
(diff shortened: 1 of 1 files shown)


csharrison (commented):
A few big comments

index.bs
```diff
@@ -483,11 +496,17 @@ To <dfn>obtain an event attribution source from a source</dfn> given an [=attrib
 1. Set |resultSource|'s [=attribution source/source identifier=] to a new unique opaque string.
 1. Set |resultSource|'s [=attribution source/source type=] to "<code>event</code>".
 1. Round |resultSource|'s [=attribution source/expiry=] away from zero to the nearest day (86400 seconds).
-1. Set |resultSource|'s [=attribution source/attribution mode=] to one of
-    "<code>truthfully</code>", "<code>never</code>", or "<code>falsely</code>", with
-    [=implementation-defined=] probability.
+1. Set |resultSource|'s [=attribution source/attribution mode=] to the result of
+    [=obtaining a randomized response=] with "`truthfully`",
+    «"`truthfully`", "`never`", "`falsely`"», and the user agent's
```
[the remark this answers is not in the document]
apasel422: I extracted a separate algorithm for the attribution mode, though I'm still not sure it's correct. PTAL.

index.bs
```diff
@@ -302,6 +302,12 @@ This specification defines a [=policy-controlled feature=] identified by the str
 
 # Structures # {#structures}
 
+<h3 dfn-type=dfn>Attribution mode</h3>
+
+An attribution mode is a [=tuple=] consisting of a mode and a random trigger data. The mode is
```
csharrison, line 307: This works for views because a report is completely described by its trigger data. This is OK for now but it's worth thinking about a future where we need more parameters like the report time.

One thing to consider is having a tuple of (mode, {set of attribution reports}) where the set of attribution reports are all of the fake reports to be generated by a particular source.

index.bs
```diff
@@ -447,6 +466,17 @@ To <dfn>parse an attribution destination</dfn> from a string |str|:
 1. Return the result of [=obtain a site|obtaining a site=] from |url|'s
     [=url/origin=].
 
+<h3 algorithm id="obtaining-random-attribution-mode">Obtaining a random attribution mode</h3>
+
+To <dfn>obtain a random attribution mode</dfn> given a positive integer |triggerDataCardinality|
+and a double |randomPickRate|:
+
+1. Let |possibleValues| be the [=set=] of [=attribution modes=] « (`"never"`, null) ».
```
csharrison, line 474: Consider pulling this into a separate algorithm, as it may get more complicated if more parameters are added in addition to the trigger data.

index.bs
```diff
@@ -656,13 +690,21 @@ To <dfn>obtain an attribution trigger</dfn> given a [=URL=] |url| and an
...
+[=attribution trigger/trigger data=]:
+0 <= [=attribution trigger/trigger data=] < [=trigger data cardinality=].
 
-<dfn>Max event source trigger data value</dfn> is a vendor-specific integer which controls the potential values of [=attribution trigger/event source trigger data=].
+<dfn>Event-source trigger data cardinality</dfn> is a vendor-specific integer that controls the valid
+range of [=attribution trigger/event source trigger data=]:
+0 <= [=attribution trigger/event source trigger data=] < [=event-source trigger data cardinality=].
 
-<dfn>Trigger data noise rate</dfn> is a vendor-specific double between 0 and 1 (both inclusive) which controls noising of [=attribution trigger/trigger data=].
+<dfn>False trigger data probability</dfn> is a vendor-specific double between 0 and 1 (both
```
csharrison, line 701: We should name these definitions to align with how we'd expose them externally in the API. I don't think they are quite right currently "false trigger data" implies this will always return something that is not correct but this is not always the case - there is a chance that the randomized output matches the true value by luck.

Some ideas:
- Randomized trigger data {probability, rate}
- Trigger data randomized response rate
- Trigger data random output rate
- etc

cc @johnivdel

csharrison (commented):
This is exactly what I had in mind, it looks great. Just a few small comments. I think we can also do some bikeshedding of names here to prep for exposing some of these values.

index.bs
apasel422: I think we can add that complexity when it's needed.

index.bs
apasel422: Seems like we could wait to do that in a followup if it's needed.

index.bs
csharrison: Yep totally fine adding it later, just wanted to highlight it.

csharrison (approved):

index.bs
apasel422: Renamed to `randomized trigger data rate`.

index.bs
csharrison: SGTM

PR #281, opened 2022-01-04
Rename dedup-key to deduplication-key

apasel422:
Fixes #280.

PR #282, opened 2022-01-10
Rename dedup-key and report_id to client-deduplication-key and server-deduplication-key, respectively

apasel422:
Fixes #281

***
<a href="https://pr-preview.s3.amazonaws.com/apasel422/conversion-measurement-api/pull/282.html" title="Last updated on Jan 10, 2022, 5:14 PM UTC (eeb9f3a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/conversion-measurement-api/282/b6d5f63...apasel422:eeb9f3a.html" title="Last updated on Jan 10, 2022, 5:14 PM UTC (eeb9f3a)">Diff</a>

### DIFF
--- a/event_attribution_reporting_clicks.md
+++ b/event_attribution_reporting_clicks.md
@@ -194,12 +194,12 @@ this API:
 to trigger attribution for all matching sources.
 
 The browser will treat redirects to a URL of the form:
-`https://<attributionreportto>/.well-known/attribution-reporting/trigger-attribution[?trigger-data=<trigger-data>&priority=<priority>&dedup-key=<dedup-key>]`
+`https://<attributionreportto>/.well-known/attribution-reporting/trigger-attribution[?trigger-data=<trigger-data>&priority=<priority>&client-deduplication-key=<client-deduplication-key>]`
 
 as a special request. The query string for the request contains additional information about the attribution trigger:
 - `trigger-data`: optional data to identify the triggering event
 - `priority`: optional signed 64-bit integer representing the priority of this trigger compared to other triggers for the same source.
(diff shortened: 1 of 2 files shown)


johnivdel:
I feel like `trigger-deuplication-key` may be a better fit than `client-deduplication-key` if we are going with verbose names anyway. Client is a bit vague as it is not clear what is actually being deduped.

`server-deduplication-key` seems reasonable to me. Another consideration would be `report-deduplication-key` as it allows reports to be deduplicated against each other.

cc @csharrison

apasel422:
> I feel like `trigger-deuplication-key` may be a better fit than `client-deduplication-key` if we are going with verbose names anyway. Client is a bit vague as it is not clear what is actually being deduped.
> 
> `server-deduplication-key` seems reasonable to me. Another consideration would be `report-deduplication-key` as it allows reports to be deduplicated against each other.
> 
> cc @csharrison

`trigger-deduplication-key` and `report-deduplication-key` seem fine to me, though arguably both prefixes are redundant in the contexts in which they occur: `trigger-deduplication-key` is a query param in the trigger redirect and `report-deduplication-key` is a JSON key in the report body. For the former, I guess we do also have `trigger-data` instead of just `data`, but we use `priority` instead of `trigger-priority`, even though that is easily confusable with source priority.

csharrison:
The reason I like client/local and server/remote is because it indicates the entity responsible for actually doing the deduplication. I think that's a very nice property of the name that is removed when you just talk about the trigger vs. the report.

I think if we remove that reference, I am concerned the names become too similar even though they have very different semantics (one is set by the ad-tech and processed by the browser, and the other is the opposite), and I'd go back to preferring e.g. `report_id` separate from the deduplication key.

PR #285, opened 2022-01-20
Hard wrap a few explainers to 80 chars

csharrison:


csharrison:
Github says no visible changes so I think nothing broke.

PR #289, opened 2022-01-27
Use jsonc for json with comments

csharrison:


AGGREGATE.md
```diff
@@ -183,7 +183,7 @@ trusted aggregation service known by the browser.
 
 The report will be JSON encoded with the following scheme:
 
-```
+```jsonc
 {
   "source_site": "https://publisher.example",
   "attribution_destination": "https://advertiser.example",
```
johnivdel, line 189: Highlighting below is a bit funny as they are being highlighting as JSON rather than values.

But I think that is moot as both of these should be encoded as strings anyway as they represent 64 bit unsigned ints.

"source_registration_time": "[timestamp in seconds]"
“report_id”: "[64 bit unsigned integer]"

EVENT.md
```diff
@@ -597,7 +597,7 @@ associated with the 7-day deadline as the 2-day deadline has passed. Roughly 5
 days later, `ad-tech.example` receives the following HTTP POST to
 `https://ad-tech.example/.well-known/attribution-reporting/report-attribution`
 with the following body:
-```json
+```jsonc
```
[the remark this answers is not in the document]
csharrison: Yes, missed one

AGGREGATE.md
csharrison: Done

johnivdel:
slgtm

PR #299, opened 2022-02-02
[aggregate] move report_id to shared_info

csharrison:
This will allow the aggregation service to dedupe duplicate reports.

### DIFF
--- a/AGGREGATE.md
+++ b/AGGREGATE.md
@@ -191,7 +191,7 @@ The report will be JSON encoded with the following scheme:
 
   // Info that the aggregation services also need encoded in JSON
   // for use with AEAD.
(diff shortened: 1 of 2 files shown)


linnan-github:
LGTM.  Do we also expect other consumer API to have a unique report id? Thanks!

csharrison:
> LGTM. Do we also expect other consumer API to have a unique report id? Thanks!

I would expect this field would be useful in other APIs that send delayed reports that we would want to have built-in support for retries.

linnan-github (approved):

linnan-github:
Another question is for event-level report, the report id is just a unique id (GUID). Shall we be consistent with aggregatable report instead of enforcing it to be 64-bit unsigned integer?

csharrison:
> Another question is for event-level report, the report id is just a unique id (GUID). Shall we be consistent with aggregatable report instead of enforcing it to be 64-bit unsigned integer?

Great point, just updated the CL to update this in both event and agg explainers. PTAL?

PR #302, opened 2022-02-03
Remove single string support from filters

linnan-github:
There's not a strong reason to support a single string as a shorthand of a list with one element, therefore proposing to remove the shorthand support to simplify the schema and implementation.

This can be added back if specific need arises.


csharrison (approved):
LGTM

PR #307, opened 2022-02-08
Add source_registration_time to privacy budget key

johnivdel:
Note we only add it to the true key used to track the batch, not the key itself as it is already visible in cleartext. For this reason, we move it into the shared info subobject. 

This will allow the reporting origin to access summary reports split by the time the ad was shown without wasting privacy budget.

### DIFF
--- a/AGGREGATE.md
+++ b/AGGREGATE.md
@@ -187,12 +187,11 @@ The report will be JSON encoded with the following scheme:
 {
   "source_site": "https://publisher.example",
   "attribution_destination": "https://advertiser.example",
-  "source_registration_time": "[timestamp in seconds]",
   "report_id": "[64 bit unsigned integer]",
 
   // Info that the aggregation services also need encoded in JSON
   // for use with AEAD.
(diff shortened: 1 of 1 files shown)


AGGREGATE.md
```diff
@@ -237,20 +233,25 @@ utilize techniques like retries to minimize data loss.
...
+* The `source_registration_time` within `shared_info` will represent (in seconds 
+  since the Unix Epoch) the time the source event was registered, rounded to the
+  nearest whole day.
+  
+* The `privacy_budget_key` is used to define distinct batches of aggregate
+  reports. It is used by the aggregation service to prevent replay attacks. It
+  will be a hash of: `reporting_origin | source_site | destination | version`.
+  Note that the true key used to track batches will be `privacy_budget_key`
+  concatenated with `round_to_hour(scheduled_report_time)` and 
+  `round_to_hour(source_registration_time`. The latter two are omitted from the key
```
csharrison, line 251: close paren

csharrison, line 251: source_registration_time should already be rounded to the nearest day, so we can just use it directly

csharrison (approved):

PR #308, opened 2022-02-08
scheduled_report_time should be a string

alexmturner:
Avoids year 2038 problems


csharrison (approved):
LGTM but I think there will be a merge conflict with my recently closed PR

PR #311, opened 2022-02-09
Just use hex for key_piece definition

csharrison:
Dealing with binary, while conceptually simpler, is practically more difficult to integrate with existing systems. Additionally, the ergonomics of having an existing "offset" parameter are arguably not worth it, as this can be easily achieved with bit-shifting operators during header generation.

Let's keep this API as simple as possible for now.

### DIFF
--- a/AGGREGATE.md
+++ b/AGGREGATE.md
@@ -73,31 +73,25 @@ on the response to the `attributionsrc` request:
 list.
 ```jsonc
 [{
-  // Generates a "101011001" key prefix named "campaignCounts"
+  // Generates a "0x159" key prefix named "campaignCounts"
   "id": "campaignCounts",
-  "key_piece": "101011001", // User saw ad from campaign 345 (out of 511)
-  "key_offset": 0
+  "key_piece": "0x159", // User saw ad from campaign 345 (out of 511)
 },
 {
-  // Generates a "101" key prefix named "geoValue"
+  // Generates a "0x5" key prefix named "geoValue"
   "id": "geoValue",
-  // Source-side geo region = 5 (US) but pad with 0s since the shop operates in
-  // ~100 separate countries.
-  "key_piece": "0000101",
-  "key_offset": 0
+  // Source-side geo region = 5 (US).
+  "key_piece": "0x5",
 }]
 ```
 This defines a list named histogram contributions, each with a piece of the
-aggregation key defined as a bit-string at a particular offset. The final
-histogram bucket key will be fully defined at trigger time using a combination
-of this piece and trigger-side pieces.
+aggregation key defined as a hex-string. The final histogram bucket key will be
+fully defined at trigger time using a combination (binary OR) of this piece and
+trigger-side pieces.
 
 Final keys will be restricted to a maximum of 128 bits. Keys longer than this
-will be truncated.
-
-TODO: consider it an option to use binary or decimal notation e.g. with 0b
-prefix. Also consider using a single parameter to specify the key vs. a
-string piece and an offset.
+will be truncated. This means that hex strings in the JSON should be limited to
+at most 32 digits.
 
 ### Attribution trigger registration
 
@@ -108,14 +102,12 @@ which generates aggregation keys.
 [
 // Each dict independently adds pieces to multiple source keys.
 {
-  "key_piece": "10",// Conversion type purchase = 2
-  "key_offset": 9,
+  "key_piece": "0x400",// Conversion type purchase = 2 at a 9 bit offset
   // Apply this suffix to:
   "source_keys": ["campaignCounts"]
 },
 {
-  "key_piece": "10101",// Purchase category shirts = 21
-  "key_offset": 7,
+  "key_piece": "0xA80",// Purchase category shirts = 21 at a 7 bit offset
   // Apply this suffix to:
   "source_keys": ["geoValue", "nonMatchingKeyIdsAreIgnored"]
 }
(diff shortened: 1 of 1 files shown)


AGGREGATE.md
```diff
@@ -73,31 +73,25 @@ on the response to the `attributionsrc` request:
 list.
 ```jsonc
 [{
-  // Generates a "101011001" key prefix named "campaignCounts"
+  // Generates a "0x159" key prefix named "campaignCounts"
   "id": "campaignCounts",
-  "key_piece": "101011001", // User saw ad from campaign 345 (out of 511)
-  "key_offset": 0
+  "key_piece": "0x159", // User saw ad from campaign 345 (out of 511)
```
linnan-github, line 78: Shall we say (out of 4095 since it's 4*3=12 bits)?

AGGREGATE.md
```diff
@@ -108,14 +102,12 @@ which generates aggregation keys.
...
-  "key_piece": "10",// Conversion type purchase = 2
-  "key_offset": 9,
+  "key_piece": "0x400",// Conversion type purchase = 2 at a 9 bit offset
   // Apply this suffix to:
   "source_keys": ["campaignCounts"]
 },
 {
-  "key_piece": "10101",// Purchase category shirts = 21
-  "key_offset": 7,
+  "key_piece": "0xA80",// Purchase category shirts = 21 at a 7 bit offset
```
alexmturner, line 110: nit: do you think something like `= 21 (0b10101)` to spell out the binary might be useful? This might help connect the idea of a bit offset and the resulting contributions at the bottom.

AGGREGATE.md
```diff
@@ -150,12 +142,12 @@ The scheme above will generate the following abstract histogram contributions:
 [
 // campaignCounts
 {
-  key: 1382, // 0b10101100110
+  key: 0x566, // 0b10101100110
```
linnan-github, line 145: It looks like the final key is incorrect for the given example.
source_key: 0x159
trigger_key: 0x400
final_key: 0x559

The difference from the original design is that currently the offset is from the lowest bit whereas it's from the highest bit before.

AGGREGATE.md
```diff
@@ -150,12 +142,12 @@ The scheme above will generate the following abstract histogram contributions:
...
   value: 32768 
 },
 // geoValue:
 {
-  key: 181, // 0b000010110101
+  key: 0xB5, // 0b000010110101
```
linnan-github, line 150: source_key: 0x5
trigger_key: 0xA80
final_key: 0xA85

AGGREGATE.md
palenica: It would be lovely to be able to support bit shift operations: 21 << 7 :)
(not realistic)

csharrison: > It would be lovely to be able to support bit shift operations: 21 << 7 :)
> (not realistic)

It is realistic, it is just spelled "offset" in the existing documentation :) I can update comment here to spell out more clearly how these were generated.

AGGREGATE.md
csharrison: Oh, great catch! Let me update all the comments here...

AGGREGATE.md
csharrison: Done (this one should have been obvious to me... :) )

AGGREGATE.md
csharrison: No, because 511 campaigns is why we need a 9 bit offset below. I clarified all this in the comments

AGGREGATE.md
```diff
@@ -73,31 +73,25 @@ on the response to the `attributionsrc` request:
 list.
 ```jsonc
 [{
-  // Generates a "101011001" key prefix named "campaignCounts"
+  // Generates a "0x159" key prefix named "campaignCounts"
```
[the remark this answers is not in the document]
csharrison: I clarified that this is the key piece specifying just the lower-order bits of the kye

AGGREGATE.md
```diff
@@ -108,14 +103,18 @@ which generates aggregation keys.
 [
 // Each dict independently adds pieces to multiple source keys.
 {
-  "key_piece": "10",// Conversion type purchase = 2
-  "key_offset": 9,
+  // Conversion type purchase = 2 at a 9 bit offset, i.e. 2 << 9.
+  // 9 bit offset is needed because there are 511 possible campaigns, which
+  // will take up 9 bits in the resulting key.
+  "key_piece": "0x400",
   // Apply this suffix to:
```
linnan-github, line 110: This is no longer suffix?

AGGREGATE.md
```diff
@@ -108,14 +103,18 @@ which generates aggregation keys.
...
   "source_keys": ["campaignCounts"]
 },
 {
-  "key_piece": "10101",// Purchase category shirts = 21
-  "key_offset": 7,
+  // Purchase category shirts = 21 at a 7 bit offset, i.e. 21 << 7.
+  // 7 bit offset needed because there are 100 regions for the geo key, which
```
linnan-github, line 115: "7 bit offset is needed"
"~100" to be consistent with comment in the source part.

AGGREGATE.md
csharrison: Done

AGGREGATE.md
csharrison: Done

linnan-github (approved):

johnivdel (approved):
lgtm

One implication of this format is that if a reporting origin is setting overlapping parts of the key by accident the browser can't really detect it. With the previous format, the browser could detect this and potentially message the error. However with the debug keys, it should be easier for reporting origins to back track and see what went wrong. 

That being said, I think the ergonomic benefits here outweigh the potential future flexibility.

palenica (approved):

csharrison:
Thanks everyone for the thoughtful review.

PR #312, opened 2022-02-10
Mention debug_mode flag in shared_info

palenica:
Mention that if debugging is enabled, debug_mode=1 will be set in shared_info to indicate to a downstream aggregation service that it's OK to treat this report as a debug report.

### DIFF
--- a/AGGREGATE.md
+++ b/AGGREGATE.md
@@ -255,8 +255,8 @@ Note: if [debugging](https://github.com/WICG/conversion-measurement-api/blob/mai
 is enabled, additional debug fields will be present in aggregatable reports,
 including the cleartext payloads, to allow downstream systems to verify that
 reports are constructed correctly. If debugging is enabled, the `shared_info` 
-will include the flag `debug_mode=true` to let the aggregation service know
-that debugging functionality can be enabled when processing this record.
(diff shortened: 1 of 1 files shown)


AGGREGATE.md
```diff
@@ -254,7 +254,9 @@ utilize techniques like retries to minimize data loss.
 Note: if [debugging](https://github.com/WICG/conversion-measurement-api/blob/main/EVENT.md#optional-extended-debugging-reports)
 is enabled, additional debug fields will be present in aggregatable reports,
 including the cleartext payloads, to allow downstream systems to verify that
-reports are constructed correctly.
+reports are constructed correctly. If debugging is enabled, the `shared_info` 
+will include the flag `debug_mode=true` to let the aggregation service know
```
csharrison, line 258: It should be in JSON so "debug_mode: 1" or something like that

csharrison (commented):
LGTM % nits

AGGREGATE.md
csharrison: still should be `debug_mode: 1` to be valid.

palenica: should be fixed now?

palenica: oops, now new and improved, with a colon

csharrison (approved):

PR #313, opened 2022-02-10
Add pending source and attribution limits to explainer

apasel422:


EVENT.md
```diff
@@ -711,6 +716,14 @@ Additionally, there should be a limit of 10 reporting origins per <source site,
 destination site, 30 days>, counted for every attribution report that is
 generated.
 
+### Pending source limits
+
+There should be a limit of 1024 pending sources per source origin.
+
+### Pending report limits
+
+There should be a limit of 1024 pending reports per destination site.
```
apasel422, line 725: @csharrison I'm not actually clear on the purpose of these limits. Are they just to ensure that a single source origin / destination site isn't taking up too much storage space? Or are there actually privacy concerns here as well?

csharrison: Yes that's right, they are just for storage limits. Let's explain that here and also possibly just condense this into a single section like "storage limits".

csharrison (approved):

csharrison:
It might be worthwhile to surface these in the security section at the bottom about exhausting budgets.

PR #314, opened 2022-02-12
Flesh out the noise corrector

### LINKED ISSUE #301: Noise correction: provide a utility function
Filing this issue for future reference. This originally came up in the [noise correction script review](https://github.com/WICG/conversion-measurement-api/pull/293#discussion_r798299541).

To use the current [noise correction script](https://github.com/WICG/conversion-measurement-api/blob/main/noise_corrector.py), a developer needs to provide specific inputs; in the current version of the script, these inputs are not straightforward to put together from raw reports and require some understanding of complex concepts like [stars and bars](https://en.wikipedia.org/wiki/Stars_and_bars_(combinatorics)).

Consider providing a utility function that:
* Takes as **input** raw reports⏤in the shape the reporting origin would actually receive them
* Provides as **output** report count arrays, that would have the right shape to be used as input to `correct_event_sources` or `correct_nav_sources` in [noise_corrector.py](https://github.com/WICG/conversion-measurement-api/blob/main/noise_corrector.py).

This could ease developer experience by bridging the gap that currently exists between the data the reporting endpoint would receive, and what `noisy_reports` (the array expected by the noise correction script) actually should look like to be consumed properly by `correct_event_sources` or `correct_nav_sources`.

csharrison:
This PR massively overhauls the noise corrector. It has turned into a fully fledged command line utility that takes as input JSON files of sources and reports.

Given JSON inputs of event-level sources and reports, It provides two types of outputs:
1. Aggregate output: The tool aggregates event-level reports and reports de-biased histograms broken out by the trigger_data, in an unbiased way.
2. Event-level output: The tool synthesizer adjusted event-level data to match (as best as possible) the de-biased aggregates

Fixes #301

Note this utility is designed to be interoperable with the JSON formats in the WIP [Chromium API simulator](https://source.chromium.org/chromium/chromium/src/+/main:tools/attribution_reporting/) in the following way:
```
    +-----------+ +------------+
    |Source JSON| |Trigger JSON|
    +-----+---+-+ +-----+------+
          |   |         |
          |  +v---------v--+
          |  |             |
          |  |  Simulator  |
          |  |             |
          |  +-------+-----+
          |          |
          |    +-----v-----+
          |    |Report JSON|
          |    +--+--------+
          |       |
  +-------v-------v-+
  |                 |
  | Noise Corrector |
  |                 |
  +-----+----------++
        |          |
+-------v---+ +----v-------+
| Corrected | | Corrected  |
| Aggregates| | Report JSON|
+-----------+ +------------+
```
Edit: given crrev.com/c/3459831 it should also be possible for the simulator to input data directly to the noise corrector too.

### DIFF
--- a/noise_corrector.py
+++ b/noise_corrector.py
@@ -47,7 +47,7 @@ JoinedList = List[Joined]
 OutputConfig = Tuple[Tuple[int, int], ...]
 
 # For generic functions.
-T = TypeVar("T")
+T = TypeVar('T')
 
 
 class ParamConfig:
@@ -64,8 +64,8 @@ class ParamConfig:
     self.name = name
 
 
-NAV_PARAMS = ParamConfig(.5, 8, 3, 3, "navigation")
-EVENT_PARAMS = ParamConfig(.0000025, 2, 1, 1, "event")
+NAV_PARAMS = ParamConfig(.5, 8, 3, 3, 'navigation')
+EVENT_PARAMS = ParamConfig(.0000025, 2, 1, 1, 'event')
 
 
 class OutputEnumeration:
@@ -73,15 +73,15 @@ class OutputEnumeration:
   # per source event, for the purposes of debiasing the randomized response
   # privacy mechanism.
   def __init__(self, report_tuple: OutputConfig, params: ParamConfig):
-    """report_tuple must be in the form ((window_index, trigger_data),...)"""
+    '''report_tuple must be in the form ((window_index, trigger_data),...)'''
     self.output: OutputConfig = tuple(sorted(report_tuple))
     self.params = params    
 
   def __str__(self):
     def report_to_string(r):
       if r is None:
-        return "No report"
-      return {"window": r[0], "data": r[1]}
+        return 'No report'
+      return {'window': r[0], 'data': r[1]}
     return json.dumps([report_to_string(r) for r in self.output])
 
   def __repr__(self): return self.output.__repr__()
@@ -97,20 +97,20 @@ class OutputEnumeration:
 
   @classmethod
   def create_from_data(cls, source_with_reports):
-    """Create an output directly from structured data
+    '''Create an output directly from structured data
 
     source_with_reports: list of (source, [report, ...]) tuples.
     See get_conversion_counts for expected structure for sources and reports.
-    """
+    '''
     source, reports = source_with_reports
-    params = EVENT_PARAMS if source["source_type"] == "event" else NAV_PARAMS
+    params = EVENT_PARAMS if source['source_type'] == 'event' else NAV_PARAMS
     max_reports = params.max_reports
     assert len(reports) <= max_reports
 
     def parse_report(source, report: dict) -> Tuple[int, int]:
       window_index = OutputEnumeration._get_window_index_for_report(
           source, report)
-      trigger_data = int(report["report"]["trigger_data"])
+      trigger_data = int(report['report']['trigger_data'])
       return (window_index, trigger_data)
 
     return cls(tuple(parse_report(source, r) for r in reports), params)
@@ -128,8 +128,8 @@ class OutputEnumeration:
     return sorted([cls(x, params) for x in deduped])
 
   def data_histogram(self) -> numpy.ndarray:
-    """Returns the number of reports associated with this output, broken
-    out by the trigger_data"""
+    '''Returns the number of reports associated with this output, broken
+    out by the trigger_data'''
 
     histogram = numpy.zeros(self.params.data_cardinality)
     for r in self.output:
@@ -137,39 +137,39 @@ class OutputEnumeration:
     return histogram
 
   def get_report_time(self, window: int, source: Source) -> int:
-    expiry = source["registration_config"]["expiry"]
-    if self.params.name == "event":
+    expiry = source['registration_config']['expiry']
+    if self.params.name == 'event':
       return expiry
 
     # TODO(csharrison): This assumes the expiry > 7 days.
     windows = [timedelta(days=2), timedelta(days=7), timedelta(seconds=expiry)]
-    source_time = source["source_time"]
+    source_time = source['source_time']
     return int(source_time + windows[window].total_seconds())
 
   def generate_reports_for_source(self, source: Source) -> List[Report]:
-    assert self.params.name == source["source_type"]
-    source_event_id = source["registration_config"]["source_event_id"]
+    assert self.params.name == source['source_type']
+    source_event_id = source['registration_config']['source_event_id']
 
     reports: List[Report] = []
     for window, trigger_data in self.output:
       reports.append({
-          "report_time": self.get_report_time(window, source),
-          "report": {
-              "source_event_id": source_event_id,
-              "source_type": self.params.name,
-              "trigger_data": str(trigger_data)
+          'report_time': self.get_report_time(window, source),
+          'report': {
+              'source_event_id': source_event_id,
+              'source_type': self.params.name,
+              'trigger_data': str(trigger_data)
           }
       })
     return reports
 
   @staticmethod
   def _get_window_index_for_report(source: dict, report: dict) -> int:
-    if source["source_type"] == "event":
+    if source['source_type'] == 'event':
       return 0
 
-    report_time = report["report_time"]
-    source_time = source["source_time"]
-    raw_expiry = source["registration_config"]["expiry"]
+    report_time = report['report_time']
+    source_time = source['source_time']
+    raw_expiry = source['registration_config']['expiry']
     expiry = timedelta(seconds=int(raw_expiry))
     delta = timedelta(seconds=report_time - source_time)
 
@@ -188,14 +188,14 @@ class OutputEnumeration:
 # response. The estimator should be unbiased.
 def estimate_true_values(observed_values: Dict[T, int],
                          randomized_response_rate: float) -> Dict[T, float]:
-  """Returns a list of corrected counts
+  '''Returns a list of corrected counts
 
   observed_values: A map of size `k` for k-ary randomized response. Each item
                    represents a randomized response output as the key, with its
                    associated count as the value.
 
   randomized_response_rate: The probability any given output is randomized
-  """
+  '''
   n = sum(observed_values.values())
   k = len(observed_values)
   x = randomized_response_rate
@@ -213,10 +213,10 @@ def estimate_true_values(observed_values: Dict[T, int],
 
 
 def join_reports_with_sources(sources: List[Source], reports: List[Report]) -> JoinedList:
-  """Returns tuple of (source, [reports]) joined with the source_event_id"""
+  '''Returns tuple of (source, [reports]) joined with the source_event_id'''
   report_map = collections.defaultdict(list)
   for r in reports:
-    report_map[r["report"]["source_event_id"]].append(r)
+    report_map[r['report']['source_event_id']].append(r)
 
   def join(source) -> Joined:
     source_event_id = source['registration_config']['source_event_id']
@@ -227,8 +227,8 @@ def join_reports_with_sources(sources: List[Source], reports: List[Report]) -> J
 ################## AGGREGATE UTILITIES ##################
 
 def get_raw_corrected_map(joined: JoinedList, params: ParamConfig) -> Dict[OutputEnumeration, float]:
-  """Returns an aggregate map of OutputEnumeration -> float, with noise debiased
-  according to params"""
+  '''Returns an aggregate map of OutputEnumeration -> float, with noise debiased
+  according to params'''
   output_counts = {o: 0 for o in OutputEnumeration.generate_all(params)}
   for j in joined:
     output = OutputEnumeration.create_from_data(j)
@@ -239,14 +239,14 @@ def get_raw_corrected_map(joined: JoinedList, params: ParamConfig) -> Dict[Outpu
 
 
 def correct_aggregates(joined: JoinedList, params: ParamConfig) -> List[dict]:
-  """Returns debiased report counts broken out by trigger_data"""
+  '''Returns debiased report counts broken out by trigger_data'''
   corrected = get_raw_corrected_map(joined, params)
   histogram = numpy.zeros(params.data_cardinality)
   for o, c in corrected.items():
     histogram += c * o.data_histogram()
   x = 10
-  return [{"trigger_data": t,
-           "report_count": v} for t, v in enumerate(histogram)]
+  return [{'trigger_data': t,
+           'report_count': v} for t, v in enumerate(histogram)]
 
 
 ################## EVENT-LEVEL UTILITIES ##################
@@ -254,7 +254,7 @@ def correct_aggregates(joined: JoinedList, params: ParamConfig) -> List[dict]:
 def adjust_to_match_distribution(values: List[T],
                                  estimated_counts: Dict[T, float],
                                  default_value: T) -> List[T]:
-  """Given a list of values, and an aggregate map that counts all possible values,
+  '''Given a list of values, and an aggregate map that counts all possible values,
   returns a list of adjusted values to match the aggregate distribution.
 
   values: list of arbitrary objects (must be possible to place in dict/set). To
@@ -268,9 +268,9 @@ def adjust_to_match_distribution(values: List[T],
   The algorithm used in this method attempts to achieve two goals:
   1. The output values match the distribution as much as possible, minimizing
      bias.
-  2. The output maximizes "overlap" with the input. That is, satisfying (1), we
+  2. The output maximizes 'overlap' with the input. That is, satisfying (1), we
      want to minimize the number of inputs that are adjusted.
-  """
+  '''
 
   # Split the distribution into two chunks: elements with mass >= 1 and not.
   # Clamp negatives to 0. Note this introduces bias, but it seems unavoidable.
@@ -330,7 +330,7 @@ def generate_corrected_event_level(joined: JoinedList, params: ParamConfig) -> J
 
 
 def main():
-  DESCRIPTION = """\
+  DESCRIPTION = '''\
   noise_corrector.py is a command-line tool that attempts to debias the noise
   output in the Attribution Reporting API for event-level reports. It takes as
   input JSON files which represent source registrations and the resulting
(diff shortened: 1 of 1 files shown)


csharrison:
@pasin30055 would you take a look at the sampling logic (and anything else you thing is relevant)?

noise_corrector.py
```diff
@@ -71,149 +207,229 @@ def estimate_true_values(observed_values, randomized_response_rate, beta=0):
...
+                                 estimated_counts: Dict[T, float],
+                                 default_value: T) -> List[T]:
+  '''Given a list of values, and an aggregate map that counts all possible values,
+  returns a list of adjusted values to match the aggregate distribution.
+
+  values: list of arbitrary objects (must be possible to place in dict/set). To
+          minimize bias, this list should be shuffled prior to calling this
+          method.
+  estimated_counts: map of value -> estimated count. All values in `values` must be
+                    present. Note that negatives are allowed, but it will mean the
```
pasin30055, line 264: I think this part is slightly misleading, at least as far as I understand the "bias claim". My understanding (and please correct me if I'm wrong) is that we view the entire algorithm, including randomization, debiasing the aggregates, and the adjustment, as a single process. In this case, the end value will always be biased since we are not allowed to assign anything less than zero. As a result, it seems a bit misleading to say that it is only biased for the negative estimates case. Wdyt?

pasin30055 (commented):
Thanks Charlie. The overall logic in `adjust_to_match_distribution` LGTM. (I guess there are multiple valid choices about e.g. handling `default_value` or too much mass, but the current handling seems good.)

noise_corrector.py
csharrison: I think this is true, I've tried rewording it to match your comment:

```
Note that negatives are allowed, but they will be
clamped to 0 internally. Note that because the we cannot
assign any output less than 0, the end values will always be
biased
```
Does that make sense to you?

csharrison:
@apasel422 can you take a look at the whole thing? In particular, I am wondering about how to best compose this with the simulator. It's tough for this utility to take all the input from stdin since it needs two separate streams. It would be nice if this could be put in a pipeline ahead of the simulator though.

apasel422:
> @apasel422 can you take a look at the whole thing? In particular, I am wondering about how to best compose this with the simulator. It's tough for this utility to take all the input from stdin since it needs two separate streams. It would be nice if this could be put in a pipeline ahead of the simulator though.

Per your diagram, do you mean `after the simulator`? Seems like the simulator output is being piped into the noise corrector. That said, one thing we could do is have a flag in the simulator that when set copies the input into the output, e.g.

```sh
attribution_simulator --copy_input <path/to/input.json | noise_simulator --output_mode=event-level
```

The simulator's output in that case would be

```js
{
  "input": {
    "sources": [/*...*/],
    "triggers": [/*...*/],
  },
  "reports": [/*...*/]
}
```

noise_corrector.py
pasin30055: Yes the new explanation LGTM. Thanks!

csharrison:
> > @apasel422 can you take a look at the whole thing? In particular, I am wondering about how to best compose this with the simulator. It's tough for this utility to take all the input from stdin since it needs two separate streams. It would be nice if this could be put in a pipeline ahead of the simulator though.
> 
> Per your diagram, do you mean `after the simulator`? Seems like the simulator output is being piped into the noise corrector.

Ah yes, "ahead" is ambiguous. I meant after :)

> That said, one thing we could do is have a flag in the simulator that when set copies the input into the output, e.g.
> 
> ```shell
> attribution_simulator --copy_input <path/to/input.json | noise_simulator --output_mode=event-level
> ```
> 
> The simulator's output in that case would be
> 
> ```js
> {
>   "input": {
>     "sources": [/*...*/],
>     "triggers": [/*...*/],
>   },
>   "reports": [/*...*/]
> }
> ```

That solution sounds pretty good to me! We could amend this utility to expect input in the above form in stdin (in the JSON-lines format the simulator outputs). I can do that in this CL. Unfortunately we don't win much from a concurrency POV as we still need to basically read the whole input before acting on it (especially in the event-level case), but I think it's a good idea ergonomically at least.

apasel422:
> > > @apasel422 can you take a look at the whole thing? In particular, I am wondering about how to best compose this with the simulator. It's tough for this utility to take all the input from stdin since it needs two separate streams. It would be nice if this could be put in a pipeline ahead of the simulator though.
> > 
> > 
> > Per your diagram, do you mean `after the simulator`? Seems like the simulator output is being piped into the noise corrector.
> 
> Ah yes, "ahead" is ambiguous. I meant after :)
> 
> > That said, one thing we could do is have a flag in the simulator that when set copies the input into the output, e.g.
> > ```shell
> > attribution_simulator --copy_input <path/to/input.json | noise_simulator --output_mode=event-level
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > The simulator's output in that case would be
> > ```js
> > {
> >   "input": {
> >     "sources": [/*...*/],
> >     "triggers": [/*...*/],
> >   },
> >   "reports": [/*...*/]
> > }
> > ```
> 
> That solution sounds pretty good to me! We could amend this utility to expect input in the above form in stdin (in the JSON-lines format the simulator outputs). I can do that in this CL. Unfortunately we don't win much from a concurrency POV as we still need to basically read the whole input before acting on it (especially in the event-level case), but I think it's a good idea ergonomically at least.

Merged in https://chromium-review.googlesource.com/c/chromium/src/+/3459831.

csharrison:
Okay this should be ready for re-review. I tested the tool using the updated simulator output that @apasel422 just landed.

noise_corrector.py
```diff
@@ -1,66 +1,204 @@
...
+
+    return cls(tuple(parse_report(source, r) for r in reports), params)
+
+  @classmethod
+  def generate_all(cls, params: ParamConfig):
+    all_single_reports = list(itertools.product(range(params.max_windows),
+                                                range(params.data_cardinality)))
+    all_reports_ordered = []
+    for r in range(params.max_reports + 1):
+      reports = itertools.product(*(r * [all_single_reports]))
```
pasin30055, line 126: Sorry I don't think I quite parse this line. What is this supposed to do?

noise_corrector.py
```diff
@@ -1,66 +1,204 @@
...
+      all_reports_ordered.extend(list(reports))
+    # Simulate shuffling within a reporting window by sorting and deduping
+    deduped = set([tuple(sorted(x)) for x in all_reports_ordered])
```
pasin30055, line 129: This seems a bit inefficient given that we may generate some state multiple times (and then dedupe). Wouldn't it be possible to just directly generate the reports that we want without duplication? For example, I had written some code below a while back (the parameter names are different but the logic should be similar). Would this work?

```
def all_possible_reports(
    data_cardinality, max_windows, max_reports, index=0, report_so_far=[]):
  yield report_so_far
  if max_reports == 0:
    return
  for next_index in range(index, data_cardinality * max_windows):
    window, data = divmod(next_index, data_cardinality)
    report_so_far.append({"window": window, "data": data})
    yield from all_possible_reports(
        data_cardinality,
        max_windows,
        max_reports - 1,
        index=next_index,
        report_so_far=report_so_far)
    report_so_far.pop()
```

noise_corrector.py
```diff
@@ -1,66 +1,204 @@
...
+    return sorted([cls(x, params) for x in deduped])
+
+  def data_histogram(self) -> numpy.ndarray:
+    '''Returns the number of reports associated with this output, broken
+    out by the trigger_data'''
+
+    histogram = numpy.zeros(self.params.data_cardinality)
```
pasin30055, line 136: Does `numpy.bincount` also work here?

noise_corrector.py
csharrison: r * [all_single_reports] --> [(all_single_reports), (all_single_reports),...]
The prefix star means that we invoke the product as if we wrote this:

itertools.product(all_single_reports, all_single_reports, ...)

csharrison: Not relevant anymore

pasin30055 (approved):
Thanks Charlie. This LGTM.

noise_corrector.py
```diff
@@ -71,149 +216,238 @@ def estimate_true_values(observed_values, randomized_response_rate, beta=0):
...
+    adjusted_reports: List[Report] = adjusted.generate_reports_for_source(
+        source)
+    return (source, adjusted_reports)
+
+  return [adjust_json(j, a) for j, a in zip(joined, adjusted_values)]
+
+
+def main():
+  DESCRIPTION = '''\
+  noise_corrector.py is a command-line tool that attempts to debias the noise
```
maudnals, line 345: Change suggestion:
Could we unify the language and always say "debias" ? Or "correct"?
I find "correct" easier to understand (but maybe "debias" is more accurate).

noise_corrector.py
```diff
@@ -1,66 +1,211 @@
...
-  print("{:<30}{:<20}{:<20}{:<20}".format(*column_names))
-  for i, v in enumerate(corrected_reports):
-    d = json.dumps(outputs[i])
-    print(f"{d:<30}{true_reports[i]:<20,}{noisy_reports[i]:<20,}{v:<20,.1f}")
+################## GENERIC UTILITIES ##################
+
+# Useful type declarations for working with JSON:
+class SourceRegistrationConfig(TypedDict, total=False):
+  source_event_id: str
+  expiry: int # Optional
```
maudnals, line 22: Why are the other configs, like priorities, not needed here?

noise_corrector.py
```diff
@@ -1,66 +1,211 @@
...
+
+NAV_PARAMS = ParamConfig(.0024, 8, 3, 3, 'navigation')
+EVENT_PARAMS = ParamConfig(.0000025, 2, 1, 1, 'event')
+
+DEFAULT_EXPIRY = 60 * 60 * 24 * 30
+
+class OutputEnumeration:
+  # Class encapsulating logic which represents the logical output of the API
+  # per source event, for the purposes of debiasing the randomized response
+  # privacy mechanism.
```
maudnals, line 75: Nit for clarity:
```suggestion
  # Class encapsulating logic which represents all possible outputs of the API
  # per source event, taking into account the randomized response
  # privacy mechanism.
```
(is it?)

noise_corrector.py
```diff
@@ -1,66 +1,211 @@
...
+    yield from gen_all(params.max_reports, 0, [])
+
+
+  def data_histogram(self) -> numpy.ndarray:
+    '''Returns the number of reports associated with this output, broken
+    out by the trigger_data'''
+
+    return numpy.bincount([o[1] for o in self.output], minlength=self.params.data_cardinality)
+
+  def get_report_time(self, window: int, source: Source) -> int:
```
maudnals, line 148: Change suggestion:
Could we explain why `get_report_time` is needed?

noise_corrector.py
```diff
@@ -71,149 +216,238 @@ def estimate_true_values(observed_values, randomized_response_rate, beta=0):
...
+  for o, c in corrected.items():
+    histogram += c * o.data_histogram()
+  x = 10
+  return [{'trigger_data': t,
+           'report_count': v} for t, v in enumerate(histogram)]
+
+
+################## EVENT-LEVEL UTILITIES ##################
+
+def adjust_to_match_distribution(values: List[T],
```
maudnals, line 263: Is `adjust_to_match_distribution` something we expect devs to use directly?
I'm confused with the concept of aggregate distribution in the context of event-level reports.

maudnals (commented):
Took a pass!
TLDR:
* Awesome to see that the corrector can be called with reports that are close to the API output, this is a really great improvement.
* Example code on how to use this would be very helpful / needed. E.g., what are the exact functions within this file we expect developers to call?
* I'm unclear on the aggregate feature/mode in this utility, I'm hoping that an example would help clarify this.
* Do we expect all/most teams who try the API to use this utility? If so, it may be valuable to invest in putting together a short `.md` doc file that would contain: what does this script correct exactly, an example (with an input and output), etc. Inline documentation sometimes hits its limit here, since this script and output is quite complex. This could be a follow-up AI to this PR, as I know that for now the priority is to make this script available to developers. If you agree this is valuable, I'm happy to contribute to this.

csharrison:
Hey Maud, thanks for the review, will respond to the inline comments shortly. Let me just answer some of these larger questions first.

> * Example code on how to use this would be very helpful / needed. E.g., what are the exact functions within this file we expect developers to call?

I've refactored the tool such that the most developers doesn't need to worry about the underlying implementation details. They should just use the tool as a command line utility. The inline documentation (i.e. not the help text) should be for "experts" who are looking to modify functionality or better understand how the tool works.

> * I'm unclear on the aggregate feature/mode in this utility, I'm hoping that an example would help clarify this.
> * Do we expect all/most teams who try the API to use this utility? If so, it may be valuable to invest in putting together a short `.md` doc file that would contain: what does this script correct exactly, an example (with an input and output), etc. Inline documentation sometimes hits its limit here, since this script and output is quite complex. This could be a follow-up AI to this PR, as I know that for now the priority is to make this script available to developers. If you agree this is valuable, I'm happy to contribute to this.

I think this could be useful to many teams, though depending on some noise values it might not be strictly necessary (e.g. strawman noise for event sources is fairly low in the existing document). I think what we could do is just put this in a new directory with it's own readme, with more fleshed out documentation in a follow-up. Does that sound good to you?

maudnals:
Thanks Charlie!

> The inline documentation (i.e. not the help text) should be for "experts" who are looking to modify functionality or better understand how the tool works.

Understood, that makes sense. 
I still see the value of a quick "how to run this" and the various inputs/output (as you did for the next question) but as agreed we can make this a follow-up.

>  In the meantime I can work on fleshing out the help text to more clearly demonstrate what the output represents. It returns output in the form...

The example is super helpful. I think this clarifies what aggregate is about.

> What we could do is just put this in a new directory with it's own readme, with more fleshed out documentation in a follow-up. Does that sound good to you?

SGTM

noise_corrector.py
csharrison: Went with "debiased" since I think it's more accurate.

noise_corrector.py
```diff
@@ -71,149 +216,238 @@ def estimate_true_values(observed_values, randomized_response_rate, beta=0):
...
+  '''
+  parser = argparse.ArgumentParser(
+      description=DESCRIPTION, formatter_class=argparse.RawTextHelpFormatter)
+
+  INPUT_MODE_HELP = '''\
+  Optional. One of `single` or `multi`. Defaults to `single`, which reads all of
+  stdin as one JSON file, with the format above. `multi` mode accepts input in
+  the JSON-lines format, where every separate line of input follows the above
+  format. It is expected in `multi` mode that events from a given browser are
+  isolated in a particular line of input.
```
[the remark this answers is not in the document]
csharrison: They are basically the same, `multi` mode is just a way to accept streaming data e.g. the output of the simulator in multi mode:
https://source.chromium.org/chromium/chromium/src/+/main:tools/attribution_reporting/simulator_main.cc;drc=88c60a87db572d12f00ec95e6d6f4905c8643b20;l=90

I am avoiding mentioning the simulator in the tool until it is more fleshed out.

noise_corrector.py
```diff
@@ -71,149 +216,238 @@ def estimate_true_values(observed_values, randomized_response_rate, beta=0):
...
+  '''
```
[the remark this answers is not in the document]
csharrison: ack

noise_corrector.py
```diff
@@ -71,149 +216,238 @@ def estimate_true_values(observed_values, randomized_response_rate, beta=0):
...
+  parser.add_argument('--input_mode', dest='input_mode',
+                      choices=['single', 'multi'],
+                      default='single', help=INPUT_MODE_HELP)
+
+  OUTPUT_MODE_HELP = '''\
+  Optional. One of `event-level` or `aggregate`. Defaults to `aggregate`, which
+  provides aggregate counts for different trigger metadata values. The
```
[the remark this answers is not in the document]
csharrison: Done

noise_corrector.py
```diff
@@ -71,149 +216,238 @@ def estimate_true_values(observed_values, randomized_response_rate, beta=0):
...
+  `event-level` mode outputs synthetic sources and their associated reports
```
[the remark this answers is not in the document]
csharrison: Done

noise_corrector.py
csharrison: No, we don't expect them to use this directly.  I tried to explain this more in the help text

csharrison (commented):
Thanks Maud. I also updated in the most recent patch to call out `event-level` mode as experimental.

noise_corrector.py
```diff
@@ -1,66 +1,211 @@
 #!/usr/bin/env python3
 
+import argparse
```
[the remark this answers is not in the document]
maudnals: Sweet 👍

noise_corrector.py
maudnals: Got it.
Does it mean that for the input to be valid, developers _have_ to strip down priorities? Or will this just be ignored by the tool? Just so we flag it.
I'm thinking of the case where the input would be real sources, so IIUC by default `registration_config` would be default contain whatever I had there at registration time

noise_corrector.py
```diff
@@ -1,66 +1,211 @@
...
 import collections
+from datetime import timedelta
 import json
-import math
 import numpy
```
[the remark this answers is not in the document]
maudnals: > the event-level mode takes that aggregate debiased data and produces synthetic event-level data based off of it
What do you think the main use case would be for these? Simulation?

noise_corrector.py
maudnals: Gotcha, thank you for clarifying; whichever you think works best (the "enumeration" term was indeed why I was confused)

noise_corrector.py
maudnals: Sounds good; it would be ideal to reflect this other places as well for clarity. Nothing blocking for this PR though.

noise_corrector.py
maudnals: Nice, I think it's helpful 👍

noise_corrector.py
```diff
@@ -71,149 +216,238 @@ def estimate_true_values(observed_values, randomized_response_rate, beta=0):
...
+  def join(source) -> Joined:
+    source_event_id = source['registration_config']['source_event_id']
+    return (source, report_map[source_event_id])
+  return (join(s) for s in sources)
+
+
+################## AGGREGATE UTILITIES ##################
+
+def get_raw_corrected_map(joined: Iterable[Joined], params: ParamConfig) -> Dict[OutputEnumeration, float]:
+  '''Returns an aggregate map of OutputEnumeration -> float, with noise debiased
```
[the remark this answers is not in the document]
maudnals: Great this helps

noise_corrector.py
```diff
@@ -71,149 +227,275 @@ def estimate_true_values(observed_values, randomized_response_rate, beta=0):
...
+            // Optional int64 in milliseconds formatted as a base-10 string.
+            // Defaults to 30 days.
+            'expiry': '864000000',
+          }
+        },
+        ...
+      ]
+    },
+    // List of zero or more reports.
+    reports: [
```
apasel422, line 392:
```suggestion
    "reports": [
```

noise_corrector.py
```diff
@@ -71,149 +227,275 @@ def estimate_true_values(observed_values, randomized_response_rate, beta=0):
...
+          // whether the source is registered on click or on view, respectively.
+          'source_type': 'navigation',
+
+          'registration_config': {
+            // Required uint64 formatted as a base-10 string.
+            'source_event_id': '123456789',
+
+            // Optional int64 in milliseconds formatted as a base-10 string.
+            // Defaults to 30 days.
+            'expiry': '864000000',
```
apasel422, line 385: All these single-quoted strings should be double-quoted instead. Same in the report JSON below.

apasel422 (approved):

maudnals (approved):
LGTM!

noise_corrector.py
csharrison: ack

noise_corrector.py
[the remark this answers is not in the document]
csharrison: It's a good question that I don't know the full answer to. I think it could be used for some data processing that involves more than just one aggregate (e.g. generating data to run multiple aggregate queries off of so they are consistent), or to integrate with existing code that would take a lot of effort to port over to accept aggregate data.

This is something that we should continue to explore with @pasin30055. For now, because we don't have a firm use-case I am going to mark the mode as experimental.

noise_corrector.py
csharrison: Done

noise_corrector.py
csharrison: Done

csharrison:
Thanks everyone, landing!

PR #320, opened 2022-02-15
Make trigger deduplication keys unsigned

### LINKED ISSUE #316: Consider giving debug keys and dedup keys the same type
Right now, debug keys are unsigned 64-bit integers, but dedup keys are signed 64-bit integers. I'm not sure there's a reason that the latter is signed.

apasel422:
For consistency with debug keys.

Fixes #316.


johnivdel (approved):
lgtm

PR #338, opened 2022-02-17
Fix formatting / clarity about hiding the # of encrypted reports

csharrison:


AGGREGATE.md
```diff
@@ -492,10 +492,12 @@ count of reports to the reporting origin could leak something sensitive as well
...
 for a single user).
 
-To hide the true number of reports, we could: Unconditionally send a null report
-for every registered attribution trigger (thus making the count a function of
-only destination-side information) Add noise to the number of reports by having
-some clients randomly add noisy reports
+To hide the true number of reports, we could: 
+- Unconditionally send a null report for every registered attribution trigger
+  (thus making the count a function of only destination-side information)
+- Add noise to the number of reports by having some clients randomly add noisy
```
johnivdel, line 498: Can we clarify these are null reports? Maybe: having some clients randomly send null reports some fraction of the time

"noisy null reports" could also work

johnivdel (approved):
lgtm with nit

AGGREGATE.md
csharrison: Done

PR #340, opened 2022-02-18
Rename source_data to filter_data in source registration JSON

apasel422:
The entire registration JSON is arguably "source data", so let's make it clear that the contents of this field are specific to filtering.


johnivdel (approved):
lgtm. Agreed that "source data" on its own is overloaded and ambiguous.

PR #341, opened 2022-02-18
Remove redundant prefixes from source and trigger priority JSON fields

apasel422:
These fields are already contained in a source- or trigger-registration JSON document, so the `source_` and `trigger_` prefixes are redundant.


csharrison (commented):
We should do the same with `trigger_priority` to keep them aligned if this is the direction we want to go.

csharrison (approved):

PR #342, opened 2022-02-22
Use more specific URL for debug reports and clarify debug report bodies

apasel422:
Using a URL other than `.well-known/attribution-reporting/debug` is more extensible.


johnivdel (approved):
We should also clarify if these early reports are sent for aggregate, but that is a separate issue.

lgtm

PR #348, opened 2022-02-28
Add alternative aggregation mode

alexmturner:


alexmturner:

AGGREGATE.md
```diff
@@ -520,6 +521,26 @@ contribution to any one bucket). Care should be taken to ensure that either:
...
 
+### Choosing among aggregation services
+
+The server can respond with an optional header
+`Attribution-Reporting-Alternative-Aggregation-Mode` which accepts a string
+value.
+
+```http
+Attribution-Reporting-Register-Aggregatable-Source: [{....}]
+Attribution-Reporting-Alternative-Aggregation-Mode: "experimental-poplar"
```
schoppmp, line 532: Is this meant to include all DPF-based aggregation? I think "poplar" unnecessarily narrows it down to the heavy-hitters application that poplar solves. Maybe "experimental-dpf", "experimental-mpc" or "experimental-secret-shares"?

AGGREGATE.md
```diff
@@ -520,6 +521,26 @@ contribution to any one bucket). Care should be taken to ensure that either:
...
+```
+
+The optional header will allow developers to choose among different options for
+aggregation infrastructure supported by the user agent. This value will allow
+experimentation with new technologies, and allows us to try out new approaches
+without interfering with core functionality provided by the default option. The
+`"experimental-poplar"` option will implement a protocol similar to
+[poplar](https://github.com/cjpatton/vdaf/blob/main/draft-patton-cfrg-vdaf.md#poplar1-poplar1)
+VDAF in the
+[PPM Framework](https://datatracker.ietf.org/doc/draft-gpew-priv-ppm/).
```
schoppmp, line 542: Should we link to the concrete [DPF implementation](https://github.com/google/distributed_point_functions)?

AGGREGATE.md
johnivdel: I think this is meant as one example of a specific protocol, rather than expecting all services which use DPFs from sharing a common setting. I support leaving as-is for now

johnivdel (approved):

Issue #347, opened 2022-02-25
Consider waiving `attributionsrc` for triggers


maudnals:
Question/Feedback from developer:

> Requiring `attributionsrc` on conversion registrations is problematic for existing image tags. 
> It may make it difficult for existing tags to test the API. 
> Based on this [TODO](https://github.com/WICG/conversion-measurement-api/blob/main/EVENT.md#triggering-attribution) (_"Finalize whether the `attributionsrc` attribute should be optional to support legacy HTML tags for triggers"_), is it possible to waive this requirement for conversion registration?

maudnals:
@johnivdel

johnivdel:
Some of the implications here:

This would dis-align the permissions model on the source and trigger side. With attributionsrc, we have a tag-level opt-in to the API. Processing headers without the attribute would be similar to the previous version, all subresources on the page would be able to register triggers.

On the source side, this permissive model would make it easy to abuse limits on unattributed sources (such as the 100 reporting origin limit). On the trigger side, we don't have any rate-limits which operate on unattributed triggers, so there is less concern for DOS style attacks.

We do have limits on attributed triggers, but this is already contingent on an action from a publisher site, and these limits include reporting origin in the scope.

Based on this, I think it would be reasonable to remove the requirement to support pre-existing conversion tags.

PR #360, opened 2022-03-14
Resolve todo on supporting legacy conversion pings

johnivdel:
Removes the attributionsrc requirement for trigger registration, fixing #347

### DIFF
--- a/EVENT.md
+++ b/EVENT.md
@@ -255,9 +255,9 @@ window.attributionReporting.registerAttributionTrigger(
 ```
 
 As a stop-gap to support pre-existing conversion tags which do not include the 
-`attributionsrc` attribute, or use a different Fetch API, the browser will also process 
-trigger registration headers for all subresource requests on the page where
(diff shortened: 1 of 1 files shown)


johnivdel:
@apasel422 PTAL

EVENT.md
```diff
@@ -255,9 +255,9 @@ window.attributionReporting.registerAttributionTrigger(
 ```
 
 As a stop-gap to support pre-existing conversion tags which do not include the 
-`attributionsrc` attribute, or use a different Fetch API, the browser will also process 
-trigger registration headers for all subresource requests on the page where
-the `attribution-reporting` Permissions Policy is enabled. 
+`attributionsrc` attribute, or use a different Fetch API, the browser will also 
+process  trigger registration headers for all subresource requests on the page
```
apasel422, line 259:
```suggestion
process trigger registration headers for all subresource requests on the page 
```

apasel422 (approved):

PR #366, opened 2022-03-16
Remove speculative <a> view API surface

johnivdel:
There is an existing TODO which speculates whether anchor tags supporting view source registration in addition to a click source is useful/necessary functionality.

This is naturally possible by instead including two separate tags:

```
<img attributionsrc="...register-view" style="display:none"/>
<a href="..." attributionsrc="..register-click"></a>
```

The speculative API to register views on anchor tags involves adding a new attribute, and a new request header to allow sites to differentiate between duplicate requests to the same url.

It makes more sense to frame this TODO as an ask for whether this API should be supported, rather than including it by default.

### DIFF
--- a/EVENT.md
+++ b/EVENT.md
@@ -129,7 +129,7 @@ window.open(
  `<img>` tags with the new `attributionsrc` attribute too:
 ```html
 <!-- TODO: is view registration via an `<a>` tag necessary?
-    It requires a new attribute and a way to differentiate the requests. -->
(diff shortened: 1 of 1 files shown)


johnivdel:
Note that this entirely removes the `Attribution-Reporting-Source-Info` header. It does not make sense to send this header for  attributionsrc's, as these tags may either register a source or a trigger which would be confusing.

apasel422 (approved):

PR #372, opened 2022-03-25
Remove comma to make JSON valid

linnan-github:


linnan-github:
johnivdel, could you PTAL, thanks!

PR #380, opened 2022-04-08
Clarify event-level debug report description

alexmturner:


alexmturner:
@apasel422 Low-priority: could you PTAL? Tried to clarify here, but feel free to reject/suggest changes if you don't think it's clearer (or it's wrong).

apasel422 (approved):

PR #389, opened 2022-04-28
Specify top-level filter matching

apasel422:
But do not set top-level trigger filters to anything but the empty map for now.

We will add negated filter matching in a followup, as top-level filters may only be positive.

#386

***
<a href="https://pr-preview.s3.amazonaws.com/apasel422/conversion-measurement-api/pull/389.html" title="Last updated on Apr 28, 2022, 2:02 PM UTC (e8e515a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/conversion-measurement-api/389/c4231c8...apasel422:e8e515a.html" title="Last updated on Apr 28, 2022, 2:02 PM UTC (e8e515a)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -308,6 +308,11 @@ An attribution mode is a [=tuple=] consisting of a mode and a random trigger dat
 either "`truthfully`", "`never`", or "`falsely`". The random trigger data is a non-negative 64-bit
 integer, if the mode is "`falsely`", or null otherwise.
 
+<h3 dfn-type=dfn>Attribution filter map</h3>
+
+An attribution filter map is an [=ordered map=] whose [=map/key|keys=] are [=strings=] and whose
+[=map/value|values=] are [=ordered sets=] of [=strings=].
+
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -732,6 +748,21 @@ Given a [=request=] |request|:
 1. [=Queue a task=] to [=trigger attribution=] with |trigger|.
 1. Return <strong>blocked</strong>.
 
+<h3 dfn id="does-filter-data-match">Does filter data match</h3>
```
johnivdel, line 751: Do you envision this and the other filtering related algorithms all going in the triggering algorithms section? Should we add a different section dedicated to filtering?

johnivdel (approved):
lgtm with question on the structure

index.bs
apasel422: Filtering only happens during triggering/attribution, so yes, I think they'd all go in the triggering section.

PR #402, opened 2022-05-03
Specify event-level trigger configurations

apasel422:
https://github.com/WICG/conversion-measurement-api/blob/main/EVENT.md#triggering-attribution
https://github.com/WICG/conversion-measurement-api/blob/main/EVENT.md#optional-attribution-filters

#386

***
<a href="https://pr-preview.s3.amazonaws.com/apasel422/conversion-measurement-api/pull/402.html" title="Last updated on May 11, 2022, 12:47 PM UTC (28a0fa3)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/conversion-measurement-api/402/3335cf0...apasel422:28a0fa3.html" title="Last updated on May 11, 2022, 12:47 PM UTC (28a0fa3)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -344,7 +344,7 @@ An attribution source is a [=struct=] with the following items:
 : <dfn>number of reports</dfn>
 :: Number of [=attribution reports=] created for this [=attribution source=].
 : <dfn>dedup keys</dfn>
-:: [=ordered set=] of [=attribution trigger/dedup keys=] associated with this [=attribution source=].
+:: [=ordered set=] of [=event-level trigger configuration/dedup keys=] associated with this [=attribution source=].
 : <dfn>randomized response</dfn>
 :: A [=randomized source response=].
 : <dfn>randomized trigger rate</dfn>
@@ -390,6 +390,26 @@ An attribution aggregatable trigger is a [=struct=] with the following items:
 
 </dl>
 
+<h3 dfn-type=dfn>Event-level trigger configuration</h3>
+
+An event-level trigger configuration is a [=struct=] with the following items:
+
+<dl dfn-for="event-level trigger configuration">
+: <dfn>trigger data</dfn>
+:: A non-negative 64-bit integer.
+: <dfn>dedup key</dfn>
+:: Null or a non-negative 64-bit integer.
+: <dfn>priority</dfn>
+:: A 64-bit integer.
+: <dfn>filters</dfn>
+:: An [=attribution filter map=].
+
+</dl>
+
+Issue: Add an [=attribution filter map=] field called `negated filters` to
+[=event-level trigger configuration=] and update [=trigger attribution=] to use
+it.
+
 <h3 dfn-type=dfn>Attribution trigger</h3>
 
 An attribution trigger is a [=struct=] with the following items:
@@ -397,22 +417,16 @@ An attribution trigger is a [=struct=] with the following items:
 <dl dfn-for="attribution trigger">
 : <dfn>attribution destination</dfn>
 :: A [=site=].
-: <dfn>navigation-source trigger data</dfn>
-:: A non-negative 64-bit integer.
-: <dfn>event-source trigger data</dfn>
-:: A non-negative 64-bit integer.
 : <dfn>trigger time</dfn>
 :: A point in time.
 : <dfn>reporting endpoint</dfn>
 :: An [=url/origin=].
-: <dfn>dedup key</dfn>
-:: Null or a 64-bit integer.
-: <dfn>priority</dfn>
-:: A 64-bit integer.
 : <dfn>filters</dfn>
 :: An [=attribution filter map=].
 : <dfn>debug key</dfn>
 :: Null or a non-negative 64-bit integer.
+: <dfn>event-level trigger configurations</dfn>
+:: A [=set=] of [=event-level trigger configuration=].
 : <dfn>aggregatable trigger</dfn>
 :: An [=attribution aggregatable trigger=].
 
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -975,17 +1024,19 @@ return a point in time.
 
 <h3 algorithm id="obtaining-a-report">Obtaining a report</h3>
 
-To <dfn>obtain a report</dfn> given an [=attribution source=] |source| and an [=attribution trigger=] |trigger|:
+To <dfn>obtain a report</dfn> given an [=attribution source=] |source|, an [=attribution trigger=]
+|trigger|, and an [=event-level trigger configuration=] |config|:
```
johnivdel, line 1028: (see above regarding config)

johnivdel (commented):
mostly lg

index.bs
```diff
@@ -664,26 +677,30 @@ To <dfn>process an attribution source</dfn> given an [=attribution source=] |sou
...
-        :: 0
         : [=attribution trigger/filters=]
         :: «[]»
         : [=attribution trigger/debug key=]
         :: null
-    1. Let |fakeReport| be the result of running [=obtain a report=] with |source| and |fakeTrigger|.
+        : [=attribution trigger/event-level trigger configurations=]
+        :: « |fakeConfig| »
+    1. Let |fakeReport| be the result of running [=obtain a report=] with |source|, |fakeTrigger|,
+        and |fakeConfig|.
```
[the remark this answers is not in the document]
apasel422: Because `obtain a report` doesn't know which the matching config is.

index.bs
```diff
@@ -701,11 +718,13 @@ To <dfn>obtain an attribution trigger</dfn> given a [=URL=] |url| and an
 [=environment settings object=] |environment|, run the following steps:
 
 1. Let |navigationSourceTriggerData| be 0.
-1. If |url|'s [=url/query=] has a "`trigger-data`" field, set |navigationSourceTriggerData| to the result of running [=parse attribution data=] with
-    the value associated with the field modulo the user agent's [=trigger data cardinality=].
+1. If |url|'s [=url/query=] has a "`trigger-data`" field, and applying the <a spec="html">rules for
```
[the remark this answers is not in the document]
apasel422: Will address this later.

index.bs
apasel422: Because this algorithm needs to know which config is matching.

index.bs
```diff
@@ -697,26 +710,30 @@ To <dfn>process an attribution source</dfn> given an [=attribution source=] |sou
...
         agent's [=max report cache size=], return.
+    1. Let |fakeConfig| be a new [=event-level trigger configuration=] with the items:
+        : [=event-level trigger configuration/trigger data=]
+        :: |source|'s [=attribution source/attribution mode=][1]
+        : [=event-level trigger configuration/dedup key=]
+        :: null
+        : [=event-level trigger configuration/priority=]
+        :: 0
+        : [=event-level trigger configuration/filters=]
+        :: «[ "`source_type`" → "`event`" ]»
```
csharrison, line 721: This doesn't type check, since a the attribution filter map's values must be an ordered set of strings. I think this should be
```
«[ "`source_type`" → « "`event`" » ]»
```

index.bs
```diff
@@ -747,35 +766,48 @@ To <dfn>obtain an attribution trigger</dfn> given a [=URL=] |url| and an
...
     in an integer, set |triggerPriority| to that value.
+1. Let |navigationConfig| be a new [=event-level trigger configuration=] with the items:
+    : [=event-level trigger configuration/trigger data=]
+    :: |navigationSourceTriggerData|
+    : [=event-level trigger configuration/dedup key=]
+    :: |dedupKey|
+    : [=event-level trigger configuration/priority=]
+    :: |triggerPriority|
+    : [=event-level trigger configuration/filters=]
+    :: «[ "`source_type`" → "`navigation`" ]»
```
csharrison, line 777: ditto for the type issue here and on line 786

index.bs
```diff
@@ -380,29 +380,43 @@ An attribution aggregatable trigger is a [=struct=] with the following items:
...
-: <dfn>dedup key</dfn>
-:: Null or a 64-bit integer.
-: <dfn>priority</dfn>
-:: A 64-bit integer.
 : <dfn>filters</dfn>
 :: An [=attribution filter map=].
 : <dfn>debug key</dfn>
 :: Null or a non-negative 64-bit integer.
+: <dfn>event-level trigger configurations</dfn>
+:: A [=set=] of [=event-level trigger configurations=].
```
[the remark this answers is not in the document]
csharrison: nit: A [=set=] of [=event-level trigger configuration=] (i.e. drop the plural, otherwise it's a circular definition)

index.bs
```diff
@@ -923,10 +955,26 @@ To <dfn>trigger attribution</dfn> given an [=attribution trigger=] |trigger| run
 1. Assert: |sourceToAttribute|'s [=attribution source/attribution mode=][0] is
     "`truthfully`" or "`never`".
 1. If the result of running
-    [=match an attribution source's filter data against an attribution trigger=] with
-    |sourceToAttribute| and |trigger| is false, return.
-1. If |trigger|'s [=attribution trigger/dedup key=] is not null and |sourceToAttribute|'s
-    [=attribution source/dedup keys=] [=list/contains=] it, return.
+    [=match an attribution source's filter data against filters=] with
+    |sourceToAttribute| and |trigger|'s [=attribution trigger/filters=] is false, return.
+1. Let |matchedConfig| be a new [=event-level trigger configuration=] with the items:
```
csharrison, line 960: I'm wondering if this default config is necessary. Maybe out of scope for this change but it seems like if none of the configs match we should not create a report.

cc @johnivdel

index.bs
apasel422: Ah, good catch.

index.bs
apasel422: That would be a behavioral change from https://github.com/WICG/conversion-measurement-api/blob/main/EVENT.md#optional-attribution-filters and the current implementation. We have top-level filters that exhibit that behavior, though.

csharrison: Ah, you're right. I think we need this if we want the filters to be an optional thing. Sorry for the noise!

johnivdel (approved):

Issue #407, opened 2022-05-04
attributionsrc set by window.open is broken with respect to URL parameters


apasel422:
The inability to use `=` within a feature parameter for `window.open` is a limitation of both the [specification](https://html.spec.whatwg.org/multipage/window-object.html#concept-window-open-features-tokenize) and [Chromium's implementation](https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/core/page/create_window.cc;l=69;drc=1d6e4b5c91adb7417f80626a9c87cd15c312d28b). For example, `attributionsrc=https://a.test?x=y&h=j` gets interpreted as:

```
attributionsrc = https://a.test?x
y&h = j
```

Some solutions we could consider:
1. Modify the `window.open` spec and implementation to permit escaping/quoting of values.
2. Require `attributionsrc` feature values to be URL-encoded and perform URL-decoding on them before using them.
3. Stop using feature strings altogether and specify `attributionsrc` via a separate parameter to `window.open`.

Originally reported in https://groups.google.com/a/chromium.org/g/attribution-reporting-api-dev/c/_Q_dYzbSo38.

alois-bissuel:
Thanks for tracking this here. 
All mentioned solutions are fine for us (with maybe a preference for 2 and 3).
Meanwhile we will encode the query part of the URL, so that the equal signs are escaped, and we will decode it in our endpoint.

apasel422:
I am leaning toward option 2 myself, as it doesn't require option 1's modification of the HTML spec (with the backwards-compatibility concerns that might entail) and doesn't require adding a completely separate API surface as in option 3.

PR #408, opened 2022-05-04
Encode attributionsrc URLs passed to window.open in event explainer

apasel422:
#407

### DIFF
--- a/EVENT.md
+++ b/EVENT.md
@@ -119,6 +119,8 @@ are two types of attribution sources, `navigation` sources and `event` sources.
 or via calls to `window.open` that occur with [transient
 activation](https://html.spec.whatwg.org/multipage/interaction.html#transient-activation):
 ```javascript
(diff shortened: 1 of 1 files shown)


EVENT.md
```diff
@@ -119,10 +119,11 @@ are two types of attribution sources, `navigation` sources and `event` sources.
 or via calls to `window.open` that occur with [transient
 activation](https://html.spec.whatwg.org/multipage/interaction.html#transient-activation):
 ```javascript
+const encoded = encodeURIComponent('https://adtech.example/attribution_source?my_ad_id=123');
```
csharrison, line 122: A one-line comment explaining why we need to do this would be good. Something like:
```
// Encode the attributionsrc URL in case it contains "=" which will cause the parameter to be improperly parsed
```

csharrison (approved):
TIL about template literals in JS, thanks!

PR #410, opened 2022-05-04
Rename attribution report to event-level attribution report

linnan-github:
There's no change to the algorithm yet, just renaming.

***
<a href="https://pr-preview.s3.amazonaws.com/WICG/conversion-measurement-api/pull/410.html" title="Last updated on May 25, 2022, 3:49 AM UTC (10f04d0)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/conversion-measurement-api/410/f7e28e9...10f04d0.html" title="Last updated on May 25, 2022, 3:49 AM UTC (10f04d0)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -37,7 +37,7 @@ for cross-site identifiers like third-party cookies.
 
 Pages/embedded sites are given the ability to register [=attribution sources=] and
 [=attribution triggers=], which can be linked by the User Agent to generate and
-send [=attribution reports=] containing information from both of those events.
+send [=event-level reports=] containing information from both of those events.
 
 A reporter `https://reporter.example` embedded on `https://source.example` is able to
 measure whether an iteraction on the page lead to an action on `https://destination.example`
@@ -53,7 +53,7 @@ the User Agent attempts to match the trigger to previously registered source eve
 where the sources/triggers were registered and configurations provided by the reporter.
 
 If the User Agent is able to attribute the trigger to a source, it will generate and
-send an [=attribution report=] to the reporter via an HTTP POST request at a later point
+send an [=event-level report=] to the reporter via an HTTP POST request at a later point
 in time.
 
 # HTML monkeypatches # {#html-monkeypatches}
@@ -278,8 +278,8 @@ An attribution source is a [=struct=] with the following items:
 :: A 64-bit integer.
 : <dfn>source time</dfn>
 :: A point in time.
-: <dfn>number of reports</dfn>
-:: Number of [=attribution reports=] created for this [=attribution source=].
+: <dfn>number of event-level reports</dfn>
+:: Number of [=event-level reports=] created for this [=attribution source=].
 : <dfn>dedup keys</dfn>
 :: [=ordered set=] of [=event-level trigger configuration/dedup keys=] associated with this [=attribution source=].
 : <dfn>randomized response</dfn>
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -120,10 +120,10 @@ with the <{a}> element.
 
 The <dfn for="a" element-attr>attributionsourceeventid</dfn> attribute is a string
 containing information about the `attribution source` and will be supplied in the
-[=attribution report=].
+[=event-level attribution report=].
```
apasel422, line 123: While we're making this change, what do we think about removing `attribution` from the name, since we're already in the attribution reporting spec?

linnan-github: That makes sense to me, but we have `attribution` in many other structures as well...

johnivdel: We can, but now seems like a good time to start :)

johnivdel (commented):
Overall looks good % removing "attribution".

Do we have any ideas on how we are going to integrate aggregate reports here with respect to common fields? Are we going to need algorithms which operate on both event+aggregatable reports?

In any case this change as-is seems like a reasonable step in that direction.

index.bs
```diff
@@ -369,11 +369,11 @@ An attribution trigger is a [=struct=] with the following items:
 
 </dl>
 
-<h3 dfn-type=dfn>Attribution report</h3>
+<h3 dfn-type=dfn>Event-level report</h3>
 
-An attribution report is a [=struct=] with the following items:
+An evnet-level report is a [=struct=] with the following items:
```
johnivdel, line 374:
```suggestion
An event-level report is a [=struct=] with the following items:
```

index.bs
```diff
@@ -975,58 +975,58 @@ To <dfn>trigger attribution</dfn> given an [=attribution trigger=] |trigger| run
...
 <dfn>Max reports per attribution destination</dfn> is a vendor-specific integer which controls how
-many [=attribution reports=] can be in the [=attribution report cache=] for an
-[=attribution report/attribution destination=].
+many [=event-level reports=] can be in the [=event-level report cache=] for an
+[=event-level report/attribution destination=].
 
 <dfn>Max attributions per navigation source</dfn> is a vendor-specific integer that controls how
 many times a single [=attribution source=] whose [=attribution source/source type=] is
-"`navigation`" can be attributed.
+"`navigation`" can be attributed on event-level.
```
johnivdel, line 1022:
```suggestion
"`navigation`" can create an [=event-level report=].
```

index.bs
```diff
@@ -975,58 +975,58 @@ To <dfn>trigger attribution</dfn> given an [=attribution trigger=] |trigger| run
...
 
 <dfn>Max attributions per event source</dfn> is a vendor-specific integer that controls how
 many times a single [=attribution source=] whose [=attribution source/source type=] is "`event`"
-can be attributed.
+can be attributed on event-level.
```
johnivdel, line 1026:
```suggestion
can create an [=event-level report=].
```

johnivdel (approved):
lgtm with comments

johnivdel:
> Thanks John. Currently I'm thinking about duplicating the common fields for aggregatable reports for clarity (#423). But we can discuss more on that PR if there's a better way to spec.

Ack, will follow up on that PR, thanks!

PR #411, opened 2022-05-05
Impose a minimum delay of 10 minutes for aggregatable reports

linnan-github:
This may help with the regretful user user case so that there's a chance for the user to delete the attributions before they get sent.

### DIFF
--- a/AGGREGATE.md
+++ b/AGGREGATE.md
@@ -215,8 +215,9 @@ The report will be JSON encoded with the following scheme:
 
 Reports will not be delayed to the same extent as they are for event level
 reports. The browser will delay them with a random delay between 10 minutes to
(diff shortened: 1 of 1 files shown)


csharrison:
Can you add a sentence explaining why the minimum delay is added?

linnan-github:
> 

Added the description, thanks.

csharrison:
Sorry, I meant in the explainer itself. I think that would be useful too.

linnan-github:
> Sorry, I meant in the explainer itself. I think that would be useful too.

Ah sorry I misunderstood. Updated the explainer, thanks.

csharrison (approved):
LGTM, thanks!

PR #412, opened 2022-05-06
Specify limits on pending sources per source origin

apasel422:
https://github.com/WICG/conversion-measurement-api/blob/main/EVENT.md#storage-limits

#386

***
<a href="https://pr-preview.s3.amazonaws.com/apasel422/conversion-measurement-api/pull/412.html" title="Last updated on May 10, 2022, 7:08 PM UTC (6424877)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/conversion-measurement-api/412/f857acd...apasel422:6424877.html" title="Last updated on May 10, 2022, 7:08 PM UTC (6424877)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -679,7 +679,7 @@ To <dfn>process an attribution source</dfn> given an [=attribution source=] |sou
 1. Let |pendingSourcesForSourceOrigin| be the [=set=] of all
     [=attribution sources=] |pendingSource| of |cache| where |pendingSource|'s
     [=attribution source/source origin=] and |source|'s
-    [=attribution source/source origin=] are equal.
+    [=attribution source/source origin=] are [=same origin=].
 1. If |pendingSourcesForSourceOrigin|'s [=list/size=] is greater than or equal
     to the user agent's [=max pending sources per source origin=], return.
 1. Let |rateLimitRecord| be a new [=attribution rate-limit record=] with the items:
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -636,6 +636,12 @@ To <dfn>process an attribution source</dfn> given an [=attribution source=] |sou
 1. [=list/Remove=] all entries in |cache| where the entry's [=attribution source/expiry time=] is less than the current time.
 1. If the [=list/size=] of |cache| is greater than or equal to the user agent's
      [=max source cache size=], return.
+1. Let |pendingSourcesForSourceOrigin| be the [=set=] of all
+    [=attribution sources=] |pendingSource| of |cache| where |pendingSource|'s
```
csharrison, line 640: s/of |cache|/in |cache|/

index.bs
```diff
@@ -636,6 +636,12 @@ To <dfn>process an attribution source</dfn> given an [=attribution source=] |sou
...
+    [=attribution source/source origin=] and |source|'s
+    [=attribution source/source origin=] are equal.
```
csharrison, line 642: Should we say "same origin" here to be precise?
https://html.spec.whatwg.org/multipage/origin.html#same-origin

We already disallow opaque origins so it should be equivalent, and I don't think there's a standard "origin equality"

csharrison (approved):
LGTM with two suggestions

index.bs
apasel422: Per https://infra.spec.whatwg.org/#list-iterate, `of` is the standard term.

csharrison: Thank you for the reference! My mistake.

Issue #378, opened 2022-04-07
Consider making source-side attributionsrc pings easier to associate with existing pings


johnivdel:
Currently to declare a source in the API you must provide a separate url which can register attribution response headers:

```
<img src="https://adtech.example/existing_ping" attributionsrc="https://adtech.example/new_ping">
```

For the ad-tech, it may difficult to associate the new ping and existing ping unless certain query parameters are added. If there are multiple reporters within the attributionsrc redirect chain, this can become more complicated as there is a reliance on other parties setting these params.

This is related to ​​#347 where it was difficult to introduce a separate ping on the trigger side.

It's possible we can improve the developer experience here by having the browser make these pings strongly associated. A few (non-exhaustive) options:

1. Have the browser send a header on each of the requests to indicate that they are associated with each other.

```
<img src="https://adtech.example/existing_ping" attributionsrc="https://adtech.example/new_ping">
```

`https://adtech.example/existing_ping` receivers a header: `Attribution-Reporting-Eligible: false, <uuid>`

`https://adtech.example/new_ping` receivers a header: `Attribution-Reporting-Eligible: true, <uuid>`

`<uuid>` would be a random id generated by the browser to associate these two requests. The adtech is then able to associate these pings with each other server-side. Redirects for each of the requests would also have this ability.

2. Make the separate request optional. Instead, if the "attributionsrc" attribute is present without a value, it will allow registration on the existing src= request.

```
<img src="https://adtech.example/existing_ping" attributionsrc>
```

Here, `https://adtech.example/existing_ping` is able to register the Attribution-Reporting headers. We still retain an element level permissions model for whether a request is allowed to use the API.

In theory, a similar behavior could be applied to anchor tags, where the `href` itself could respond with headers. However, because these are potentially top-level navigations, these requests will behave very differently than attributionsrc requests. Namely, they would be in a different first party context.

It is possible (1) could be used in conjunction with (2) to provide additional flexibility. Sending a header itself on attributionsrcs may also make it easier for the browser to signal whether a request can actually register (e.g. permissions policy checks, browser support without UA sniffing).

(2) has the benefit that it uses less network bandwidth + avoids making two requests in some cases. 

Because (2) re-uses existing pings, the browser would have less flexibility in regards to performing+processing these pings. This may not be a problem in practice, but worth pointing out.

bmayd:
[...]

PR #428, opened 2022-05-16
Replace dedicated JS registration APIs with generic mechanism via fetch

### LINKED ISSUE #363: Is `window.attributionReporting.registerTrigger` necessary?
As far as I can tell, it doesn't provide any enhancements over `window.fetch`.

apasel422:
#378
Closes #363

### DIFF
--- a/EVENT.md
+++ b/EVENT.md
@@ -250,7 +250,11 @@ similar mechanism is used as source event registration, via HTML:
 ```
 or JavaScript:
 ```javascript
-window.fetch("https://adtech.example/attribution_trigger?purchase=13");
+const headers = {
+  'Attribution-Reporting-Eligible': 'true'
+};
+window.fetch("https://adtech.example/attribution_trigger?purchase=13",
+             { headers });
 ```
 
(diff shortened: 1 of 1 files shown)


EVENT.md
```diff
@@ -139,8 +139,11 @@ window.open(
 ```
 or via a JavaScript API:
 ```javascript
-window.attributionReporting.registerSource(
-  "https://adtech.example/attribution_source?my_ad_id=123");
+const headers = {
+  'Attribution-Reporting-Eligible': 'true'
+};
+window.fetch("https://adtech.example/attribution_source?my_ad_id=123",
+             { headers });
```
csharrison, line 146: Maybe it's just early, but why doesn't this need to be
```
window.fetch("...", {headers: headers})
```
I looked at the MDN docs and they also have this example, but I don't know why it works.

apasel422: This is a shorthand object literal (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Object_initializer#new_notations_in_ecmascript_2015) though we can certainly just inline it for clarity.

PR #437, opened 2022-05-23
Fix event explainer formatting

apasel422:


csharrison (approved):
LGTM. It feels like we might want to invest in some commit hooks to catch some of these things before landing, especially the quotes which I think can cause actual bugs if copy-pasted incorrectly.

PR #440, opened 2022-05-23
Remove erroneous data-limit comment from event explainer

apasel422:
Limits on trigger data are applied based on the type of the attributed source, which is discussed in the Data Limits section.

### DIFF
--- a/EVENT.md
+++ b/EVENT.md
@@ -273,7 +273,9 @@ about how to treat the trigger event:
 }]
 ```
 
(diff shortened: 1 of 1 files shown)


csharrison:
This is fine with me but it basically reverts https://github.com/WICG/conversion-measurement-api/pull/298 so we might still want to make a call out that this gets sanitized?

cc @maudnals

csharrison (approved):

PR #442, opened 2022-05-23
Replace boolean Attribution-Reporting-Eligible header with dictionary

apasel422:
#378

### DIFF
--- a/EVENT.md
+++ b/EVENT.md
@@ -26,6 +26,7 @@ extension on top of this.
   - [Publisher-side Controls for Attribution Source
     Declaration](#publisher-side-controls-for-attribution-source-declaration)
   - [Triggering Attribution](#triggering-attribution)
+  - [Registration requests](registration-requests)
   - [Data limits and noise](#data-limits-and-noise)
   - [Trigger attribution algorithm](#trigger-attribution-algorithm)
   - [Multiple sources for the same trigger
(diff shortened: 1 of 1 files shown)


EVENT.md
```diff
@@ -294,6 +296,69 @@ same-origin children, but disabled in cross-origin children.
 Navigation sources may be attributed up to 3 times. Event sources may be
 attributed up to 1 time.
 
+### Registration requests
```
csharrison, line 299: nit: this is a bit too much detail IMO. I think we can say that (paraphrasing):

1. `<a>` and `window.open` will have `navigation-source`
2. Other APIs which automatically set `Attribution-Reporting-Eligible` (like ``) will contain `event-source, trigger` 
3. Request from Javascript e.g. `fetch()` can set this header manually, but it is an error for such requests to specify `navigation-source`
4. All other requests will not have the `Attribution-Reporting-Eligible` header. For those requests we still allow trigger registration.

My main concern is that we don't try to be exhaustive, since I don't want to have to update this section if we add, e.g. `<script>` support. I also don't think we need JS snippets for all those examples.

csharrison (approved):
Generally looks good with a few suggestions

csharrison (approved):

johnivdel (approved):
LGTM w/ question

PR #448, opened 2022-05-23
Make event-trigger registration optional

apasel422:
This improves the utility of the API for reporting origins that are only interested in aggregatable reports, reducing unnecessary local storage and bandwidth.

### DIFF
--- a/AGGREGATE.md
+++ b/AGGREGATE.md
@@ -172,6 +172,10 @@ L1 = 1 << 16
 true_agg_campaign_counts = raw_agg_campaign_counts / (L1 / 2)
 true_agg_geo_value = 1024 * raw_agg_geo_value / (L1 / 2)
 ```
+
+Note that aggregatable trigger registration is independent of event-level
+trigger registration.
+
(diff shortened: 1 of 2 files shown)


johnivdel (approved):
LGTM

I think this is the first time we are referencing the aggregate explainer from within the content of the event explainer, but this adds a lot of clarity so seems fine to me.

csharrison:
> I think this is the first time we are referencing the aggregate explainer from within the content of the event explainer, but this adds a lot of clarity so seems fine to me.

Does it make more sense to include the documentation in the aggregate explainer? i.e. mention that the trigger for aggregatable contributions is independent of event registration.

apasel422:
> Does it make more sense to include the documentation in the aggregate explainer? i.e. mention that the trigger for aggregatable contributions is independent of event registration.

I'm open to this, but weirdly the aggregate explainer doesn't even mention right now that it only succeeds if the `Register-Event-Trigger` header is also present.

csharrison (approved):

Issue #446, opened 2022-05-23
Support view registration on more HTML elements


johnivdel:
The current explainer allows view source registration via JS, or by using the attributionsrc attribute on <img> elements. There are a variety of other HTML elements which also support making requests, such as <script>, which could presumably register sources as well.

This could make adoption easier as reporting origins won't need to transition to different request APIs, similar to the rationale given in https://github.com/WICG/conversion-measurement-api/issues/378#issuecomment-1122691996

csharrison:
I'm not opposed to supporting any of these surfaces if it's actually useful, but we should probably not add support until there is evidence of a need of some kind, otherwise we're just adding complexity to these other elements.

PR #460, opened 2022-05-27
Specify monkeypatch for <script> attributionsrc

apasel422:
#446


apasel422:
> Do we want to use a mixin here to prevent duplicating this for each element that we want attributionsrc on?
> 
> E.g. similar to https://html.spec.whatwg.org/#api-for-a-and-area-elements

I was not aware of this. Done.

index.bs
```diff
@@ -58,52 +59,36 @@ in time.
...
+HTMLScriptElement includes HTMLAttributionSrcElementUtils;
 </pre>
 
-The IDL attribute {{HTMLImageElement/attributionSrc}} must
-<a spec=html>reflect</a> the respective content attribute of the same name.
-
-The <dfn for="img" element-attr>attributionsrc</dfn> attribute is a string
-representing the [=URL=] of the resource that will register an
-[=attribution source=] or [=attribution trigger=] when set.
+Add the following <a spec=html>content attributes</a>:
```
johnivdel, line 74: Might make sense to align this a bit more with: 

https://wicg.github.io/priority-hints/#img

Add to the <{a}>, <{img}>, <{script}> <a spec=html>content attributes</a>:

attributionsrc - [=URL=] for attribution registration

Add to the <{a}> content attribute descriptions:
...

Add to the...

johnivdel (approved):
lgtm with suggestion on for adding both the attributes + descriptions in the monkeypatch

index.bs
apasel422: Could you clarify the exact text you want? IMO the way I have it now seems concise and complete, but I might be missing something.

johnivdel: So the list structure is fine. I was trying to point out that this monkeypatch only adds content attribute descriptions not the attributes themselves.

So maybe we could modify this to be:
```
Add the following <a spec=html>content attributes</a>:

: <{a}>
:: `attributionsrc` - [=URL=] for attribution registration
: <{img}>
:: `attributionsrc` - [=URL=] for attribution registration
: <{script}>
:: `attributionsrc` - [=URL=] for attribution registration

Add the following content attribute descriptions:
: <{a}>
:: The <dfn for="a" element-attr>attributionsrc</dfn> attribute is a [=string=]
    representing the [=URL=] of the resource that will register an
    [=attribution source=] when the <{a}> is navigated.
: <{img}>
:: The <dfn for="img" element-attr>attributionsrc</dfn> attribute is a
    [=string=] representing the [=URL=] of the resource that will register an
    [=attribution source=] or [=attribution trigger=] when set.
: <{script}>
:: The <dfn for="script" element-attr>attributionsrc</dfn> attribute is a
    [=string=] representing the [=URL=] of the resource that will register an
    [=attribution source=] or [=attribution trigger=] when set.

The IDL attribute {{HTMLAttributionSrcElementUtils/attributionSrc}}
must <a spec=html>reflect</a> the respective content attribute of the same
name.
```

johnivdel (approved):
slgtm

PR #468, opened 2022-06-01
Specify filter-data parsing

apasel422:
***
<a href="https://pr-preview.s3.amazonaws.com/apasel422/conversion-measurement-api/pull/468.html" title="Last updated on Jun 6, 2022, 7:05 PM UTC (85495ac)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/attribution-reporting-api/468/40bfa3a...apasel422:85495ac.html" title="Last updated on Jun 6, 2022, 7:05 PM UTC (85495ac)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -443,6 +443,12 @@ maximum value that can be used as an [=attribution source/event id=].
 maximum value that can be used as an [=attribution source/expiry=]. It must be
 greater than or equal to 30 days.
 
+<dfn>Max entries per attribution filter map</dfn> is a positive integer that
+controls the maximum [=map/size=] of an [=attribution filter map=].
+
+<dfn>Max values per attribution filter entry</dfn> is a positive integer that
+controls the maximum [=set/size=] of an [=attribution filter map=] entry.
+
 <dfn>Max pending sources per source origin</dfn> is a positive integer that
 controls how many [=attribution sources=] can be in the
 [=attribution source cache=] per [=attribution source/source origin=].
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -516,6 +522,31 @@ controls the maximum number of attributions for a
 
 # General Algorithms # {#general-algorithms}
 
+<h3 id="parsing-filter-data">Parsing filter data</h3>
+
+To <dfn>parse filter data</dfn> given a |value| and an optional [=boolean=]
```
csharrison, line 527: Optional nit: a boolean is slightly less clear than an enum, where the value says what is for.

index.bs
```diff
@@ -516,6 +522,31 @@ controls the maximum number of attributions for a
...
+1. If |value| is not a [=map=], return null.
+1. If |forSource| is true and |value|["`source_type`"] [=map/exists=], return
+    null.
+1. If |value|'s [=map/size=] is greater than the user agent's
+    [=max entries per attribution filter map=], return null.
+1. Let |result| be a new [=attribution filter map=].
+1. [=map/iterate|For each=] |filter| → |data| of |value|:
+    1. If |data| is not a [=list=], return null.
+    1. Let |set| be a new [=ordered set=].
+    1. [=list/iterate|For each=] |d| of |data|:
```
johnivdel, line 539: Do we want to use a more specific name here like filterValueItem? I'm fine with the short form d as well though.

index.bs
```diff
@@ -643,6 +680,7 @@ To <dfn noexport>parse source-registration JSON</dfn> given a [=string=]
         [=max attributions per event source=].
 1. Let |numReportWindows| be the result of running
     [=obtain the number of report windows=] with |sourceType|.
+1. [=map/Set=] |filterData|["`source_type`"] to « |sourceType| ».
```
csharrison, line 683: Why not have this above right after invoking `parse filter data`?

index.bs
```diff
@@ -839,6 +883,11 @@ To <dfn noexport>parse trigger-registration JSON</dfn> given a [=string=]
         <a spec="html">rules for parsing non-negative integers</a> to
         |value|["`debug_key`"].
     1. If |debugKey| is an error, set |debugKey| to null.
+1. Let |filters| be a new [=attribution filter map=].
```
johnivdel, line 886: Unrelated:

Now that the headers are collapsed should we support top level negation on the trigger side? Seems like there is no reason not to.

johnivdel (approved):
LGTM thanks!

csharrison (approved):

index.bs
apasel422: It has a short scope, so I think it's fine as is.

index.bs
apasel422: We could easily do so, but let's discuss in a separate bug.

Issue #365, opened 2022-03-15
Consider replacing privacy_budget_key hash with components


alexmturner:
Instead of having the client construct a hash for the privacy_budget_key, the inputs to the hash could be provided in the shared_info field. This would allow the aggregation service to construct a hash itself, providing some additional flexibility. It would also ensure that no information is (implicitly) duplicated, preventing complicated edge cases caused by bad (malicious or broken) clients. For example, a bad client could have an inconsistency in the reporting_origin in the privacy_budget_key hash input and in its dedicated field.

We could also consider grouping these fields into a JSON sub-object (e.g. `privacy_budget_components`)

csharrison:
I think the simplest thing to do is to just embed everything into `shared_info`:
```
{
  "attribution_destination": "https://advertiser.example",
  "report_id":"[UUID]",
  "reporting_origin":"https://reporter.example"
  "scheduled_report_time":"[timestamp in seconds]",
  "source_registration_time": "[timestamp in seconds]",
  "source_site": "https://publisher.example",
  "version":"[api version]",
}
```

PR #471, opened 2022-06-02
Remove privacy_budget_key

csharrison:
Fixes https://github.com/WICG/attribution-reporting-api/issues/365 and simplifies the aggregatable report.

- sorts the keys in ascending lexicographic order
- moves the attribution_destination and source_registration_time keys into shared_info


johnivdel (commented):
I think you are missing the removal of a refernce to this in https://github.com/WICG/attribution-reporting-api/blob/main/AGGREGATE.md#data-processing-through-a-secure-aggregation-service

csharrison:
Thanks, done.

PR #474, opened 2022-06-03
Specify distinct destination limit covered  by pending sources

linnan-github:


index.bs
```diff
@@ -676,6 +681,27 @@ Issue: Parse |value|["`filter_data`"].
 
 Issue: Parse |value|["`aggregation_keys`"].
 
+<h3 dfn id="should-block-processing-for-destination-limit">Should processing be blocked by destination limit</h3>
+
+Given an [=attribution source=] |source|:
```
apasel422, line 686: I think this should be written as
```
To <dfn>...</dfn> given...
```

index.bs
```diff
@@ -477,6 +477,11 @@ controls how many times a single [=attribution source=] whose
 [=attribution source/source type=] is "`event`" can create an
 [=event-level report=].
 
+<dfn>Max destinations covered by pending sources</dfn> is a positive
+integer that controls the maximum number of distinct [=attribution source/attribution destinations=]
+for a ([=attribution source/source site=], [=attribution source/reporting endpoint=]) covered by
+pending [=attribution sources=].
```
apasel422, line 483:
```suggestion
for pending [=attribution sources=] with a given ([=attribution source/source site=], [=attribution source/reporting endpoint=]).
```

index.bs
```diff
@@ -676,6 +681,27 @@ Issue: Parse |value|["`filter_data`"].
...
 Issue: Parse |value|["`aggregation_keys`"].
 
+<h3 dfn id="should-block-processing-for-destination-limit">Should processing be blocked by destination limit</h3>
+
+Given an [=attribution source=] |source|:
+1. Let |pendingSources| be all entries in the [=attribution source cache=] where all of the following are true:
+     * entry's [=attribution source/source site=] and |source|'s [=attribution source/source site=]
+        are [=same site=].
+     * entry's [=attribution source/reporting endpoint=] and |source|'s
+        [=attribution source/reporting endpoint=] are [=same origin=].
```
apasel422, line 691: Elsewhere in the spec, we compare this field by saying "equal" rather than "same origin". If those uses are wrong, could we update them first? If not, let's be consistent and say "equal".

index.bs
```diff
@@ -676,6 +681,27 @@ Issue: Parse |value|["`filter_data`"].
...
+     * entry's [=attribution source/number of event-level reports=] is equal to 0.
+1. Let |distinctDestinations| be a new [=set/empty=] [=set=].
```
apasel422, line 693:
```suggestion
1. Let |distinctDestinations| be a new [=set/is empty|empty=] [=set=].
```

Otherwise, this links to the method for clearing a set.

apasel422 (changes requested):

index.bs
linnan-github: Thanks Andrew. Yeah, I agree, but I can't think of a good way to describe like this... Any suggestions?

apasel422:
```
To <dfn>check if an attribution source meets the pending destination limit</dfn>...
```

Something like that?

index.bs
```diff
@@ -676,6 +680,29 @@ Issue: Parse |value|["`filter_data`"].
...
+     * entry's [=attribution source/source site=] and |source|'s [=attribution source/source site=]
+        are equal.
+     * entry's [=attribution source/reporting endpoint=] and |source|'s
+        [=attribution source/reporting endpoint=] are [=same origin=].
+     * entry's [=attribution source/number of event-level reports=] is equal to 0.
+1. Let |distinctDestinations| be a new [=set/is empty|empty=] [=set=].
+1. For each [=attribution source=] |pendingSource| of |pendingSources|:
+    1. If |source|'s [=attribution source/attribution destination=] and |pendingSource|'s
+        [=attribution source/attribution destination=] are equal, then return true.
+    1. If |distinctDestinations| [=set/contains=] |pendingSource|'s
```
johnivdel, line 698: I think this adds unnecessary work, set/append should implicitly avoid the double counting.

index.bs
```diff
@@ -676,6 +680,29 @@ Issue: Parse |value|["`filter_data`"].
...
+        [=attribution source/attribution destination=], [=iteration/continue=].
+    1. [=set/Append=] |pendingSource|'s [=attribution source/attribution destination=] to
+        |distinctDestinations|.
+1. If |distinctDestinations|'s [=set/size=] is greater than or equal to
+    the user agent's [=max destinations covered by pending sources=], then return false.
+1. Return true.
+
```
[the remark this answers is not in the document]
linnan-github: SGTM, updated, thanks!

index.bs
```diff
@@ -676,6 +680,29 @@ Issue: Parse |value|["`filter_data`"].
...
 
+<h3 id="check-pending-destination-limit">Checking pending destination limit</h3>
+
+To <dfn>check if an [=attribution source=] meets the pending destination limit</dfn> given an
+[=attribution source=] |source|, run the following steps:
+
+1. Let |pendingSources| be all entries in the [=attribution source cache=] where all of the following are true:
+     * entry's [=attribution source/source site=] and |source|'s [=attribution source/source site=]
+        are equal.
+     * entry's [=attribution source/reporting endpoint=] and |source|'s
```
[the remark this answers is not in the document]
linnan-github: As discussed offline, same origin check is fine.

PR #475, opened 2022-06-03
Use same origin instead of equals 

linnan-github:
***
<a href="https://pr-preview.s3.amazonaws.com/WICG/attribution-reporting-api/pull/475.html" title="Last updated on Jun 3, 2022, 2:46 PM UTC (fc98986)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/attribution-reporting-api/475/6d3b06c...fc98986.html" title="Last updated on Jun 3, 2022, 2:46 PM UTC (fc98986)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -101,7 +101,7 @@ name.
 <h3 id="monkeypatch-navigation">Navigation</h3>
 
 This section ensures that an [=attribution source=] associated with a navigation
-results in a top-level navigation whose final URL is [=same site=] with the
+results in a top-level navigation whose final URL equals the
 [=attribution source/attribution destination=].
 
 <h4 id="monkeypatch-navigation-params">Navigation Params</h4>
@@ -683,7 +683,7 @@ To <dfn>maybe process a navigation attribution source</dfn> given a <a spec="HTM
 1. If |browsingContext| is not a <a spec="html">top-level browsing context</a>, return.
 1. Let <var>attributionSource</var> be |navigationParams|'s [=navigation params/attribution source=].
 1. If |attributionSource| is null, return.
-1. If |attributionSource|'s [=attribution source/attribution destination=] is not [=same site=] to |navigationParams|'s
+1. If |attributionSource|'s [=attribution source/attribution destination=] does not equal |navigationParams|'s
     <a href="https://html.spec.whatwg.org/multipage/browsing-the-web.html#navigation-params-origin">origin</a>, return.
 1. [=Queue a task=] to [=process an attribution source=] with |attributionSource|.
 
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -723,7 +723,7 @@ To <dfn>process an attribution source</dfn> given an [=attribution source=] |sou
 
 1. Let |cache| be the user agent's [=attribution source cache=].
 1. [=list/Remove=] all entries in |cache| where all of the following are true:
-    * the entry's [=attribution source/attribution destination=] and |source|'s [=attribution source/attribution destination=] are equal.
+    * the entry's [=attribution source/attribution destination=] is [=same site=] with |source|'s [attribution source/attribution destination].
```
apasel422, line 726: From the same site link:

> Two origins, A and B, are said to be same site...

But `attribution destination` is a site, so I'm not sure the operation is defined here, and `equals` may actually be correct.

linnan-github: Yeah, that's right. Currently there are two places we use `same site` for a site, will replace them to `equals` as well.

index.bs
```diff
@@ -101,7 +101,7 @@ name.
 <h3 id="monkeypatch-navigation">Navigation</h3>
 
 This section ensures that an [=attribution source=] associated with a navigation
-results in a top-level navigation whose final URL is [=same site=] with the
+results in a top-level navigation whose final URL equals the
```
apasel422, line 104: I don't think this is exactly right. The URL itself might not equal the destination.

index.bs
```diff
@@ -683,7 +683,7 @@ To <dfn>maybe process a navigation attribution source</dfn> given a <a spec="HTM
 1. If |browsingContext| is not a <a spec="html">top-level browsing context</a>, return.
 1. Let <var>attributionSource</var> be |navigationParams|'s [=navigation params/attribution source=].
 1. If |attributionSource| is null, return.
-1. If |attributionSource|'s [=attribution source/attribution destination=] is not [=same site=] to |navigationParams|'s
+1. If |attributionSource|'s [=attribution source/attribution destination=] does not equal |navigationParams|'s
```
apasel422, line 686: Not sure this is right, either. The attribution destination is a site, but it's being compared to an origin here.

apasel422 (changes requested):

index.bs
linnan-github: Yeah, probably better to get site from origin first and then compare the site. Will do this in a separate PR.

apasel422:
LGTM, but please update the PR title

apasel422 (approved):

PR #477, opened 2022-06-06
Add top-level negated filters

apasel422:
***
<a href="https://pr-preview.s3.amazonaws.com/apasel422/attribution-reporting-api/pull/477.html" title="Last updated on Jun 16, 2022, 1:57 PM UTC (1caed10)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/attribution-reporting-api/477/0a1cbfc...apasel422:1caed10.html" title="Last updated on Jun 16, 2022, 1:57 PM UTC (1caed10)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -353,6 +353,8 @@ An attribution trigger is a [=struct=] with the following items:
 :: An [=url/origin=].
 : <dfn>filters</dfn>
 :: A [=filter map=].
+: <dfn>negated filters</dfn>
+:: A [=filter map=].
 : <dfn>debug key</dfn>
 :: Null or a non-negative 64-bit integer.
 : <dfn>event-level trigger configurations</dfn>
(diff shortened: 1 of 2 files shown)


index.bs
```diff
@@ -928,6 +930,11 @@ To <dfn noexport>parse trigger-registration JSON</dfn> given a [=string=]
     1. Set |filters| to the result of running [=parse filter data=] with
         |value|["`filters`"] and "`trigger`".
     1. If |filters| is null, return null.
+1. Let |negatedFilters| be a new [=filter map=].
+1. If |value|["`not_filters`"] exists:
+    1. Set |negatedFilters| to the result of running [=parse filter data=] with
```
johnivdel, line 935: Minor nit: We could consider using non-optiona named parameters for parse filter data. It's not clear when reading the spec what `trigger` is. 

https://infra.spec.whatwg.org/#algorithm-params

"Optional named parameters, instead of positional ones, can be used to increase clarity and flexibility at the call site. Such parameters are marked up as both variables and definitions, and linked to from their call sites."

johnivdel (approved):

index.bs
apasel422: Ack. Will do that in a followup for consistency with the other parsing.

PR #480, opened 2022-06-07
Remove truncation wording for aggregation keys

linnan-github:
The current description is a bit confusing and the implementation just errors if the specified key piece is longer than the limit. It seems to be more clear as the truncation can be in either direction.

### DIFF
--- a/AGGREGATE.md
+++ b/AGGREGATE.md
@@ -92,7 +92,7 @@ fully defined at trigger time using a combination (binary OR) of this piece and
 trigger-side pieces.
 
 Final keys will be restricted to a maximum of 128 bits. This means that hex
-strings in the JSON should be limited to at most 32 digits.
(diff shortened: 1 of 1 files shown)


AGGREGATE.md
```diff
@@ -91,9 +91,8 @@ aggregation key defined as a hex-string. The final histogram bucket key will be
 fully defined at trigger time using a combination (binary OR) of this piece and
 trigger-side pieces.
 
-Final keys will be restricted to a maximum of 128 bits. Keys longer than this
-will be truncated. This means that hex strings in the JSON should be limited to
-at most 32 digits.
+Final keys will be restricted to a maximum of 128 bits. This means that hex
+strings in the JSON should be limited to at most 32 digits.
```
csharrison, line 95: s/should/must

csharrison (approved):
LGTM with one comment

PR #483, opened 2022-06-08
Add "api" string enum field to shared info

alexmturner:
Allows for the aggregation service to support multiple APIs in the future

### DIFF
--- a/AGGREGATE.md
+++ b/AGGREGATE.md
@@ -193,7 +193,7 @@ The report will be JSON encoded with the following scheme:
 {
   // Info that the aggregation services also need encoded in JSON
   // for use with AEAD.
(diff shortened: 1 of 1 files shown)


csharrison:
What do you think about shortening the API string to save bytes on disk?

alexmturner:
We could try something like "attribution" instead. That being said, I'm not sure how valuable shortening is given the length of some of our other keys/values. So I'd have a slight preference to leaving it as is, but happy either way.

csharrison (approved):
I think this is fine.

johnivdel:
I do like that `attribution-reporting` matches the permission policy name. But even something like `ara` (to distinguish from augmented reality) seems very reasonable imo. This will also save bandwidth for the browser and the developer when uploading batches.

My expectation is the main consumer of this field will be the aggregation service, rather than the developer who is getting reports on an api specific endpoint. Going with the shortened version seems like it is serving everyones interests. The aggregation service is complicated enough that a more obfuscated name should be manageable.

alexmturner:
I think let's stick with "attribution-reporting" then; this better aligns with the "common words" and "consistency" naming principles (https://www.w3.org/TR/design-principles/#naming-is-hard)

PR #484, opened 2022-06-08
Shorten aggregation key id example

linnan-github:
The existing key id is actually longer than the limit and will fail.

***
<a href="https://pr-preview.s3.amazonaws.com/WICG/attribution-reporting-api/pull/484.html" title="Last updated on Nov 30, 2022, 8:59 PM UTC (d2d1a80)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/attribution-reporting-api/484/8020de4...d2d1a80.html" title="Last updated on Nov 30, 2022, 8:59 PM UTC (d2d1a80)">Diff</a>

### DIFF
--- a/AGGREGATE.md
+++ b/AGGREGATE.md
@@ -129,7 +129,7 @@ header](https://github.com/WICG/attribution-reporting-api/blob/main/EVENT.md#tri
       // which will take up 7 bits of space in the resulting key.
       "key_piece": "0xA80",
       // Apply this key piece to:
-      "source_keys": ["geoValue", "nonMatchingKeyIdsAreIgnored"]
+      "source_keys": ["geoValue", "nonMatchingKeyIdsIgnored"]
     }
   ],
   "aggregatable_values": {


johnivdel:
Would it be possible to document limits either in the explainer/spec as part of this change?

Ideally we could put it in the spec, although I am not sure if all of the parsing is in place for that. We could also just add placeholder constants in the spec with a TODO to use them

linnan-github:
> Would it be possible to document limits either in the explainer/spec as part of this change?
> 
> Ideally we could put it in the spec, although I am not sure if all of the parsing is in place for that. We could also just add placeholder constants in the spec with a TODO to use them

Thanks John. Yeah, that makes sense. Added a placeholder for the constant in the spec. PTAL.

linnan-github:
@johnivdel @apasel422 could you PTAL, thanks!

apasel422:
This PR has been open for over a year. What's the status?

apasel422 (approved):

PR #492, opened 2022-06-27
Add script validating header JSON

apasel422:


validate-json.mjs
```diff
@@ -0,0 +1,225 @@
...
+    return {errors: this.errors, warnings: this.warnings};
+  }
+
+  validate(obj, checks) {
+    Object.entries(checks).forEach(([key, check]) =>
+        this.scope(key, () => check(this, obj, key)));
+
+    Object.keys(obj).forEach(key => {
+      if (!(key in checks)) {
+        this.scope(key, () => this.warn('is an unknown field'));
```
maudnals, line 36: Suggestion:
"error type: field" (e.g. "unknown field: xxx") as suggested above in `formattedError` may be easier to parse and quicker to act upon than "field error type" ("xxx is an unknown field"). 
It also makes it potentially easier later to group errors per field, which maps better to the debugging flow (fix fieldX, then fix fieldY).

```suggestion
        this.scope(key, () => this.warn('unknown field'));
```

validate-json.mjs
```diff
@@ -0,0 +1,225 @@
...
+  }
+};
+
+function required(f = () => {}) {
+  return (state, object, key) => {
+    if (key in object) {
+      f(state, object[key]);
+      return;
+    }
+    state.error('must be present');
```
maudnals, line 48: Suggestion:
Same reasoning as above
```suggestion
    state.error('missing required field');
```

validate-json.mjs
```diff
@@ -0,0 +1,225 @@
...
+          state.scope(index, () => f(state, value)));
+      return;
+    }
+    state.error('must be a list');
+  }
+}
+
+const uint64 = string((state, value) => {
+  if (!uint64Regex.test(value)) {
+    state.error(`must match ${uint64Regex}`);
```
maudnals, line 95: Nit:
Same as above
```suggestion
    state.error(`must be a uint64 (must match ${uint64Regex})`);
```

validate-json.mjs
```diff
@@ -0,0 +1,225 @@
...
+  const max = 2n ** (64n - 1n) - 1n;
+  const min = (-2n) ** (64n - 1n);
+  if (val < min || val > max) {
+    state.error('must fit in a signed 64-bit integer');
+  }
+});
+
+const hex128 = string((state, value) => {
+  if (!hex128Regex.test(value)) {
+    return state.error(`must match ${hex128Regex}`);
```
maudnals, line 122: Nit: 
Same as above:
```suggestion
    return state.error(`must be a hex128 (must match ${hex128Regex})`);
```

maudnals (changes requested):
LGTM % comments!

maudnals (approved):
LGTM!

PR #499, opened 2022-06-30
Check url.search instead of url.searchParams

### LINKED ISSUE #497: Bug: error generated on valid destination
To reproduce: 
* Run `node ./index.js`
* Expected behavior:
```javascript
{ source_event_id: '12340873456', destination: 'https://example.com' }
```
should be a valid header.
It should log `✅ 0 ERRORS, 0 WARNINGS` 
* Observed behavior:
```
⚠️  1 WARNING(S):
contains a query string that will be ignored: ["destination"]
```

apasel422:
The former is a string, the latter an object.

Fixes #497

### DIFF
--- a/validate-json.mjs
+++ b/validate-json.mjs
@@ -141,7 +141,7 @@ const destination = string((state, url) => {
     state.warn('contains a path that will be ignored');
   }
 
-  if (url.searchParams !== '') {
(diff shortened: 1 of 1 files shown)


maudnals (approved):
LGTM!

PR #507, opened 2022-07-01
Made error messages more precise

### LINKED ISSUE #498: DevX: make error messages more precise to ease debugging
* When an unknown field value is used, explicitly display what are the allowed field names for that specific header (e.g. `Field must be one of...`) to make it easier to find the mistake/confusion
*  When destination contains a path/fragment, log that as well to make it easier to find the culprit
*  Tweak formatting to improve display of values within array and of nested values

(Maybe: use `console.table` to display an error table with col1 = field, col2 = error message. Same for warnings.)

maudnals:
Fixes #498

### DIFF
--- a/header-validator/validate-json.js
+++ b/header-validator/validate-json.js
@@ -287,7 +287,7 @@ export function formatIssue({ msg, path }) {
     context = path
       .map((p, idx) =>
         // For non-nested individual values, this displays field x
-        // For nested individual values, this displays fieldx > field y
+        // For nested individual values, this displays field x > field y
         // For values within arrays, this displays array[idx]
         typeof p === 'number' ? `${path[idx - 1]}[${p}]` : `${p}`
       )
--- a/header-validator/validate-json.js
+++ b/header-validator/validate-json.js
@@ -2,6 +2,39 @@ const uint64Regex = /^[0-9]+$/
 const int64Regex = /^-?[0-9]+$/
 const hex128Regex = /^0[xX][0-9A-Fa-f]{1,32}$/
 
+const sourceFields = Object.freeze({
+  AGGREGATION_KEYS: 'aggregation_keys',
+  DEBUG_KEY: 'debug_key',
+  DESTINATION: 'destination',
+  FILTER_DATA: 'filter_data',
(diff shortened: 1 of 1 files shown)


maudnals:
image

header-validator/validate-json.js
```diff
@@ -176,22 +213,24 @@ const aggregationKeys = object((state, key, value) => {
...
-    destination: required(destination),
-    filter_data: optional(filters(/*allowSourceType=*/ false)),
-    priority: optional(int64),
-    source_event_id: required(uint64),
+    [sourceFields.AGGREGATION_KEYS]: optional(aggregationKeys),
+    [sourceFields.DEBUG_KEY]: optional(uint64),
+    [sourceFields.DESTINATION]: required(destination),
+    [sourceFields.FILTER_DATA]: optional(filters(/*allowSourceType=*/ false)),
+    [sourceFields.PRIORITY]: optional(int64),
+    [sourceFields.SOURCE_EVENT_ID]: required(uint64),
```
apasel422, line 221: Instead of referencing the `sourceFields` object keys like this, can we instead lift up this entire dictionary so it's:

```js
const sourceFields = Object.freeze({
  'aggregation_keys': optional(aggregationKeys),
  'debug_key': optional(uint64),
  ...
})
```

header-validator/validate-json.js
```diff
@@ -246,9 +285,14 @@ export function formatIssue({ msg, path }) {
     context = 'JSON root'
   } else {
     context = path
-      .map((p) => (typeof p === 'number' ? `[${p}]` : `["${p}"]`))
-      .join('')
+      .map((p, idx) =>
+        // For non-nested individual values, this displays field x
+        // For nested individual values, this displays fieldx > field y
```
apasel422, line 290:
```suggestion
        // For nested individual values, this displays field x > field y
```

apasel422 (changes requested):

header-validator/validate-json.js
apasel422: Using regular functions sounds good to me.

PR #513, opened 2022-07-06
Handle non-object values gracefully

### LINKED ISSUE #509: Bug in event trigger data validation
Validating `  {
    event_trigger_data: ['2343']
  },` (which would be a header config mistake) fails throws as follows:

```
if (key in object) {
            ^
TypeError: Cannot use 'in' operator to search for 'deduplication_key' in 2343
```

In the fix, would it be also possible to point the dev to their mistake?

apasel422:
Fixes #509


maudnals (approved):
LGTM!

Issue #482, opened 2022-06-08
Source event limits should be time bounded


csharrison:
In https://github.com/WICG/attribution-reporting-api/blob/main/EVENT.md#limiting-the-number-of-unique-destinations-covered-by-pending-sources, we propose a limit on the number of sources that can be registered as "pending" at one time;

> Strawman: 100 distinct destination sites per-{source site, reporting origin}, applied to all pending sources regardless of type.

We should consider changing this to:

> Strawman: 100 distinct destination sites per-{source site, reporting origin, time window}, applied to all sources regardless of type

Benefits:
- Enforcement is only applied via information from a single site, so things like error reporting based on this limit no longer becomes privacy sensitive
- The limit is no longer susceptible to a "fast converting" attack pattern where an attacker who can convert very fast can bypass this limit.

johnivdel:
Another idea for discussion would be adjusting the limit while keeping the concept of limiting "simultaneous" sources:

> 100 distinct destination sites per-{source site, reporting origin}, applied at source time by checking whether there were 100 source registrations whose [source time, source time expiry] intersects with the new source.

This would require the browser to have knowledge of all source registations (including that those that have been attributed already).

This allows reporting origins to tradeoff between expiry window, and the number of destinations they may register in some larger time frame.

PR #514, opened 2022-07-07
Introduce optional error reports

### LINKED ISSUE #160: Consider sending aggregate error reports for event-level reports
There are a number of failure modes which can be difficult to observe due to the delayed reporting mechanism, and opaque way the browser handles sources and triggers for event level reports.

Some examples are:

- a source declares the wrong `attributiondestination`, causing the browser to drop the source
- a given user hits the [report rate limit](https://github.com/WICG/conversion-measurement-api#triggering-attribution-multiple-times-for-the-same-source), causing all subsequent reports to not be sent in the time period
- a source is removed due to the [pending destination limit](https://github.com/WICG/conversion-measurement-api/blob/main/event_attribution_reporting.md#limiting-the-number-of-unique-sites-covered-by-pending-sources)

In these cases, there is no straight-forward way for a developer to monitor this behavior outside of inferring they aren't receiving as many reports as they would normally.

It's possible the API could proactively send reports for these types of errors in parallel. To prevent cross-site information from being revealed the browser could send very little information in the error reports, for example just revealing {reporting_origin,error_code}.

It could also be possible to create and send the error reports similar to [aggregate reports](https://github.com/WICG/conversion-measurement-api/blob/main/AGGREGATE.md).

linnan-github:
Fixes #160. 

Currently we only support one failure mode, as other failure modes may have privacy/security implications and need more thoughts/mitigations.

This proposal has a hard dependency on #482.

### DIFF
--- a/EVENT.md
+++ b/EVENT.md
@@ -630,7 +630,7 @@ included in the request body as a JSON object:
 }
 ```
 
-These error reports will be sent immediately upon the error occuring during
+These error reports will be sent immediately upon the error occurring during
 source registration.
 
 Note that if other failures are reported using this framework, the browser may


EVENT.md
```diff
@@ -603,6 +604,66 @@ Set-Cookie: ar_debug=1; SameSite=None; Secure; HttpOnly
 ```
 If a cookie of this form is not present, debugging information will be ignored.
 
+### Optional: error reports
+
+We also introduce a non-cookie based debugging framework to allow developers to
```
apasel422, line 609:
```suggestion
We also introduce a non-cookie-based debugging framework to allow developers to
```

EVENT.md
```diff
@@ -603,6 +604,66 @@ Set-Cookie: ar_debug=1; SameSite=None; Secure; HttpOnly
...
+
+The browser will send error reports in the following source registration failure
+modes:
+
+* a source is rejected due to the [reporting origin
+  limits](#reporting-origin-limits)
+* a source is rejected due to the [destination
+  limits](#limiting-the-number-of-unique-destinations-covered-by-pending-sources)
+
+The browser will send non-credentialed secure HTTP POST requests to the
```
apasel422, line 620:
```suggestion
The browser will send non-credentialed secure HTTP `POST` requests to the
```

EVENT.md
```diff
@@ -603,6 +604,66 @@ Set-Cookie: ar_debug=1; SameSite=None; Secure; HttpOnly
...
+reporting endpoints, see [below](#reporting-endpoints). The report data is
+included in the request body as a JSON object:
+```jsonc
+{
+  "type": "<report type>", // could be "unattributed-reporting-origin-limit" or
+  "source-destination-limit"
+  "body": {
+    "limit": 100, // a constant integer
```
apasel422, line 628: `constant` is a bit ambiguous here. Maybe just: `// the browser's limit`?

EVENT.md
```diff
@@ -603,6 +604,66 @@ Set-Cookie: ar_debug=1; SameSite=None; Secure; HttpOnly
...
+  }
+}
+```
+
+These error reports will be sent immediately upon the error occuring during
+source registration.
+
+Note that if other failures are reported using this framework, the browser may
+enforce a random small delay and drop some data (e.g. `source_event_id`) to
+avoid the page being identifiable.
```
apasel422, line 642:
```suggestion
preserve privacy.
```

EVENT.md
```diff
@@ -603,6 +604,66 @@ Set-Cookie: ar_debug=1; SameSite=None; Secure; HttpOnly
...
+
+#### Reporting endpoints
+
+The reporting origins may opt in to receive error reports by adding a new
+`error_reporting` dictionary field to the `Attribution-Reporting-Register-Source` header:
+```jsonc
+{
+  ... // exsiting fields
```
apasel422, line 650:
```suggestion
  ... // existing fields
```

EVENT.md
```diff
@@ -603,6 +604,66 @@ Set-Cookie: ar_debug=1; SameSite=None; Secure; HttpOnly
...
+
+  "error_reporting": true // default to false if not present
```
apasel422, line 652:
```suggestion
  "error_reporting": true // defaults to false if not present
```

EVENT.md
```diff
@@ -603,6 +604,66 @@ Set-Cookie: ar_debug=1; SameSite=None; Secure; HttpOnly
...
+}
+```
+
+The error reports will be sent to a new endpoint:
+```
+https://<reporting origin>/.well-known/attribution-reporting/error
+```
+
+The top-level sites may opt in to receive error reports by declaring a reporting
```
apasel422, line 661:
```suggestion
The top-level sites may opt in to receiving error reports by declaring a reporting
```

EVENT.md
```diff
@@ -603,6 +604,66 @@ Set-Cookie: ar_debug=1; SameSite=None; Secure; HttpOnly
...
+endpoint via a HTTP header in the top-level document response:
```
apasel422, line 662:
```suggestion
endpoint via an HTTP header in the top-level document response:
```

apasel422 (dismissed):

EVENT.md
```diff
@@ -603,6 +604,66 @@ Set-Cookie: ar_debug=1; SameSite=None; Secure; HttpOnly
...
+```
+Attribution-Error-Reporting-Endpoint: <reporting endpoint>
```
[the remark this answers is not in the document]
linnan-github: Noted that it can be an arbitrary URL.

It looks like the Reporting API doesn't require the reporting endpoint to be same-origin. Do we need this restriction?

EVENT.md
```diff
@@ -603,6 +604,65 @@ Set-Cookie: ar_debug=1; SameSite=None; Secure; HttpOnly
...
+
+The error reports will be sent to a new endpoint:
+```
+https://<reporting origin>/.well-known/attribution-reporting/error
+```
+
+The top-level sites may opt in to receiving error reports by declaring a reporting
+endpoint via an HTTP header in the top-level document response:
+```
+Attribution-Reporting-Error-Endpoint: https://example.com/reports // an arbitrary URL
```
apasel422, line 663: To avoid confusion, I'd replace the `// an arbitrary URL` comment with a new paragraph beneath the code block, and probably also mention that while the URL can be arbitrary, its origin must be potentially trustworthy.

apasel422: Sure, that's fine.

apasel422 (approved):

EVENT.md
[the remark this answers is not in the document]
maudnals: Chiming in from a developer experience standpoint: unless there's a security/privacy reason to mandate same-origin, there's value in this URL being potentially arbitrary. This opens the possibility for developers to leverage third-party error monitoring services, as it's the case today for Reporting API. While these services don't exist yet for Attribution Reporting, they may in the future.

EVENT.md
```diff
@@ -603,6 +604,59 @@ Set-Cookie: ar_debug=1; SameSite=None; Secure; HttpOnly
...
+
+* a source is rejected due to the [destination
+  limits](#limiting-the-number-of-unique-destinations-covered-by-pending-sources)
+
+The browser will send non-credentialed secure HTTP `POST` requests to the
+reporting endpoints, see [below](#reporting-endpoints). The report data is
+included in the request body as a JSON object:
+```jsonc
+{
+  "type": "<report type>", // "source-destination-limit"
```
apasel422, line 623:
```suggestion
  "type": "<report type>", // e.g. "source-destination-limit"
```

EVENT.md
```diff
@@ -603,6 +604,59 @@ Set-Cookie: ar_debug=1; SameSite=None; Secure; HttpOnly
...
+    "attribution_destination": "https://destination.example"
+  }
+}
+```
+
+These error reports will be sent immediately upon the error occuring during
+source registration.
+
+Note that if other failures are reported using this framework, the browser may
+enforce a random small delay and omit some data (e.g. `source_event_id`) to
```
apasel422, line 637:
```suggestion
enforce a small random delay and omit some data (e.g. `source_event_id`) to
```

apasel422 (changes requested):

apasel422 (approved):

EVENT.md
```diff
@@ -603,6 +604,59 @@ Set-Cookie: ar_debug=1; SameSite=None; Secure; HttpOnly
...
+  "body": {
+    "limit": 100, // the browser's limit
+    "source_event_id": "<source event id in the source registration>",
+    "source_site": "https://source.example",
+    "attribution_destination": "https://destination.example"
+  }
+}
+```
+
+These error reports will be sent immediately upon the error occuring during
```
apasel422, line 633:
```suggestion
These error reports will be sent immediately upon the error occurring during
```

apasel422 (changes requested):

apasel422 (approved):

csharrison:
This seems OK to me, but can you:
1. Link to https://github.com/WICG/attribution-reporting-api/issues/160 to attribute to the corresponding issue
2. Mention why there are not more error reports and for what reason some are omitted?

Also, does this proposal have a hard dependency on https://github.com/WICG/attribution-reporting-api/issues/482?

PR #518, opened 2022-07-13
Make source_event_id field optional

### LINKED ISSUE #517: Make source_event_id optional
In case one is only interested in summary reports, is the `source_event_id` strictly necessary for them? Is there a reason it's particularly helpful for us to nudge developers into using it by mandating it?

If not, shall we consider making this an optional field?

cc @apasel422 @linnan-github

apasel422:
This simplifies source registration for users of the API who only care about aggregatable reports, for which the source_event_id value is irrelevant.

Fixes #517

***
<a href="https://pr-preview.s3.amazonaws.com/apasel422/attribution-reporting-api/pull/518.html" title="Last updated on Jul 26, 2022, 1:48 PM UTC (983ad9a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/attribution-reporting-api/518/9bbe9bd...apasel422:983ad9a.html" title="Last updated on Jul 26, 2022, 1:48 PM UTC (983ad9a)">Diff</a>

### DIFF
--- a/EVENT.md
+++ b/EVENT.md
@@ -184,9 +184,9 @@ header `Attribution-Reporting-Register-Source` of the form:
 - `destination`: an origin whose eTLD+1 is where attribution will be triggered
 for this source.
 
-- `source_event_id`: A string encoding a 64-bit unsigned integer which
+- `source_event_id`: (optional) A string encoding a 64-bit unsigned integer which
 represents the event-level data associated with this source. This will be
-limited to 64 bits of information but the value can vary.
+limited to 64 bits of information but the value can vary. Defaults to 0.
 
 - `expiry`: (optional) expiry in seconds for when the source should be
 deleted. Default is 30 days, with a maximum value of 30 days. The maximum expiry
--- a/header-validator/data.source.js
+++ b/header-validator/data.source.js
@@ -88,10 +88,6 @@ export const invalidSourceHeadersAsObjects = [
     source_event_id: '12340873456',
     destination: 'http://example.com',
   },
-  // Missing required field Source event ID
-  {
-    destination: 'https://example.com',
-  },
   // Missing required field Destination
   {
     source_event_id: '12340873456',
@@ -157,7 +153,7 @@ export const invalidSourceHeadersAsObjects = [
   // Multiple errors
   {
     destination: 'foo',
-    priority: '-12340873456',
+    priority: 'bar',
   },
   // ⚠️ ⚠️ ⚠️ WARNINGS ⚠️ ⚠️ ⚠️
   // Unknown field `foo`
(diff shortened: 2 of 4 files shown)


johnivdel:
I think #517 is discussing a slightly different behavior where a null event_id will ensure that event level reports are not generated (even if the event_trigger_data) is supplied.

apasel422:
> I think #517 is discussing a slightly different behavior where a null event_id will ensure that event level reports are not generated (even if the event_trigger_data) is supplied.

I'm fine with either behavior, though preventing event-report generation is a bigger behavioral change. I'm not sure what @maudnals was actually suggesting.

maudnals:
In the current API, IMU:
(1) if you set a source event ID but no trigger data: The report will not be generated
(2) if you set no source event ID and do set trigger data: This is not supported

While that wasn't initially #517, once we support no source event ID, it seems the behaviour more consistent with (1) would be to not generate the report when trigger data is present, as @johnivdel is suggesting. 
(I may be missing some use cases but it in today's behaviour no source event ID also doesn't generate a report since this fails, so IMU that change wouldn't be breaking)

apasel422:
Allowing a `null` `source_event_id` to disable event-level attribution for a source would seem to complicate the attribution logic.

Consider source A with null event ID and priority 20, source B with non-null event ID and priority 10, and a trigger with event trigger data but no aggregatable trigger data. When we process this trigger, should it be attributed to source B or to nothing?

Then consider the case where that trigger has both event trigger data and aggregatable trigger data.

I think this is easier to reason about with the following:

1. If you don't care about event-level reports at all, then simply never specify event trigger data on your triggers.
2. If you care about event-level reports sometimes, then add something like `"filter_data": {"event-eligible": []}` to your source and the appropriate `filters` / `not_filters` to your tigger data.

Then we can retain the simpler behavior here of defaulting a missing or invalid `source_event_id` to 0, while still making the source eligible for event-level attribution.

maudnals:
> Allowing a null source_event_id to disable event-level attribution for a source would seem to complicate the attribution logic.

I agree!
I meant that no event ID set at all (i.e. absent field from the header) should be allowed, and not generate a report.

apasel422:
> I agree! I meant that no event ID set at all i.e. absent field from the header should be allowed (and not generate a report).

To clarify: What I'm suggesting is that if the source registration field omits the `source_event_id` field (or it is present but invalid), that field's value is defaulted to 0. The source will still be able to be attributed for event-level reports if a trigger containing event trigger data matches it.

In other words, my proposal has no effect on the attribution / report-generation logic, and is simply a surface-level change analogous to how the `priority` fields defaults to 0 if omitted.

maudnals:
> my proposal has no effect on the attribution / report-generation logic, and is simply a surface-level change analogous to how the priority fields defaults to 0 if omitted.

SGTM

johnivdel (approved):
LGTM

PR #527, opened 2022-07-28
Add aggregatable deduplication key field to trigger registration

### LINKED ISSUE #521: [Aggregate API] Tracking multiple trigger types
Splitting the budget in the aggregate API can be difficult especially when tracking multiple trigger types with one type which we want to fire only once per source.
For the sake of the argumentation, we will take an adtech example, but I believe the issue is relevant for other use cases.
An advertiser (or the adtech acting on its behalf) may want to know if users really clicked on its ads, then browsed on its site and finally bought a product. In other words, the advertiser/adtech wants to track two type of trigger events: 

1. a landing, ie that the user did at least visit the website after clicking (in other words it is not a misclick)
2. a sale action (it could be any other desired action on the website though, such as registering to a newsletter, etc).

For reporting purposes, advertisers are interested in having a landing fired only once per click.

In my current understanding of the aggregate API, it is hard to cover this use case easily. There are some hacky ways to do it, but ultimately they do not really fit the bill:

1. Use a priority and deduplication system (as proposed in #472). Pros: easy to implement and mirrors the event-level API. Cons: limited to a window size of roughly 10 minutes as told in the [doc](../blob/main/AGGREGATE.md#aggregatable-reports), hence it does not cover the use case (if a user stays more than 10 minutes on the website, and we generate landing keys for every page of the site, then we will burn the budget).
2. Advertiser/adtech endpoints are stateful, and fire a landing event only once per "session". Pros: no further complication of the API. Cons: same as 1. (ie that it does not really cover our use case), but this time the window size is configurable. Also a big complexity increase for users of the API.

What I am proposing is to add a predefined (at source registration time) budget per key. The sum of budgets should not exceed `L1`. We should find a graceful way out if they do. Budget tracking will be done per key and per source, ie any contribution (at key-level) will be silently dropped if the key budget or the total budget is exhausted.

Another way out would be the mechanism proposed by @johnivdel in #373, if filtering by key is allowed.

Thanks a lot for your comments and inputs on this problem!

P.S. Tagging #485 as this issue is somewhat about budgeting in the aggregate API.

linnan-github:
Fixes #521 

https://github.com/WICG/attribution-reporting-api/issues/521#issuecomment-1195907890


johnivdel (commented):
LGTM

PR #530, opened 2022-08-03
Validate list/object lengths

apasel422:


maudnals (approved):
LGTM.

A few questions though (non-blocking, feel free to merge this PR):

The existence of limits for `maxAggregationKeys`, `maxFilters`, `maxFilterValues`, `maxAggregatableTriggerData` make sense to me.

* Are these limits browser-defined, and why these numbers?
* Are these privacy protections or technical limitations?
* My understanding was that `event_trigger_data` being a list was a quirk in the API, because in practice `event_trigger_data` would always just contain one item (i.e. one `trigger_data, priority, deduplication_key` object). Given the `maxEventTriggerData` is set to `10`, it seems that was wrong. Can you help me understand⏤in which case would one declare several `event_trigger_data` items in the list? Why is the limit 10? (It seems to me this is unrelated to capping and other privacy protections)

apasel422:
> * Are these limits browser-defined, and why these numbers?

Yes, these are implementation-defined. However, we may need to define minimum values that can be relied on across all implementations.

> * Are these privacy protections or technical limitations?

IIRC the filter, aggregatable trigger data, and event trigger data limits are to avoid excessive storage requirements on the user's machine, and the aggregation key limit is privacy-related.

> * My understanding was that `event_trigger_data` being a list was a quirk in the API, because in practice `event_trigger_data` would always just contain one item (i.e. one `trigger_data, priority, deduplication_key` object). Given the `maxEventTriggerData` is set to `10`, it seems that was wrong. Can you help me understand⏤in which case would one declare several `event_trigger_data` items in the list? Why is the limit 10? (It seems to me this is unrelated to capping and other privacy protections)

It's not a quirk, but a feature. The first `event_trigger_data` entry whose `filters` and `not_filters` match the attributed source is the one whose `trigger_data`, `priority`, and `deduplication_key` are used. If none match, no event-level report is created.

maudnals:
Thanks Andrew for the details. 
Two follow-up items on that: one on tooling and one on error reports, but I'll file these outside of this PR.

PR #539, opened 2022-08-10
Fix syntax of Attribution-Reporting-Support header

apasel422:
Commas are used to separate dictionary members, not semi-colons, and per
the spec, values of true (`=?1`) are omitted when serialized.


johnivdel:
LGTM

johnivdel (approved):

PR #556, opened 2022-09-15
Allow source expiry to be independently configured for aggregatable attribution

### LINKED ISSUE #522: Decouple Impression Expiry and Reporting window for Event level Attribution API
**Problem**
Today we have the last reporting window for event-level reports coupled with the impression expiry. The max expiry that Chrome supports is 30 days. However, many consumers of the event level data may need this report sooner than that. To support that, we will have to set smaller impression expiry, but as a result of that we lose all conversions (even the aggregated ones) that happen after that shorter expiry period, hence impacting the reporting needs.

If we have max_event_reporting_window decoupled with the impression’s expiry, then we can get event level data sooner without sacrificing the coverage of Aggregate API.

**Proposal**
We propose to add a new optional field “max_event_reporting_window” in the impression registration JSON.
{
"source_event_id": "12340873456",
"destination": "[eTLD+1]",
"expiry": "[64-bit signed integer]",
"max_event_reporting_window": "[64-bit signed integer]",
"priority": "[64-bit signed integer]"
}

When Chrome determines expiry window for an event, it uses the “max_event_reporting_window” field, if available; Otherwise it falls back to “expiry” field

This proposal is in line with Chrome’s existing treatment of a given event’s eligibility for the two APIs “independently”.

apasel422:
Fixes #522

### DIFF
--- a/AGGREGATE.md
+++ b/AGGREGATE.md
@@ -69,8 +69,8 @@ example an ad-tech will use the API to collect:
 * Aggregate purchase values at a per geo level 
 ### Attribution source registration
 
-Registering sources eligible for aggregate reporting entails adding a new
-`aggregation_keys` dictionary field to the JSON dictionary of the
+Registering sources eligible for aggregate reporting entails adding new
+`aggregation_keys` and `aggregatable_expiry` dictionary fields to the JSON dictionary of the
 [`Attribution-Reporting-Register-Source` header](https://github.com/WICG/conversion-measurement-api/blob/main/EVENT.md#registering-attribution-sources):
 ```jsonc
 {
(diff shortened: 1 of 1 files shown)


csharrison (approved):

JASONWOMACK1977:
Jlwomack1977@gmail.com. JasonWomack1977  womackautomotives@gmail.com.    Womack Automotive Repair

abdulbarokah32-rgb:
Abdulbarokah

abdulbarokah32-rgb:
Abdulbarokah celen

abdulbarokah32-rgb:
[google_privacy_policy_id.pdf](https://github.com/user-attachments/files/23416703/google_privacy_policy_id.pdf)
Abdulbarokah celen48

PR #569, opened 2022-10-03
Change unique destination limit for sources to be time bounded

### LINKED ISSUE #482: Source event limits should be time bounded

linnan-github:
Fixes #482

### DIFF
--- a/EVENT.md
+++ b/EVENT.md
@@ -48,8 +48,8 @@ extension on top of this.
   - [Browsing history reconstruction](#browsing-history-reconstruction)
     - [Adding noise to whether a trigger is
       genuine](#adding-noise-to-whether-a-trigger-is-genuine)
-    - [Limiting the number of unique destinations covered by pending
-      sources](#limiting-the-number-of-unique-destinations-covered-by-pending-sources)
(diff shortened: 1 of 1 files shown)


csharrison (approved):

johnivdel (approved):
LGTM as well

linnan-github:
> Do we still need to [update the spec](https://wicg.github.io/attribution-reporting-api/#issue-1ffadd57) for this?

Updated in https://github.com/WICG/attribution-reporting-api/pull/575

Issue #561, opened 2022-09-22
Consider hour-level expiry within a one-day conversion


logicad:
Currently, Attribution-Reporting-Register-Source's expiry is rounded to the nearest day.

On the other hand, in our system (and business), it is important to distinguish conversions that arise within 24 hours after a click at the hourly level.

Thus, we want the nearest-hour round of the expiry when its value is within 24 hours after the click.

Of course, it may not be appropriate from the viewpoint of the user's privacy.

We would appreciate it if anyone could give any comments on this idea.

csharrison:
Hey @logicad , thanks for filing this issue. I don't think I fully follow what you are asking for though. I think it would help to provide an example of what cases you would set the expiry to <24h and why. What is the shape of the resulting data that you want to obtain?

There might be mechanisms in the API to help you achieve your goals without changing the expiry rounding (which does indeed have some privacy benefit).

logicad:
Hi @csharrison, thank you for your comments.
The reasons of setting the expiry to <24h is  to count only last click or direct conversions from our ad click.
The longer the expiry, the more likely the conversion is not our last click conversion, for example, from search engines or other ad-tech ad clicks.

Currently, about 3% of our advertisers are set up to count only one hour from the ad click as a conversion.

csharrison:
Thanks @logicad this makes sense to me. The concern we had when the expiry is hourly is the following attack:
- User lands on site A
- site A logs 24 impressions, with expiries at hour 24, 23, ... , 2, 1 from the impression.
- User converts at time T, which is attributed to the impression with the smallest expiration that hasn't expired yet
- This lets the site learn the hour the user converted

This attack is somewhat convoluted, and it is mitigated if we:
- Allowing this only when the expiry < 24 hours (limits the damage to differentiating 24 different impressions)
- Allowing this only for clicks (and not views), which makes it much more difficult to have a lot of pending impressions.

But still there is a marginal privacy regression if we allow this. A few questions:
Can you verify that you only need these for clicks?
Would it be acceptable to have coarser data than hourly, e.g. 3-hour granularity, 12-hour, etc? This further mitigates the privacy leak.

Note that there are techniques to do this at the filtering layer which don’t have this negative privacy property (because failing a filter drops the trigger rather than picking another source to attribute to), but doing this with filters is fairly complicated as they were designed for binary matching vs. numerical comparison.

logicad:
@csharrison Thank you for your response.
 
I will answer your questions.
[...]

PR #577, opened 2022-10-06
Replace aggregatable_expiry with event_attribution_window

johnivdel:
***
<a href="https://pr-preview.s3.amazonaws.com/WICG/attribution-reporting-api/pull/577.html" title="Last updated on Oct 25, 2022, 4:51 PM UTC (1e430ff)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/attribution-reporting-api/577/be60ef1...1e430ff.html" title="Last updated on Oct 25, 2022, 4:51 PM UTC (1e430ff)">Diff</a>

### DIFF
--- a/AGGREGATE.md
+++ b/AGGREGATE.md
@@ -84,6 +84,8 @@ Registering sources eligible for aggregate reporting entails adding a new
     // Generates a "0x5" key piece (low order bits of the key) for the key named "geoValue".
     "geoValue": "0x5" // Source-side geo region = 5 (US), out of a possible ~100 regions
   }
+  
+  "aggregatable_report_window": 86400
 }
 ```
 This defines a dictionary of named aggregation keys, each with a piece of the
@@ -93,6 +95,10 @@ trigger-side pieces.
 
 Final keys will be restricted to a maximum of 128 bits. This means that hex
 strings in the JSON must be limited to at most 32 digits.
+
+There is also a new optional field, `aggregatable_report_window`, which is the duration 
+in seconds after source registration during which aggregatable reports may be created 
+for this source.
 
 ### Attribution trigger registration
 
--- a/AGGREGATE.md
+++ b/AGGREGATE.md
@@ -69,8 +69,8 @@ example an ad-tech will use the API to collect:
 * Aggregate purchase values at a per geo level 
 ### Attribution source registration
 
-Registering sources eligible for aggregate reporting entails adding new
-`aggregation_keys` and `aggregatable_expiry` dictionary fields to the JSON dictionary of the
+Registering sources eligible for aggregate reporting entails adding a new
+`aggregation_keys` dictionary field to the JSON dictionary of the
 [`Attribution-Reporting-Register-Source` header](https://github.com/WICG/conversion-measurement-api/blob/main/EVENT.md#registering-attribution-sources):
 ```jsonc
 {
@@ -83,9 +83,7 @@ Registering sources eligible for aggregate reporting entails adding new
 
     // Generates a "0x5" key piece (low order bits of the key) for the key named "geoValue".
     "geoValue": "0x5" // Source-side geo region = 5 (US), out of a possible ~100 regions
-  },
-
-  "aggregatable_expiry": "[64-bit signed integer]"
+  }
 }
 ```
 This defines a dictionary of named aggregation keys, each with a piece of the
@@ -95,11 +93,6 @@ trigger-side pieces.
 
 Final keys will be restricted to a maximum of 128 bits. This means that hex
 strings in the JSON must be limited to at most 32 digits.
-
-`aggregatable_expiry` is an optional string field containing a signed 64-bit
-integer representing the number of seconds for which the source is eligible for
-aggregatable attribution. If omitted, the field defaults to the value specified
-in the `expiry` field.
 
 ### Attribution trigger registration
 
--- a/AGGREGATE.md
+++ b/AGGREGATE.md
@@ -83,9 +83,9 @@ Registering sources eligible for aggregate reporting entails adding a new
 
     // Generates a "0x5" key piece (low order bits of the key) for the key named "geoValue".
     "geoValue": "0x5" // Source-side geo region = 5 (US), out of a possible ~100 regions
-  }
-  
-  "aggregatable_report_window": 86400
+  },
+
+  "aggregatable_report_window": "86400"
 }
 ```
 This defines a dictionary of named aggregation keys, each with a piece of the
--- a/EVENT.md
+++ b/EVENT.md
@@ -200,9 +200,9 @@ received, the browser will find the matching source with highest
 `priority` value and generate a report. The other sources will not
 generate reports.
 
-- `event_report_window`: (optional) time in seconds relative to the source
-where event reports may be created for this source. Also controls the last
-window where reports will be sent.
+- `event_report_window`: (optional) duration in seconds relative to the source
+during which event reports may be created for this source. Also controls the last
+window in which reports will be sent.
 
 
 Once this header is received, the browser will proceed with [handling an
--- a/EVENT.md
+++ b/EVENT.md
@@ -200,10 +200,9 @@ received, the browser will find the matching source with highest
 `priority` value and generate a report. The other sources will not
 generate reports.
 
-- `event_report_window`: (optional) duration in seconds relative to the source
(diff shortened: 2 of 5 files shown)


EVENT.md
```diff
@@ -199,6 +200,11 @@ received, the browser will find the matching source with highest
 `priority` value and generate a report. The other sources will not
 generate reports.
 
+- `event_report_window`: (optional) time in seconds relative to the source
+where event reports may be created for this source. Also controls the last
+window where reports will be sent.
```
apasel422, line 205:
```suggestion
- `event_report_window`: (optional) duration in seconds relative to the source
during which event reports may be created for this source. Also controls the last
window in which reports will be sent.
```

index.bs
```diff
@@ -1310,11 +1316,10 @@ To <dfn>trigger attribution</dfn> given an [=attribution trigger=] |trigger| run
 
 <h3 algorithm id="delivery-time">Establishing report delivery time</h3>
 
-To <dfn>obtain an expiry deadline</dfn> given an [=attribution source=]
+To <dfn>obtain an event report window deadline</dfn> given an [=attribution source=]
 |source|:
 
-1. Return the maximum of |source|'s [=attribution source/expiry=] and 2 days
-    (172800 seconds).
+1. Return |source|'s [=attribution source/event report window=].
```
apasel422, line 1322: Is it still worth keeping this definition around if it's trivially returning the field?

apasel422 (changes requested):
Please also update the header validator to remove the `aggregatable_expiry` field and add the new one.

csharrison:
Two points:
1. We should consider making event_attribution_window extensible enough that it could take a begin and end time. This satisfies the use-case of partitioning users/sources across a lookback window (e.g. some sources look at 1-7 days, some only at 7-14, some only at 14-30). I don't think it needs to be in this PR but we should consider this extension for the future (e.g. you can input either a scalar or a tuple here).
2. Don't we also want an aggregate_attribution_window?

index.bs
```diff
@@ -1227,6 +1232,7 @@ To <dfn>trigger attribution</dfn> given an [=attribution trigger=] |trigger| run
 1. Let |sourceToAttribute| be the first item in |matchingSources|.
 1. [=Assert=]: |sourceToAttribute|'s [=attribution source/randomized response=] is
     null or an [=set/is empty|empty=] [=set=].
+1. If |sourceToAttribute|'s [=attribution source/event report window time=] is less than the current time, return.
```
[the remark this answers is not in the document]
johnivdel: I don't think we can, as we still want to include them in sources collected for attribution just skip event level report generation. This will pobably be more complicated when agg reports are added to this algorithm.

johnivdel:
> Two points:
> 
> 1. We should consider making event_attribution_window extensible enough that it could take a begin and end time. This satisfies the use-case of partitioning users/sources across a lookback window (e.g. some sources look at 1-7 days, some only at 7-14, some only at 14-30). I don't think it needs to be in this PR but we should consider this extension for the future (e.g. you can input either a scalar or a tuple here).
> 2. Don't we also want an aggregate_attribution_window?

1)

I think that is reasonable, but I don't think using a single value will make swapping to a tuple or similar too difficult, so I would prefer to err on the side of YAGNI.

2) Good question, the issue in question seems to only request the capability for event level reports, which is why I only added an event window param. I think we could add one for aggregate to make this more flexible. That being said, do we have a use-case where you would want to stop generating aggregate reports and only generate event reports?

johnivdel:
> Please also update the header validator to remove the `aggregatable_expiry` field and add the new one.

Done, PT another look.

csharrison:
> 1. I think that is reasonable, but I don't think using a single value will make swapping to a tuple or similar too difficult, so I would prefer to err on the side of YAGNI.

That's fine with me.


> 2. Good question, the issue in question seems to only request the capability for event level reports, which is why I only added an event window param. I think we could add one for aggregate to make this more flexible. That being said, do we have a use-case where you would want to stop generating aggregate reports and only generate event reports?

I was thinking of use-cases that would use this capability and also want consistency with aggregatable reports. For instance,  the use-case in https://github.com/WICG/attribution-reporting-api/issues/561 might take advantage of this for both reports, where we have privacy restrictions about small `expiry` values but no privacy restrictions for these windows because they take effect after attribution matching.

johnivdel:
> > 1. I think that is reasonable, but I don't think using a single value will make swapping to a tuple or similar too difficult, so I would prefer to err on the side of YAGNI.
> 
> That's fine with me.
> 
> > 2. Good question, the issue in question seems to only request the capability for event level reports, which is why I only added an event window param. I think we could add one for aggregate to make this more flexible. That being said, do we have a use-case where you would want to stop generating aggregate reports and only generate event reports?
> 
> I was thinking of use-cases that would use this capability and also want consistency with aggregatable reports. For instance, the use-case in #561 might take advantage of this for both reports, where we have privacy restrictions about small `expiry` values but no privacy restrictions for these windows because they take effect after attribution matching.

If they want consistency, they would be able to set expiry == event_report_window. Do we think a reporter would ever want to:

1. set aggregate_report_window < event_report_window
2. set both aggregate_report_window and event_report_window < expiry?

EVENT.md
```diff
@@ -199,6 +200,11 @@ received, the browser will find the matching source with highest
 `priority` value and generate a report. The other sources will not
 generate reports.
 
+- `event_report_window`: (optional) duration in seconds relative to the source
+during which event reports may be created for this source. Also controls the last
+window in which reports will be sent.
+
+
```
apasel422, line 207:
```suggestion
```

header-validator/validate-json.js
```diff
@@ -198,7 +198,6 @@ const aggregationKeys = object((state, key, value) => {
 export function validateSource(source) {
   const state = new State()
   state.validate(source, {
-    aggregatable_expiry: optional(int64),
```
apasel422, line 201: Shouldn't this be replaced by the new field?

header-validator/data.source.js
```diff
@@ -9,7 +9,6 @@ export const validSourceHeadersAsObjects = [
     priority: '4567898765678',
     debug_key: '9876786543456',
     expiry: '123',
-    aggregatable_expiry: '456',
```
apasel422, line 12: Shouldn't this be replaced by the new field?

EVENT.md
```diff
@@ -199,6 +200,11 @@ received, the browser will find the matching source with highest
 `priority` value and generate a report. The other sources will not
 generate reports.
 
+- `event_report_window`: (optional) duration in seconds relative to the source
```
apasel422, line 203: `relative to the source` is a bit ambiguous to me. Maybe `duration in seconds after source registration during which...`?

apasel422 (changes requested):

EVENT.md
```diff
@@ -419,11 +424,10 @@ different sites.
 
 ### Sending Scheduled Reports
 
-Reports for `event` sources will be sent 1 hour after the source expires at its
-`expiry`.
+Reports for `event` sources will be sent 1 hour after the sources `event_report_window`.
```
apasel422, line 427:
```suggestion
Reports for `event` sources will be sent 1 hour after the source's `event_report_window`.
```

AGGREGATE.md
```diff
@@ -83,9 +83,9 @@ Registering sources eligible for aggregate reporting entails adding new
 
     // Generates a "0x5" key piece (low order bits of the key) for the key named "geoValue".
     "geoValue": "0x5" // Source-side geo region = 5 (US), out of a possible ~100 regions
-  },
-
-  "aggregatable_expiry": "[64-bit signed integer]"
+  }
+  
+  "aggregatable_report_window": 86400
```
apasel422, line 88:
```suggestion
  "aggregatable_report_window": "86400"
```

AGGREGATE.md
```diff
@@ -83,9 +83,9 @@ Registering sources eligible for aggregate reporting entails adding new
 
     // Generates a "0x5" key piece (low order bits of the key) for the key named "geoValue".
     "geoValue": "0x5" // Source-side geo region = 5 (US), out of a possible ~100 regions
-  },
-
-  "aggregatable_expiry": "[64-bit signed integer]"
+  }
+
```
apasel422, line 87:
```suggestion
  },

```

apasel422 (changes requested):

apasel422 (approved):

PR #589, opened 2022-10-21
Extend verbose debug reports for cookie-based debug data

linnan-github:


apasel422 (approved):

csharrison (approved):
LGTM. Maybe in the debug type we should include something in the name that indicates whether it requires the cookie.

PR #591, opened 2022-10-25
Clarify language on bad OS response headers

csharrison:
The current language is vague and can be interpreted in two ways.


johnivdel (approved):
LGTM

PR #592, opened 2022-10-26
Remove landing page restriction for navigation sources

### LINKED ISSUE #590: Consider removing landing page destination restriction
Currently in the API for navigation sources, we have a restriction that a user must land on the destination site for the source to be valid. Otherwise, the source registration is ignored.

We are hearing cases of legitimate use-cases where this restriction is difficult to conform to, e.g.
- Landing pages that are managed by third parties with different domains than the advertiser site
- Conversion pages that are managed by third parties (e.g. purchase flows) on a different domain than the advertiser site or landing page

Additionally, the restriction in the API is no longer fully “load bearing” because event sources already have the capability to leak some amount of browsing history, which we protect with noise that is also present on navigation sources. Therefore it doesn’t feel as necessary to add additional protections on clicks which are already rate-limited by user interaction.

Note: this doesn’t fully solve the use cases detailed in https://github.com/WICG/attribution-reporting-api/issues/549 but it does partially solve the case where only a single domain is required to register conversions. Note that this also partially fixes https://github.com/WICG/attribution-reporting-api/issues/44.

csharrison:
Fixes #590

***
<a href="https://pr-preview.s3.amazonaws.com/WICG/attribution-reporting-api/pull/592.html" title="Last updated on Oct 26, 2022, 9:07 PM UTC (d844202)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/attribution-reporting-api/592/0b4ce8a...d844202.html" title="Last updated on Oct 26, 2022, 9:07 PM UTC (d844202)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -173,9 +173,6 @@ to set the [=navigation params/attribution source=] of |navigationParams| to |at
 
 <h4 id="monkeypatch-document-creation">Document creation</h4>
 
-At the time <a spec="html">create and initialize a <code>Document</code> object</a> is invoked, the user agent knows the final URL used for the
-navigation and can validate the [=attribution source/attribution destination=].
-
 In <a spec="html">create and initialize a <code>Document</code> object</a>, before
 
 > 2. Let permissionsPolicy be the result of creating a permissions policy from a response given browsingContext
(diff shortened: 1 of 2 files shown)


index.bs
```diff
@@ -173,9 +173,6 @@ to set the [=navigation params/attribution source=] of |navigationParams| to |at
 
 <h4 id="monkeypatch-document-creation">Document creation</h4>
 
-At the time <a spec="html">create and initialize a <code>Document</code> object</a> is invoked, the user agent knows the final URL used for the
```
johnivdel, line 176: I think most likely all of this routing through navigate steps can be ripped out, but given the attributionsrc= side isn't implemented in the draft spec, I think this state is fine for now.

johnivdel (approved):
LGTM

index.bs
csharrison: Yeah, I agree.

apasel422 (approved):

Issue #549, opened 2022-08-29
Multiple Impression Registrations on Existing Tracking Pings


chandan-giri:
**Multi Conversion Domain Tracking**
Advertisers may track conversions on different conversion domains than original click/impression destinations. Eg: User clicks on an Ad and lands on advertiser.us, performs interactions to further navigate and convert on advertiser.com. If we register for only the click/impression domain, AdTech will not be able to track conversions on advertiser.com. It can also happen for the above case that AdTech is tracking page view conversion on advertiser.us and purchase conversion on advertiser.com domain. For a given ad click or impression, AdTech may need to register multiple impressions for a given Ad.
1. The additional impressions are always registered as views. For these additional impressions registrations, view validations and corresponding 1 conversion per view limits for EventApi will apply. This is true even when the original event is a click. Hence for these additional impression registration, we will not validate that registration destination matches click destination.
2. The “source_event_id” of these impression registrations may be all different and depend on AdTech implementation.
3. Aggregation dimensions may change from original event registration and depend on AdTech implementation. AdTech may want to measure conversions for these x-domain registration slices. Eg: For the case where the original event is a click, 3 bits conversion metadata and 3 conversions per click is tracked for original click whereas 1 bit conversion metadata and 1 conversion per click is tracked for additional view registrations.
4. Attribution filters may change from original events since newer aggregation dimensions and conversion metadata may be output differently for cross-domain registrations. Eg: For the original click 3 bits of metadata will be output, however for additional x-domain view registrations, only 1 bit of metadata is allowed.

**Constraints With Current Proposal**
1. Support for 3rd party trackers: 3rd party trackers can only register impressions on click tracking pings. Currently there is no mechanism available to register additional views to track cross-domain conversions. Also for click, these cross-domain registrations will be rejected since destination doesn’t match click destination.
2. Rendering/Serving overhead: For current registration mechanism, changes are likely required in rendering/serving of all ad formats to fire new registration pings. Newer feature requests will remove these overheads and make integration more convenient.
3. Additional Network Requests: Since currently only one registration is allowed per registration ping, new registration pings order of magnitude of additional cross-domain registrations is generated. The # of request pings will increase by lot for such use cases.

**Feature Request**
We are requesting the following extension to the current impression registration mechanism
1. Allow for multiple events registration for given registration ping.
2. Allow for events registration to be downgraded as view. For these explicitly downgraded views’ registration, the view registration validation will apply.
[...]

PR #601, opened 2022-11-04
Update explainer and spec to support multiple destinations

### LINKED ISSUE #44: Conversions coming from a different domain than landing page
Hi @csharrison and @michaelkleber !

## Context

In past W3C web-adv calls, we've discussed the case where conversions come from a different domain than the landing page's domain. This could happen for a number of reasons--localized websites (i.e. shop.com taking you to shop.co.uk), commerce platforms (i.e. my-store.com taking you to my-store.shopify.com) or others (i.e. gap.com taking you to gapfactory.com).

My understanding is that the current draft specification would lose the conversion. From the calls, it seems that people believe this is a legitimate use case, so I'm eager to work with you all to come up with a privacy-safe way of supporting this use case.

I'm not attached to any particular solution if we can solve these problems, but to kick off the discussion, here are a few thoughts:

## Delegation of trust

This could be either implicit or explicit. When my-store.com (the "landing page domain") takes you to my-store.shopify.com (the "destination domain"), it could be annotated (in the anchor tag or redirect) that it's delegating the reporting permissions to the destination. When conversions finally happen on the destination domain, they are reported to the publisher as if they came from the landing page domain. 

This introduces more complexity because the browser has to track the pathway to the conversion. I don't think this requires significant amounts of new data but I appreciate that it has its costs.

Do people see privacy concerns with this?

## Support of multiple domains in the original ad annotation

Instead of listing just one domain, perhaps the publisher can list multiple domains. There would have to be some limit to the number of domains, with perhaps an exception for country-level TLD variations. This could also be used in conjunction with the above proposal to add more protections, so that both the publisher side and the advertiser side need to certify the delegation.

## Quick-redirect through the conversion domain

This was brought up by @csharrison during a call. It has a few downsides:

1. It presumes that the publisher knows where the conversion will happen. This is the case for commerce platforms but not the case for multi-national domains.
2. It adds another redirect overhead to the initial page load, which is a bad user experience
3. I'm not sure that it's compatible with PCM's privacy model. My understanding is that the ad must land on the `adDestination` domain, not merely traverse through it. 

## Next steps

How do we move this discussion forward? Are there other ideas for addressing this major use case without compromising privacy? I'm open to feedback on how we could build upon or improve any of the ideas above, or any other ones that can solve this problem.

Thanks!

csharrison:
Fixes #44 as far as I can tell. This PR attempts to be a fully backwards compatible change, so existing deployments will not have to update any code to accomodate.

Addresses #549, not calling it fixed until further verification.

***
<a href="https://pr-preview.s3.amazonaws.com/WICG/attribution-reporting-api/pull/601.html" title="Last updated on Nov 7, 2022, 6:32 PM UTC (f9ac370)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/attribution-reporting-api/601/2db4718...f9ac370.html" title="Last updated on Nov 7, 2022, 6:32 PM UTC (f9ac370)">Diff</a>

### DIFF
--- a/EVENT.md
+++ b/EVENT.md
@@ -184,7 +184,8 @@ header `Attribution-Reporting-Register-Source` of the form:
 ```
 
 - `destination`: an origin whose eTLD+1 is where attribution will be triggered
-for this source.
+for this source. This field can optionally be repeated in a list of some small size
+(e.g. 3-5).
 
 - `source_event_id`: (optional) A string encoding a 64-bit unsigned integer which
 represents the event-level data associated with this source. This will be
@@ -217,8 +218,8 @@ since it is the origin that will end up receiving attribution reports.
 A `navigation` attribution source event will be logged to storage if  the navigation occurs with [transient
 user activation](https://html.spec.whatwg.org/multipage/interaction.html#transient-activation). `event` sources don’t require activation.
 
-An attribution source will be eligible for reporting if any page on the
-`destination` eTLD+1 (advertiser site) triggers attribution for the associated
+An attribution source will be eligible for reporting if any page on any of the
+associated `destination` eTLD+1s (advertiser sites) triggers attribution for the associated
 reporting origin.
 
 ### Publisher-side Controls for Attribution Source Declaration
@@ -257,7 +258,7 @@ issue](https://github.com/w3c/webappsec-permissions-policy/issues/252).
 ### Triggering Attribution
 
 Attribution can only be triggered for a source on a page whose eTLD+1 matches
-the eTLD+1 of the site provided in `destination`. To trigger attribution, a
+the eTLD+1 of (one of) the site(s) provided in `destination`. To trigger attribution, a
 similar mechanism is used as source event registration, via HTML:
 ```html
 <img src="https://ad-tech.example/conversionpixel"
@@ -380,7 +381,7 @@ reflect a final set of parameters.
 
 ### Trigger attribution algorithm
 
-When the browser receives an attribution trigger redirect on a URL matching the
+When the browser receives an attribution trigger redirect on a URL matching a
 `destination` eTLD+1, it looks up all sources in storage that match
 <`attributionsrc` origin, `destination`> and picks the one with the greatest
 `priority`. If multiple sources have the greatest `priority`, the
@@ -858,10 +859,10 @@ browsers.
 ### Reporting cooldown / rate limits
 
 To limit the amount of user identity leakage between a <source site,
-destination> pair, the browser should throttle the amount of total information
+ site> pair, the browser should throttle the amount of total information
 sent through this API in a given time period for a user. The browser should set
 a maximum number of attributions per
-<source site, destination, reporting origin, user> tuple per time period. If this
+<source site, destination site, reporting origin, user> tuple per time period. If this
 threshold is hit, the browser will stop scheduling reports the API for the
 rest of the time period for attributions matching that tuple.
 
@@ -874,7 +875,7 @@ leak when multiple origins collude with each other. However, the alternative
 makes it very difficult to adopt the API if all reporting origins had to share a
 budget.
 
-Strawman rate limit: 100 attributions per {source site, destination, reporting
+Strawman rate limit: 100 attributions per {source site, destination site, reporting
 origin, 30 days}
 
 ### Less trigger-side data
@@ -898,13 +899,10 @@ attribution window a source was triggered.
 
 ### Browsing history reconstruction
 
-Reporting attribution without a pre-existing navigation allows the
-`reportingorigin` to learn whether a given user on the source site visited the
-`destination` site at all. For click-through reports, this is not an issue
-because the `reportingorigin` knows a priori the user was navigating to
-`destination`.
+Reporting attribution allows the `reportingorigin` to learn whether a given user
+on the source site visited (one of) the `destination` site(s) at all.
 
-This new threat is be mitigated in a number of ways:
+This threat is be mitigated in a number of ways:
 
 #### Adding noise to whether a trigger is genuine
 
@@ -920,7 +918,7 @@ trying to measure user visits on, the browser can limit the number `destination`
 eTLD+1s represented by unexpired sources for a source-site.
 
 The browser can place a limit on the number of a source site's unexpired source's
-unique `destination`s. When an attribution source is registered for an eTLD+1
+unique `destination` sites. When an attribution source is registered for an eTLD+1
 that is not already in the unexpired sources and a source site is at its limit,
 the browser will drop the new source.
 
--- a/EVENT.md
+++ b/EVENT.md
@@ -859,7 +859,7 @@ browsers.
 ### Reporting cooldown / rate limits
 
 To limit the amount of user identity leakage between a <source site,
- site> pair, the browser should throttle the amount of total information
+destination site> pair, the browser should throttle the amount of total information
 sent through this API in a given time period for a user. The browser should set
 a maximum number of attributions per
 <source site, destination site, reporting origin, user> tuple per time period. If this
--- a/index.bs
+++ b/index.bs
@@ -42,7 +42,7 @@ send [=attribution reports=] containing information from both of those events.
 
 A reporter `https://reporter.example` embedded on `https://source.example` is able to
 measure whether an iteraction on the page lead to an action on `https://destination.example`
-by registering an [=attribution source=] with an [=attribution source/attribution destinations=]
+by registering an [=attribution source=] with [=attribution source/attribution destinations=]
 of « `https://destination.example` ». Reporters are able to register sources through a variety
 of surfaces, but ultimately the reporter is required to provide the User Agent with an
 HTTP-response header which allows the source to be eligible for attribution.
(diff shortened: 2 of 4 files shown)


index.bs
```diff
@@ -42,8 +42,8 @@ send [=attribution reports=] containing information from both of those events.
 
 A reporter `https://reporter.example` embedded on `https://source.example` is able to
 measure whether an iteraction on the page lead to an action on `https://destination.example`
-by registering an [=attribution source=] with an [=attribution source/attribution destination=]
-of `https://destination.example`. Reporters are able to register sources through a variety
+by registering an [=attribution source=] with an [=attribution source/attribution destinations=]
```
apasel422, line 45:
```suggestion
by registering an [=attribution source=] with [=attribution source/attribution destinations=]
```

index.bs
```diff
@@ -275,8 +275,8 @@ An attribution source is a [=struct=] with the following items:
 :: An [=url/origin=].
 : <dfn>event ID</dfn>
 :: A non-negative 64-bit integer.
-: <dfn>attribution destination</dfn>
-:: A [=site=].
+: <dfn>attribution destinations</dfn>
+:: An [=ordered set=] of [=site=].
```
apasel422, line 279:
```suggestion
:: An [=ordered set=] of [=sites=].
```

index.bs
```diff
@@ -571,12 +573,12 @@ for triggers that are attributed to an [=attribution source=] whose
 
 <dfn>Max event-level reports per attribution destination</dfn> is a positive integer that
 controls how many [=event-level reports=] can be in the
-[=event-level report cache=] per
-[=event-level report/attribution destination=].
+[=event-level report cache=] per each [=site=] in
```
apasel422, line 576:
```suggestion
[=event-level report cache=] per [=site=] in
```

index.bs
```diff
@@ -737,6 +739,23 @@ To <dfn>parse an attribution destination</dfn> from a string |str|:
...
+To <dfn>parse attribution destinations</dfn> from a value |val|:
+1. Let |result| be an [=ordered set=].
+1. If |val| is a [=string=], [=set/append=] the result of [=parse an attribution destination=] to |result|, and return |result|.
+1. If |val| is not a [=list=], return null.
+1. [=list/iterate|For each=] |value| of |val|
+    1. If |value| is not a [=string=], return null.
+    1. Let |destination| be the result of [=parse an attribution destination=] with |value|.
+    1. If |destination| is null, return null.
+    1. [=set/append=] |destination| to |result|.
+1. If |result|'s [=set/size=] is greater than 3, return null.
```
apasel422, line 753: Should we explicitly make 3 something that is specified by the user agent?

index.bs
```diff
@@ -933,16 +951,12 @@ To <dfn>check if an [=attribution source=] exceeds the unexpired destination lim
      * |record|'s [=attribution rate-limit record/source site=] and |source|'s [=attribution source/source site=] are equal
      * |record|'s [=attribution rate-limit record/reporting endpoint=] and |source|'s [=attribution source/reporting endpoint=] are equal
      * |record|'s [=attribution rate-limit record/expiry time=] is greater than |source|'s [=attribution source/source time=]
+1. Let |unexpiredDestinations| be an [=set=]
```
apasel422, line 954:
```suggestion
1. Let |unexpiredDestinations| be an [=set/empty=] [=set=].
```

index.bs
```diff
@@ -933,16 +951,12 @@ To <dfn>check if an [=attribution source=] exceeds the unexpired destination lim
...
-    1. If |source|'s [=attribution source/attribution destination=] and |unexpiredSource|'s
-        [=attribution source/attribution destination=] are equal, then return false.
-    1. If |distinctDestinations| [=set/contains=] |unexpiredSource|'s
-        [=attribution source/attribution destination=], [=iteration/continue=].
-    1. [=set/Append=] |unexpiredSource|'s [=attribution source/attribution destination=] to
-        |distinctDestinations|.
-1. If |distinctDestinations|'s [=set/size=] is greater than or equal to
-    the user agent's [=max destinations covered by unexpired sources=], then return true.
-1. Return false.
+    1. Let |unexpiredDestinations| be the result of taking the [=set/union=] of |unexpiredDestinations| and |unexpiredSource|'s [=attribution source/attribution destinations=]'
```
apasel422, line 956:
```suggestion
    1. Let |unexpiredDestinations| be the result of taking the [=set/union=] of |unexpiredDestinations| and |unexpiredSource|'s [=attribution source/attribution destinations=].
```

index.bs
```diff
@@ -933,16 +951,12 @@ To <dfn>check if an [=attribution source=] exceeds the unexpired destination lim
...
+1. Let |newDestinations| be the result of taking the [=set/union=] of |unexpiredDestinations| and |source|'s [=attribution source/attribution destinations=]'
```
apasel422, line 957:
```suggestion
1. Let |newDestinations| be the result of taking the [=set/union=] of |unexpiredDestinations| and |source|'s [=attribution source/attribution destinations=].
```

index.bs
```diff
@@ -933,16 +951,12 @@ To <dfn>check if an [=attribution source=] exceeds the unexpired destination lim
...
+1. Return whether |newDestinations|'s [=set/size=]' is  is greater than or equal to the user agent's [=max destinations covered by unexpired sources=].
```
apasel422, line 958:
```suggestion
1. Return whether |newDestinations|'s [=set/size=] is greater than or equal to the user agent's [=max destinations covered by unexpired sources=].
```

index.bs
```diff
@@ -999,7 +1013,7 @@ an [=attribution source=] |source|:
 1. If |source|'s [=attribution source/debug reporting enabled=] is false, return null.
 1. Let |body| be a new [=map=] with the following key/value pairs:
     : "`attribution_destination`"
-    :: |source|'s [=attribution source/attribution destination=].
+    :: |source|'s [=attribution source/attribution destinations=].
```
apasel422, line 1016: This should use the algorithm you defined as "serialize [=event-level report/attribution destinations=]"

index.bs
```diff
@@ -1647,14 +1662,27 @@ To <dfn>serialize an integer</dfn>, represent it as a string of the shortest pos
 Issue: This would ideally be replaced by a more descriptive algorithm in Infra. See
 <a href="https://github.com/whatwg/infra/issues/201">infra/201</a>
 
+<h3 id="serialize-destinations">Serialize attribution destinations</h3>
+
+To <dfn>serialize [=event-level report/attribution destinations=]</dfn> |destinations|, run the following steps:
+
+1. [=Assert=]: |destinations| does not [=set/contain=] the [=opaque origin=].
```
apasel422, line 1669: Not sure this works exactly. I would just assert that each destination is not opaque within the `for each` below.

index.bs
```diff
@@ -1647,14 +1662,27 @@ To <dfn>serialize an integer</dfn>, represent it as a string of the shortest pos
...
+1. [=Assert=]: |destinations| is not [=set/empty=].
+1. Let |destinationStrings| be a [=list=]
+1. [=list/iterate|For each=] |destination| in |destinations|:
+    1. [=list/append=] |destination| <a href="https://html.spec.whatwg.org/multipage/origin.html#serialization-of-a-site">serialized</a> to |destinationStrings|
+1. If |destinationStrings|'s [=set/size=] is equal to 1, return |destinationStrings|[0].
+1. Return the result of running [=serialize an infra value to JSON bytes=] on |destinationStrings|.
```
apasel422, line 1675: Do we actually want to serialize this here? It seems like the final report body should be:

```json
{
  "attribution_destination": ["a", "b", "c"]
}
```
or
```json
{
  "attribution_destination": "a"
}
```
rather than
```json
{
  "attribution_destination": "\"a\",\"b\",\"c\""
}
```

apasel422 (changes requested):

apasel422 (approved):

apasel422:
@csharrison Would it be possible for you to update the header validator in this PR as well? If not, I can do it in a followup.

csharrison:
> @csharrison Would it be possible for you to update the header validator in this PR as well? If not, I can do it in a followup.

Done

apasel422 (approved):

EVENT.md
```diff
@@ -858,10 +859,10 @@ browsers.
 ### Reporting cooldown / rate limits
 
 To limit the amount of user identity leakage between a <source site,
-destination> pair, the browser should throttle the amount of total information
+ site> pair, the browser should throttle the amount of total information
```
johnivdel, line 862: Should we still include destination here?

index.bs
```diff
@@ -933,16 +951,12 @@ To <dfn>check if an [=attribution source=] exceeds the unexpired destination lim
...
-    1. If |source|'s [=attribution source/attribution destination=] and |unexpiredSource|'s
-        [=attribution source/attribution destination=] are equal, then return false.
-    1. If |distinctDestinations| [=set/contains=] |unexpiredSource|'s
-        [=attribution source/attribution destination=], [=iteration/continue=].
-    1. [=set/Append=] |unexpiredSource|'s [=attribution source/attribution destination=] to
-        |distinctDestinations|.
-1. If |distinctDestinations|'s [=set/size=] is greater than or equal to
-    the user agent's [=max destinations covered by unexpired sources=], then return true.
-1. Return false.
+    1. Let |unexpiredDestinations| be the result of taking the [=set/union=] of |unexpiredDestinations| and |unexpiredSource|'s [=attribution source/attribution destinations=].
```
johnivdel, line 956: unexpiredSource is a rate-limit record, rather than a source. Given we are adding a rate limit record for each destination, I think we can use set/append here rather than union.

johnivdel (commented):
lgtm % comments

EVENT.md
csharrison: Yep, good catch. Done.

index.bs
csharrison: Done.

johnivdel (approved):
LGTM

PR #602, opened 2022-11-04
Remove destinations from OS-Source header

johnivdel:
With destination matching removed from the core web APIs(https://github.com/WICG/attribution-reporting-api/pull/592), we do not need to mirror it for cross app and web sources.


apasel422:
Please also update:

https://github.com/WICG/attribution-reporting-api/blob/main/header-validator/validate-os.js
https://github.com/WICG/attribution-reporting-api/blob/main/header-validator/test.os.js

apasel422 (changes requested):
we need to update the header validator as well

apasel422 (approved):

csharrison (approved):

PR #603, opened 2022-11-07
Add sequence diagram for cross app/web

csharrison:


app_to_web.md
```diff
@@ -27,6 +27,21 @@ The proposal here takes advantage of OS-level support for attribution. In partic
 *   While focusing initially on Android support, the proposal should be generic enough to be supported on other platforms
 
 ## API changes
+```mermaid
+sequenceDiagram
+  participant B as Browser
+  participant O as OS platform
+  participant R as Reporting endpoint
+  Note over B,R: For either source or trigger registration
+  B->>R: Registration request advertising OS support
```
johnivdel, line 36: optional nitpick: Could consider this and the line below dashed lines as they are crossing over the platform in the middle

johnivdel (approved):
LGTM with optional comment

app_to_web.md
csharrison: Hm I don't feel like that has enough semantic information to warrant changing the line dashing (although to be fair we aren't using sequence diagrams as they are typically formally described in UML).

PR #608, opened 2022-11-09
Use assertions in Eligible header validation instead of log output

apasel422:
This allows the test cases to be validated programatically instead of via manual inspection of log output.

I will do the same for the other tests in followup PRs.


csharrison (approved):
Thank you this is much nicer

PR #612, opened 2022-11-09
Use assertions in JSON header validation instead of log output

apasel422:
This allows the validators to be tested, rather than require them to be manually inspected.


header-validator/data.source.js
```diff
@@ -1,262 +1,325 @@
...
+    }`,
+    expectedErrors: [{
+      path: ["aggregatable_report_window"],
+      msg: "must be an int64 (must match /^-?[0-9]+$/)",
+    }],
   },
-  // Too many filter values
+
   {
-    source_event_id: '0',
```
linnan-github, line 226: Do we plan to add back the test cases for data length?

apasel422: Sure. I'll do that in a followup.

linnan-github (approved):

PR #628, opened 2022-11-22
Spec debug reports for common trigger errors

linnan-github:
Spec debug reports for the following errors common to event-level and aggregatable reports.
* No matching top-level filter data.
* Attributions rate-limit reached.
* Reporting origin rate-limit reached.

***
<a href="https://pr-preview.s3.amazonaws.com/WICG/attribution-reporting-api/pull/628.html" title="Last updated on Nov 22, 2022, 7:53 PM UTC (c9336ab)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/attribution-reporting-api/628/05918d1...c9336ab.html" title="Last updated on Nov 22, 2022, 7:53 PM UTC (c9336ab)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -376,6 +376,9 @@ Possible values are:
 
 <ul>
 <li>"`trigger-no-matching-source`"</li>
+<li>"`trigger-no-matching-filter-data`"</li>
+<li>"`trigger-attributions-per-source-destination-limit`"</li>
+<li>"`trigger-reporting-origin-limit`"</li>
 <li>"`trigger-unknown-error`"</li>
 </ul>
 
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -1365,9 +1377,12 @@ and a [=boolean=] |topLevelFiltersMatch|, run the following steps:
...
 1. If |sourceToAttribute|'s [=attribution source/randomized response=] is not null and is not [=set/is empty|empty=]:
     1. Assert: |sourceToAttribute|'s [=attribution source/event-level attributable=] is false.
-    1. Return <strong>dropped</strong>.
-1. If |topLevelFiltersMatch| is false, return <strong>dropped</strong>.
-1. If |sourceToAttribute|'s [=attribution source/event report window time=] is less than the current time, return <strong>dropped</strong>.
+    1. Return [ "`status`" → "`dropped`" ].
+1. If |topLevelFiltersMatch| is false:
+    1. Let |debugData| be the result of running [=obtain debug data on trigger registration=]
+        with "`trigger-no-matching-filter-data`", |trigger| and |sourceToAttribute|.
+    1. Return [ "`status`" → "`dropped`",  "`debug_data`" → |debugData| ].
```
apasel422, line 1384: Instead of returning a map here, should we introduce a struct type for this? It would allow fields to be cross-referenced automatically.

apasel422 (changes requested):

index.bs
apasel422: If it's really verbose, you could use a tuple struct instead: https://infra.spec.whatwg.org/#tuples

linnan-github: Updated with tuple, thanks.

apasel422 (changes requested):

PR #633, opened 2022-11-23
Update spec to make source keys in aggregatable trigger data optional

### LINKED ISSUE #617: Handling of source_keys field is inconsistent
The spec requires that the `source_keys` field be present on aggregatable trigger data, but allows it to be empty. If an empty list is allowed, we should consider allowing the field to be omitted entirely and treat that case as equivalent to an empty list, for consistency with our other fields.

linnan-github:
Fixes #617.

***
<a href="https://pr-preview.s3.amazonaws.com/linnan-github/conversion-measurement-api/pull/633.html" title="Last updated on Nov 23, 2022, 8:06 PM UTC (1691ab7)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/attribution-reporting-api/633/bfc0e7d...linnan-github:1691ab7.html" title="Last updated on Nov 23, 2022, 8:06 PM UTC (1691ab7)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -777,6 +777,8 @@ To <dfn>parse aggregation keys</dfn> given an [=ordered map=] |map|:
     1. [=map/Set=] |aggregationKeys|[|key|] to |keyPiece|.
 1. Return |aggregationKeys|.
 
+Issue: Determine whether to limit [=string/length=] or [=string/code point length=] for |key| above.
+
 To <dfn noexport>parse source-registration JSON</dfn> given a [=string=]
 |json|, a [=suitable origin=] |sourceOrigin|, a [=suitable origin=] |reportingOrigin|, and a
 [=source type=] |sourceType|:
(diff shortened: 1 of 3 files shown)


index.bs
```diff
@@ -1114,13 +1114,14 @@ To <dfn>parse aggregatable trigger data</dfn> given an [=ordered map=] |map|:
...
-        1. If |sourceKey| is not a [=string=], return null.
-        1. [=set/Append=] |sourceKey| to |sourceKeys|.
+    1. Let |sourceKeys| be a new [=ordered set=].
+    1. If |value|["`source_keys`"] [=map/exists=]:
+        1. If |value|["`source_keys`"] is not a [=list=], return null.
+        1. If |value|["`source_keys`"]'s [=list/size=] is greater than the user agent's
+            [=max aggregation keys per attribution=], return null.
+        1. [=list/iterate|For each=] |sourceKey| of |value|["`source_keys`"]:
+            1. If |sourceKey| is not a [=string=], return null.
+            1. [=set/Append=] |sourceKey| to |sourceKeys|.
```
apasel422, line 1124: Just checking: Are we also supposed to be validating the length of these keys?

apasel422 (approved):

apasel422 (changes requested):
Please also update the header validator.

index.bs
linnan-github: Similar to filters string, we need to decide whether to validate that in the spec.

apasel422 (approved):

PR #637, opened 2022-12-01
Use reporting origin instead of attributionsrc origin for source matching

linnan-github:
The reporting origin doesn't always be attributionsrc origin.

### DIFF
--- a/EVENT.md
+++ b/EVENT.md
@@ -383,7 +383,7 @@ reflect a final set of parameters.
 
 When the browser receives an attribution trigger redirect on a URL matching a
 `destination` eTLD+1, it looks up all sources in storage that match
-<reporting origin, `destination`> and picks the one with the greatest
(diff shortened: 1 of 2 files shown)


EVENT.md
```diff
@@ -383,12 +383,12 @@ reflect a final set of parameters.
 
 When the browser receives an attribution trigger redirect on a URL matching a
 `destination` eTLD+1, it looks up all sources in storage that match
-<`attributionsrc` origin, `destination`> and picks the one with the greatest
+<reporting origin, `destination`> and picks the one with the greatest
```
apasel422, line 386:
```suggestion
<reporting origin, `destination` eTLD+1> and picks the one with the greatest
```

Not sure if we should be usign `eTLD+1` or `site` as our terminology, but we should try to be consistent here and in `verbose_debugging_reports.md`.

apasel422 (changes requested):

PR #638, opened 2022-12-01
Group attribution-success and verbose debugging reports under transitional debugging reports

linnan-github:
This is to avoid confusion with the two types of debugging reports.

### DIFF
--- a/AGGREGATE.md
+++ b/AGGREGATE.md
@@ -12,7 +12,7 @@
   - [Attribution trigger registration](#attribution-trigger-registration)
   - [Aggregatable reports](#aggregatable-reports)
   - [Optional: transitional debugging reports](#optional-transitional-debugging-reports)
-    - [Attribution-succeeded debugging reports](#attribution-succeeded-debugging-reports)
+    - [Attribution-succeed debugging reports](#attribution-succeed-debugging-reports)
     - [Verbose debugging reports](#verbose-debugging-reports)
   - [Contribution bounding and budgeting](#contribution-bounding-and-budgeting)
   - [Storage limits](#storage-limits)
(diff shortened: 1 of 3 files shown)


csharrison:
Hm, if these are the debug reports that go away when cookies go away, it seems a bit confusing to call them "primary". Did you consider other alternatives?

apasel422 (approved):

apasel422:
> Hm, if these are the debug reports that go away when cookies go away, it seems a bit confusing to call them "primary". Did you consider other alternatives?

@csharrison raises a good point. `Transitional debug reports` seems like a good candidate?

linnan-github:
> > Hm, if these are the debug reports that go away when cookies go away, it seems a bit confusing to call them "primary". Did you consider other alternatives?
> 
> @csharrison raises a good point. `Transitional debug reports` seems like a good candidate?

Most of verbose debug reports are transitional as well, so `Transitional debug reports` may not fit.

We've considered `Basic debug reports` as well. @maudnals @akashnadan

maudnals:
Deferring to @akashnadan on the name here

AGGREGATE.md
```diff
@@ -326,9 +329,11 @@ future versions of the API), the endpoint URL should also change.
 **Note:** The browser may need some mechanism to ensure that the same set of
 keys are delivered to different users.
 
-#### Optional: extended debugging reports
+### Optional: transitional debugging reports
 
-If [debugging](https://github.com/WICG/conversion-measurement-api/blob/main/EVENT.md#optional-extended-debugging-reports)
+#### Attribution-succeed debugging reports
```
apasel422, line 334: IMO the name should be either `Attribution-succeeded` or `Attribution-success`. `Attribution-succeed` doesn't seem right grammatically.

apasel422 (changes requested):

AGGREGATE.md
linnan-github: Thanks, updated to Attribution-success debugging reports.

EVENT.md
```diff
@@ -562,6 +563,22 @@ fully understood during roll-out and help flush out any bugs (either in browser
 or caller code), and more easily compare the performance to cookie-based
 alternatives.
 
+To ensure that this data (which could enable cross-site tracking) is only
+available in a transitional phase while third-party cookies are available and
+are already capable of user tracking, the browser will check (at both source
+and trigger registration) for the presence of a special `SameSite=None` cookie
```
apasel422, line 569:
```suggestion
and trigger registration) for the presence of a special cookie
```
No need to repeat this since it's in the example code immediately below.

apasel422 (changes requested):

apasel422 (approved):

PR #622, opened 2022-11-18
Update AGGREGATE.md

akashnadan:
Update to the "Data processing through a Secure Aggregation Service" section. Highlighting future aggregation service cloud support and a new optional field

PR #640, opened 2022-12-02
Spec aggregation coordinator and update header validator

linnan-github:
https://github.com/WICG/attribution-reporting-api/pull/622

***
<a href="https://pr-preview.s3.amazonaws.com/linnan-github/conversion-measurement-api/pull/640.html" title="Last updated on Dec 5, 2022, 11:24 PM UTC (0e1aae1)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/attribution-reporting-api/640/9305760...linnan-github:0e1aae1.html" title="Last updated on Dec 5, 2022, 11:24 PM UTC (0e1aae1)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1250,7 +1250,7 @@ To <dfn noexport>create an attribution trigger</dfn> given a [=string=]
 1. Let |aggregationCoordinator| be null.
 1. If |value|["`aggregation_coordinator_identifier`"] [=map/exists=]:
     1. If |value|["`aggregation_coordinator_identifier`"] is not a [=string=], return null.
-    1. If |value|["`aggregation_coordinator_identifier`"] is not any of [=aggregation coordinator=], return null.
+    1. If |value|["`aggregation_coordinator_identifier`"] is not an [=aggregation coordinator=], return null.
     1. Set |aggregationCoordinator| to |value|["`aggregation_coordinator_identifier`"].
 1. Let |trigger| be a new [=attribution trigger=] with the items:
     : [=attribution trigger/attribution destination=]
@@ -1934,7 +1934,7 @@ of running the following steps:
 To <dfn>obtain the public key for encryption</dfn> given an optional [=aggregation coordinator=]
 |aggregationCoordinator|, asynchronously return a user-agent-determined public key
 or an error in the event that the user agent failed to obtain the public key.
-A user-agent-determined default [=aggregation coordinator=] may be used if
+A user-agent-determined default [=aggregation coordinator=] should be used if
 |aggregationCoordinator| is null.
 
 Note: The user agent might enforce weekly key rotation. If there are multiple keys, the user agent


index.bs
```diff
@@ -240,6 +240,17 @@ An event-level trigger configuration is a [=struct=] with the following items:
 
 </dl>
 
+<h3 id="aggregation-coordinator-header">Aggregation coordinator</h3>
+
+An <dfn>aggregation coordinator</dfn> is a non-empty [=string=] that specifies
+the depoloyment option for the aggregation service and is one of the following:
```
apasel422, line 246:
```suggestion
the deployment option for the aggregation service and is one of the following:
```

index.bs
```diff
@@ -957,6 +972,8 @@ a [=trigger state=] |triggerState|:
     :: null
     : [=attribution trigger/debug reporting enabled=]
     :: false
+    : [=attribution trigger/aggregation coordinator=]
+    :: "<code>[=aggregation coordinator/aws-cloud=]</code>"
```
apasel422, line 976: Shouldn't this be an implementation-determined default?

apasel422 (dismissed):

index.bs
linnan-github: Updated to use null as default and implementation can determine what to do with it.

index.bs
```diff
@@ -1236,6 +1253,12 @@ To <dfn noexport>create an attribution trigger</dfn> given a [=string=]
 1. Let |debugReportingEnabled| be false.
 1. If |value|["`debug_reporting`"] [=map/exists=] and is a [=boolean=], set
     |debugReportingEnabled| to value["`debug_reporting`"].
+1. Let |aggregationCoordinator| be "<code>[=aggregation coordinator/aws-cloud=]</code>".
```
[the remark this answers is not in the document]
linnan-github: Updated to use null as default and implementation can determine what to do with it.

index.bs
```diff
@@ -240,6 +240,17 @@ An event-level trigger configuration is a [=struct=] with the following items:
...
 </dl>
 
+<h3 id="aggregation-coordinator-header">Aggregation coordinator</h3>
+
+An <dfn>aggregation coordinator</dfn> is a non-empty [=string=] that specifies
+the depoloyment option for the aggregation service and is one of the following:
+
+<dl dfn-for="aggregation coordinator">
+: "<dfn><code>aws-cloud</code></dfn>"
+:: The aggregation service is deployed on AWS cloud.
```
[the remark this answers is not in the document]
linnan-github: Yeah I'm also hesitant to call out all possible values in the spec. Updated to leave it opaque and user agent determined.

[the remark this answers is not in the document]
linnan-github: I'm not sure what's the best link to add here, probably https://aws.amazon.com/?

index.bs
```diff
@@ -270,6 +275,8 @@ An attribution trigger is a [=struct=] with the following items:
 :: A [=byte sequence=].
 : <dfn>debug reporting enabled</dfn>
 :: A [=boolean=].
+: <dfn>aggregation coordinator</dfn>
+:: Null or an [=aggregation coordinator=].
```
apasel422, line 279: What's the benefit of null vs an implementation-defined default?

index.bs
```diff
@@ -1236,6 +1247,12 @@ To <dfn noexport>create an attribution trigger</dfn> given a [=string=]
 1. Let |debugReportingEnabled| be false.
 1. If |value|["`debug_reporting`"] [=map/exists=] and is a [=boolean=], set
     |debugReportingEnabled| to value["`debug_reporting`"].
+1. Let |aggregationCoordinator| be null.
+1. If |value|["`aggregation_coordinator_identifier`"] [=map/exists=]:
+    1. If |value|["`aggregation_coordinator_identifier`"] is not a [=string=], return null.
+    1. If |value|["`aggregation_coordinator_identifier`"] is not an [=ASCII case-insensitive=] match
+        for any of [=aggregation coordinator=], return null.
```
apasel422, line 1254: Are we sure we want to prevent source-registration entirely if the browser doesn't support a particular coordinator?

apasel422 (dismissed):

index.bs
linnan-github: It's trigger-registration. This was brought up in https://chromium-review.googlesource.com/c/chromium/src/+/4067585/comment/efd17f05_99960408 and per discussion, I think it's preferred to error out?

@johnivdel Any thoughts here? Thanks.

index.bs
linnan-github: Not particular, just to avoid defining another implementation-defined default, but I don't feel strongly. Wdyt?

index.bs
linnan-github: If a particular coordinator is not supported by the browser, then even if it's encrypted using the default coordinator, it may not be decrypted properly. In this sense, probably it's fine to drop the trigger registration?

index.bs
```diff
@@ -240,6 +240,11 @@ An event-level trigger configuration is a [=struct=] with the following items:
 
 </dl>
 
+<h3 id="aggregation-coordinator-header">Aggregation coordinator</h3>
+
+An <dfn>aggregation coordinator</dfn> is one of a user-agent-determined [=set=]
```
alexmturner, line 245: nit: s/user-agent-determined/user agent-determined/

index.bs
```diff
@@ -240,6 +240,11 @@ An event-level trigger configuration is a [=struct=] with the following items:
...
+of [=strings=] that specifies the deployment option for the aggregation service.
```
alexmturner, line 246: optional: "that specifies which aggregation service deployment to use."

index.bs
alexmturner: I think I'd slightly lean towards not null (but I don't feel strongly). That being said, if we keep null, it probably makes sense to explicitly call out that null implies use a default in "obtain the public key for encryption".

index.bs
```diff
@@ -1236,6 +1247,12 @@ To <dfn noexport>create an attribution trigger</dfn> given a [=string=]
 1. Let |debugReportingEnabled| be false.
 1. If |value|["`debug_reporting`"] [=map/exists=] and is a [=boolean=], set
     |debugReportingEnabled| to value["`debug_reporting`"].
+1. Let |aggregationCoordinator| be null.
+1. If |value|["`aggregation_coordinator_identifier`"] [=map/exists=]:
+    1. If |value|["`aggregation_coordinator_identifier`"] is not a [=string=], return null.
+    1. If |value|["`aggregation_coordinator_identifier`"] is not an [=ASCII case-insensitive=] match
```
alexmturner, line 1253: Do want to bring up the case sensitivity question again. It seems that HTML attributes are always case insensitive and javascript enums are always case sensitive. Do we know if there's anything more comparable to this situation to model off of?

One thing that maybe leans towards case sensitivity is that we do exact matches for the JSON keys, right? So maybe it also makes sense to match values exactly.

alexmturner (commented):
Generally looks good, minor things only

index.bs
linnan-github: We use `user-agent-determined` in public key section, so will leave as is for consistency.

index.bs
linnan-github: Updated to call out that a default may be used if it is null. Please let me know if anyone feels strongly to use not null. Thanks.

index.bs
linnan-github: @johnivdel @csharrison for thoughts, thanks!

index.bs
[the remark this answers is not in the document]
linnan-github: Not needed anymore.

index.bs
linnan-github: Updated to do a case-sensitive match, thanks!

alexmturner (approved):
LGTM w two nits :)

index.bs
```diff
@@ -1236,6 +1247,11 @@ To <dfn noexport>create an attribution trigger</dfn> given a [=string=]
 1. Let |debugReportingEnabled| be false.
 1. If |value|["`debug_reporting`"] [=map/exists=] and is a [=boolean=], set
     |debugReportingEnabled| to value["`debug_reporting`"].
+1. Let |aggregationCoordinator| be null.
+1. If |value|["`aggregation_coordinator_identifier`"] [=map/exists=]:
+    1. If |value|["`aggregation_coordinator_identifier`"] is not a [=string=], return null.
+    1. If |value|["`aggregation_coordinator_identifier`"] is not any of [=aggregation coordinator=], return null.
```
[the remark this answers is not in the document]

index.bs
```diff
@@ -1911,8 +1931,11 @@ of running the following steps:
 
 <h3 id="obtain-aggregatable-report-aggregation-service-payloads">Obtaining an aggregatable report's aggregation service payloads</h3>
 
-To <dfn>obtain the public key for encryption</dfn>, asynchronously return a user-agent-determined public key
+To <dfn>obtain the public key for encryption</dfn> given an optional [=aggregation coordinator=]
+|aggregationCoordinator|, asynchronously return a user-agent-determined public key
 or an error in the event that the user agent failed to obtain the public key.
+A user-agent-determined default [=aggregation coordinator=] may be used if
```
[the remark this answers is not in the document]
linnan-github: Updated with "should", thanks!

apasel422 (approved):

PR #641, opened 2022-12-02
Update README for header validator

linnan-github:


header-validator/README.md
```diff
@@ -1,7 +1,8 @@
 # Run tests
 
 `npm run test`
-`npm run test-source`
-`npm run test-trigger`
+`npm run test-eligible`
+`npm run test-json`
+`npm run test-os`
```
apasel422, line 6: I think we just need a single command: `npm run test`.

https://github.com/WICG/attribution-reporting-api/blob/6393bc05de80f1132eda1ba244e1330deaa2cc4e/header-validator/package.json#L7

maudnals (approved):

header-validator/README.md
linnan-github: yeah, but probably it doesn't hurt to list all tests available?

apasel422: I don't think it's beneficial to list them redundantly here; as evidenced by this PR itself, it's easy for the doc to become out of date with respect to the actual package.

linnan-github: That's fair, updated with only `npm run test`, thanks.

PR #650, opened 2022-12-12
Fix unexpired destinations limit return to handle new algorithm

ThomasQuesadilla:
The algorithm changes for checking unexpired destination limit from #601 includes appending the new source's destinations with the set holding the currently stored destinations when checking if the size has surpassed the max destinations per source site. Previously, the new source's destination was never appended to this set, so it made sense to have the set's size check be greater than or equal to the max limit, but now it makes more sense to make the condition just greater than.

***
<a href="https://pr-preview.s3.amazonaws.com/WICG/attribution-reporting-api/pull/650.html" title="Last updated on Dec 15, 2022, 1:35 PM UTC (0259ad7)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/attribution-reporting-api/650/14b657c...0259ad7.html" title="Last updated on Dec 15, 2022, 1:35 PM UTC (0259ad7)">Diff</a>


csharrison (approved):

csharrison:
I would avoid using the term "implementation" to refer to a spec algorithm. Can you reword the subject / description to use something like "algorithm" instead?

PR #651, opened 2022-12-12
Update the explainer for Accept-Attribution-Reporting-Support

### LINKED ISSUE #552: Attribution-Reporting-Support for subresource trigger registration
The event explainer allows all subresources on a page to to respond with the `Attribution-Reporting-Register-Trigger` header, even without the `Attribution-Reporting-Eligible` header present on the request.

The [cross app and web explainer proposes](https://github.com/WICG/attribution-reporting-api/blob/main/app_to_web.md#api-changes) adding a new Attribution-Reporting-Support header on all requests which specify the `Attribution-Reporting-Eligible` header. This means it is not possible for some trigger requests to have access to the support header when deciding how to register.

Ideally there could be a mechanism that allows these generic subresource requests to gain access to the header.

One idea would be to allow a subresource request to respond with a new `Accept-Attribution-Reporting` header that indicates the support header should be added to subsequent redirects (somewhat similar to Accept-CH header).

```mermaid
sequenceDiagram
    Browser-->>adtech.example: https://adtech.example/register-conversion
    adtech.example-->>Browser: Location: register-conversion-redirect<br/>Accept-Attribution-Reporting:support
    Browser-->>adtech.example: https://adtech.example/register-conversion-redirect<br/>Attribution-Reporting-Support: os,web
     adtech.example-->>Browser: Attribution-Reporting-Register-OS-Trigger: ...
```

This allows the adtech to gain access to the support header, at the cost of an additional redirect. To avoid an additional redirect, the adtech may always add the eligibility header to the request registering the trigger.

For registration redirect chains, this also only requires one party in the chain to return the `Accept-Attribution-Reporting` header.

An alternative would be to add the header on all subresource requests, but exposing this information on all requests seems wasteful and increases fingerprinting potential as discussed in the [explainer](https://github.com/WICG/attribution-reporting-api/blob/main/app_to_web.md#new-fingerprint-vector).

linnan-github:
Fixes #552

### DIFF
--- a/app_to_web.md
+++ b/app_to_web.md
@@ -55,7 +55,7 @@ the server can optionally respond to the request with a [boolean structured head
 Accept-Attribution-Reporting-Support: ?1
 ```
 
-If this header exists and contains a true value, the browser will add the
+If this header exists and contains a true value, the browser will set the
 `Attribution-Reporting-Support` header in all subsequent requests in the same redirect chain.
 
 If the `Attribution-Reporting-Support` header indicates OS support, the reporting origin can optionally respond to the request with a [string structured header](https://datatracker.ietf.org/doc/html/draft-ietf-httpbis-header-structure-15#section-3.3.3) that indicates a desire to use the OS's attribution API instead of the browser's. Note that the API also allows browsers to only support OS-level attribution if they choose.
--- a/app_to_web.md
+++ b/app_to_web.md
@@ -50,7 +50,7 @@ Attribution-Reporting-Support: os, web
 ```
 
 For subresource requests without the `Attribution-Reporting-Eligible` header,
-the reporting origin can optionally respond to the request with a [boolean structured header](https://datatracker.ietf.org/doc/html/draft-ietf-httpbis-header-structure-15#section-3-3-6):
+the server can optionally respond to the request with a [boolean structured header](https://datatracker.ietf.org/doc/html/draft-ietf-httpbis-header-structure-15#section-3-3-6):
 ```
 Accept-Attribution-Reporting-Support: ?1
 ```
--- a/app_to_web.md
+++ b/app_to_web.md
@@ -50,9 +50,7 @@ Attribution-Reporting-Support: os, web
 ```
 
 For subresource requests without the `Attribution-Reporting-Eligible` header,
-the reporting origin can optionally respond to the request with a [boolean
-structured
-header](https://datatracker.ietf.org/doc/html/draft-ietf-httpbis-header-structure-15#section-3-3-6):
+the reporting origin can optionally respond to the request with a [boolean structured header](https://datatracker.ietf.org/doc/html/draft-ietf-httpbis-header-structure-15#section-3-3-6):
 ```
 Accept-Attribution-Reporting-Support: ?1
 ```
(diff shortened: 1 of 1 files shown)


app_to_web.md
```diff
@@ -48,7 +48,17 @@ The existing API involves sending requests to the reporting origin to register e
...
+For subresource requests without the `Attribution-Reporting-Eligible` header,
+the server can optionally respond to the request with a [boolean structured header](https://datatracker.ietf.org/doc/html/draft-ietf-httpbis-header-structure-15#section-3-3-6):
+```
+Accept-Attribution-Reporting-Support: ?1
+```
+
+If this header exists and is a true value, the browser will add the
+`Attribution-Reporting-Support` header in all subsequent redirect requests.
+
+If the `Attribution-Reporting-Support` header indicates OS support, the reporting origin can optionally respond to the request with a [string structured header](https://datatracker.ietf.org/doc/html/draft-ietf-httpbis-header-structure-15#section-3.3.3) that indicates a desire to use the OS’s attribution API instead of the browser’s. Note that the API also allows browsers to only support OS-level attribution if they choose.
```
apasel422, line 61:
```suggestion
If the `Attribution-Reporting-Support` header indicates OS support, the reporting origin can optionally respond to the request with a [string structured header](https://datatracker.ietf.org/doc/html/draft-ietf-httpbis-header-structure-15#section-3.3.3) that indicates a desire to use the OS's attribution API instead of the browser's. Note that the API also allows browsers to only support OS-level attribution if they choose.
```

apasel422 (changes requested):

apasel422 (approved):

app_to_web.md
```diff
@@ -48,7 +48,17 @@ The existing API involves sending requests to the reporting origin to register e
 ```
 Attribution-Reporting-Support: os, web
 ```
-If this header indicates OS support, the reporting origin can optionally respond to the request with a [string structured header](https://datatracker.ietf.org/doc/html/draft-ietf-httpbis-header-structure-15#section-3.3.3) that indicates a desire to use the OS’s attribution API instead of the browser’s. Note that the API also allows browsers to only support OS-level attribution if they choose.
+
+For subresource requests without the `Attribution-Reporting-Eligible` header,
+the server can optionally respond to the request with a [boolean structured header](https://datatracker.ietf.org/doc/html/draft-ietf-httpbis-header-structure-15#section-3-3-6):
+```http
+Accept-Attribution-Reporting-Support: ?1
```
johnivdel, line 55: What is the browser behavior is false is specified? While this header allows such a thing, I don't think it is particularly useful.

In #552, I tried to go with a design that somewhat aligns to client hints, where you denote what headers you want to receive, although that is potentially a bit heavy handed given we only have one supported header.

Do we have any other examples of APIs which use a boolean in this way?

linnan-github: If false is specified then it’s the header will not be set.

Looks like topics API (Observe-Browsing-Topics header) uses a boolean this way as well (https://github.com/patcg-individual-drafts/topics#specific-details).

linnan-github: Agree that it’s more flexible to have a generic Accept-Attribution-Reporting header that takes a list of headers, but given that currently we only have one header, I feel it’s a bit heavy handed. If we do have other headers to support in the future, we can consider adding one at that time. Wdyt?

johnivdel: I think that sounds reasonable, and it is fairly similar to the topics use case.

johnivdel (approved):
LGTM!

PR #682, opened 2023-01-20
Skip verbose debug report if no event-level trigger data is provided

### LINKED ISSUE #680: Discrepancy between event-level-only and aggregatable-only triggers
For a trigger that only specifies `event_trigger_data` (and not `aggregatable_trigger_data`), no verbose debug report is sent corresponding to the lack of resulting aggregatable report, since the intent of the trigger is event-level only. This corresponds to the early return with null `debugData` in https://wicg.github.io/attribution-reporting-api/#trigger-aggregatable-attribution when the aggregatable trigger data is empty.

But for a trigger that only specifies `aggregatable_trigger_data`, the specification currently mandates that a `trigger-event-no-matching-configurations` verbose debug report is sent, even though the intent of the trigger is aggregatable-only. This is a consequence of `matchedConfig` being null in https://wicg.github.io/attribution-reporting-api/#trigger-event-level-attribution.

This difference in behavior is potentially confusing, and is also a waste of bandwidth. We should treat the two types of reports the same way here, and I propose that we treat "lack of `event_trigger_data`" (missing from the JSON entirely or present as an empty list) as a success condition not resulting in a verbose debug report.

apasel422:
Fixes #680

***
<a href="https://pr-preview.s3.amazonaws.com/apasel422/attribution-reporting-api/pull/682.html" title="Last updated on Jan 20, 2023, 9:10 PM UTC (f684008)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/attribution-reporting-api/682/9214d1e...apasel422:f684008.html" title="Last updated on Jan 20, 2023, 9:10 PM UTC (f684008)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1514,6 +1514,9 @@ To <dfn>trigger event-level attribution</dfn> given an [=attribution trigger=] |
 [=attribution source=] |sourceToAttribute|, an [=attribution rate-limit record=] |rateLimitRecord|
 and a [=boolean=] |topLevelFiltersMatch|, run the following steps:
 
+1. If |trigger|'s [=attribution trigger/event-level trigger configurations=]
+    [=list/is empty=], return the [=triggering result=]
+    ("<code>[=triggering status/dropped=]</code>", null).
 1. If |sourceToAttribute|'s [=attribution source/randomized response=] is not null and is not [=set/is empty|empty=]:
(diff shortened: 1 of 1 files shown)


linnan-github (approved):

csharrison:
This feels fine to me, and I'm happy deferring to Nan.

PR #687, opened 2023-01-23
No aggregation keys in the matching source should be considered as no contributions

### LINKED ISSUE #685: No verbose debug report when no aggregation keys are specified in the matching source
Currently no verbose debug report is sent if no aggregation keys are specified in the matching source, even if aggregatable trigger data is specified in the trigger.

We should probably consider sending a `trigger-aggregate-no-contributions` verbose debug report in this case to debug whether the trigger is attributed properly.

linnan-github:
Fixes #685


apasel422 (approved):
LGTM but just making sure we don't want a separate type for this.

linnan-github:
> LGTM but just making sure we don't want a separate type for this.

I feel we don't need a separate type, this can be easily found out by the debug keys.

apasel422 (approved):

PR #692, opened 2023-01-26
Filter-based aggregatable deduplication

### LINKED ISSUE #664: Allow filter based aggregatable dedup key
Currently the API only supports aggregatable deduplication per trigger based on trigger side data. 

Similar to event-level deduplication, a filter-based aggregatable dedup key may allow deduplication based on cross-site data.

An optional `aggregatable_deduplication_keys` may be added to `Attribution-Reporting-Register-Trigger` header:

```json
{
   ...,
   aggregatable_deduplication_keys: {
     [{
         deduplication_key: "[unsigned 64-bit integer]",
         filters: {},
         not_filters: {}
     }]
  }
}
```
The first matching deduplication key will be selected.

linnan-github:
Fixes #664

***
<a href="https://pr-preview.s3.amazonaws.com/linnan-github/conversion-measurement-api/pull/692.html" title="Last updated on Feb 2, 2023, 11:28 PM UTC (b9f3cad)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/attribution-reporting-api/692/75a7c8e...linnan-github:b9f3cad.html" title="Last updated on Feb 2, 2023, 11:28 PM UTC (b9f3cad)">Diff</a>

### DIFF
--- a/header-validator/validate-json.js
+++ b/header-validator/validate-json.js
@@ -8,6 +8,7 @@ const limits = {
   maxEventTriggerData: 10,
   maxFilters: 50,
   maxFilterValues: 50,
+  maxAggregatableDedupKeys: 10,
 }
 
 class State {
@@ -269,6 +270,16 @@ const aggregationCoordinator = string((state, value) => {
   state.error(`must match '${awsCloud}' (case-sensitive)`)
 })
 
+const aggregatableDedupKeys = list(
+  (state, value) =>
+    state.validate(value, {
+      deduplication_key: optional(uint64),
+      filters: optional(filters()),
+      not_filters: optional(filters()),
+    }),
+  limits.maxAggregatableDedupKeys
+)
+
 export function validateTrigger(trigger) {
   const state = new State()
   state.validate(trigger, {
@@ -280,7 +291,7 @@ export function validateTrigger(trigger) {
     event_trigger_data: optional(eventTriggerData),
     filters: optional(filters()),
     not_filters: optional(filters()),
-    aggregatable_deduplication_key: optional(uint64),
+    aggregatable_deduplication_keys: optional(aggregatableDedupKeys),
   })
   return state.result()
 }
--- a/index.bs
+++ b/index.bs
@@ -204,7 +204,7 @@ An attribution source is a [=struct=] with the following items:
 : <dfn>aggregatable budget consumed</dfn>
 :: A non-negative integer, total [=aggregatable contribution/value=] of all [=aggregatable contributions=] created with this [=attribution source=].
 : <dfn>aggregatable dedup keys</dfn>
-:: [=ordered set=] of [=attribution trigger/aggregatable dedup keys=] associated with this [=attribution source=].
+:: [=ordered set=] of [=aggregatable dedup key/dedup key|aggregatable dedup keys=] associated with this [=attribution source=].
 : <dfn>debug reporting enabled</dfn>
 :: A [=boolean=].
 
@@ -228,6 +228,20 @@ An aggregatable trigger data is a [=struct=] with the following items:
 :: A non-negative 128-bit integer.
 : <dfn>source keys</dfn>
 :: An [=ordered set=] of [=strings=].
+: <dfn>filters</dfn>
+:: A [=filter map=].
+: <dfn>negated filters</dfn>
+:: A [=filter map=].
+
+</dl>
+
+<h3 dfn-type=dfn>Aggregatable dedup key</h3>
+
+An aggregatable dedup key is a [=struct=] with the following items:
+
+<dl dfn-for="aggregatable dedup key">
+: <dfn>dedup key</dfn>
+:: Null or a non-negative 64-bit integer.
 : <dfn>filters</dfn>
 :: A [=filter map=].
 : <dfn>negated filters</dfn>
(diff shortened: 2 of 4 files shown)


header-validator/validate-json.js
```diff
@@ -269,6 +270,16 @@ const aggregationCoordinator = string((state, value) => {
   state.error(`must match '${awsCloud}' (case-sensitive)`)
 })
 
+const aggregatableDedupKeys = list(
+  (state, value) =>
+    state.validate(value, {
+      deduplication_key: optional(uint64),
```
apasel422, line 276: Is there a benefit to making this field optional? Wouldn't we just want to omit the list entry instead?

index.bs
```diff
@@ -235,6 +235,20 @@ An aggregatable trigger data is a [=struct=] with the following items:
 
 </dl>
 
+<h3 dfn-type=dfn>Aggregatable dedup key</h3>
+
+An aggregatable dedup key is a [=struct=] with the following items:
+
+<dl dfn-for="aggregatable dedup key">
+: <dfn>dedup key</dfn>
+:: Null or a non-negative 64-bit integer.
```
apasel422, line 244: What is the benefit of making this nullable?

index.bs
```diff
@@ -204,7 +204,7 @@ An attribution source is a [=struct=] with the following items:
 : <dfn>aggregatable budget consumed</dfn>
 :: A non-negative integer, total [=aggregatable contribution/value=] of all [=aggregatable contributions=] created with this [=attribution source=].
 : <dfn>aggregatable dedup keys</dfn>
-:: [=ordered set=] of [=attribution trigger/aggregatable dedup keys=] associated with this [=attribution source=].
+:: [=ordered set=] of [=aggregatable dedup key/dedup key|aggregatable dedup keys=] associated with this [=attribution source=].
```
apasel422, line 207: This makes it seem like the source needs to retain the filter information for the deduplication keys, but I don't think it does.

index.bs
```diff
@@ -282,8 +296,8 @@ An attribution trigger is a [=struct=] with the following items:
 : <dfn>aggregatable values</dfn>
 :: An [=ordered map=] whose [=map/key|keys=] are [=strings=] and whose
     [=map/value|values=] are non-negative 32-bit integers.
-: <dfn>aggregatable dedup key</dfn>
-:: Null or a non-negative 64-bit integer.
+: <dfn>aggregatable dedup keys</dfn>
+:: A [=set=] of [=aggregatable dedup key=].
```
apasel422, line 300: I think this comment applies to existing problems elsewhere in the spec, but what does it mean for this field to be a set? In other words, do the 3 fields of the `aggregatable dedup key` structure need to be unique for each entry in the set?

If the set aspect is irrelevant, let's make this a list for simplicity.

apasel422 (dismissed):

header-validator/validate-json.js
linnan-github: It may be useful if one want to use two filters and only set the dedup key if the first filter matches.

```json
{
  "aggregatable_deduplication_keys": [
    {
      "filters": {"x": ["y"]}
    },
    {
      "deduplication_key": "123",
     "filters": {"x": ["z"]}
    }
  ]
}
```

apasel422: Wouldn't the behavior be the same if the first list entry were omitted in your example?

linnan-github: Not exactly. If both filters match, the first matched one will be used. In my example it will be null, but if it's omitted it will be "123".

index.bs
linnan-github: Updated to `aggregatable dedup key values`, hopefully it's a bit clearer.

index.bs
linnan-github: Updated to use list, thanks.

linnan-github:
> Just making sure that the same goal can't be achieved by making `deduplication_key` a field of `aggregatable trigger data`, analogous to the same field in `event trigger data`.

I also considered that and technically it's probably achievable. But `aggregatable trigger data` is different from `event trigger data` in that each data in the list is processed separately, whereas only the first matched `event trigger data` (as well as `aggregatable dedup key`) is used. So it may be a little confusing if we make that a field of `aggregatable trigger data`.

johnivdel:
> > Just making sure that the same goal can't be achieved by making `deduplication_key` a field of `aggregatable trigger data`, analogous to the same field in `event trigger data`.
> 
> I also considered that and technically it's probably achievable. But `aggregatable trigger data` is different from `event trigger data` in that each data in the list is processed separately, whereas only the first matched `event trigger data` (as well as `aggregatable dedup key`) is used. So it may be a little confusing if we make that a field of `aggregatable trigger data`.

+1, this should apply across all values / data.

index.bs
```diff
@@ -1645,8 +1693,14 @@ To <dfn>trigger aggregatable attribution</dfn> given an [=attribution trigger=]
         "<code>[=trigger debug data type/trigger-aggregate-no-contributions=]</code>", |trigger|, |sourceToAttribute| and
         [=obtain debug data on trigger registration/report=] set to null.
     1. Return the [=triggering result=] ("<code>[=triggering status/dropped=]</code>", |debugData|).
-1. If |trigger|'s [=attribution trigger/aggregatable dedup key=] is not null and
-    |sourceToAttribute|'s [=attribution source/aggregatable dedup keys=]
+1. Let |matchedDedupKey| be null.
+1. [=list/iterate|For each=] [=aggregatable dedup key=] |aggregatableDedupKey| of |trigger|'s [=attribution trigger/aggregatable dedup keys=]:
+    1. If the result of running [=match an attribution source's filter data against filters and negated filters=]
+        with |sourceToAttribute|, |aggregatableDedupKey|'s [=aggregatable dedup key/filters=], and
+        |aggregatableDedupKey|'s [=aggregatable dedup key/negated filters=] is false, [=iteration/continue=].
```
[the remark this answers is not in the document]
linnan-github: Updated, thanks!

index.bs
linnan-github: We actually null check below to see whether a dedup key exists or not. It's set to null if it's not specified in the JSON or invalid format.

We need the dedup key to be nullable to allow developers to specify that there's no dedup key for a filter.

apasel422 (approved):

johnivdel (approved):

PR #698, opened 2023-02-09
Explanation of the enumeration of the output states

alois-bissuel:
Adding a quite detailed explanation of the enumeration of the output states, including a short intro to the stars and bars method. Slightly verbose, and it could be well shortened if needed.

### DIFF
--- a/EVENT.md
+++ b/EVENT.md
@@ -958,8 +958,14 @@ mechanism.
 Note that the number of all possible output states k in the above design is:
 - 2925 for click sources. This results from a particular "stars and bars"
   counting method which derives `k = (num_reporting_windows * num_trigger_data +
-  num_reports choose num_reports) = (3 * 8 + 3 choose 3)`. TODO: outline why this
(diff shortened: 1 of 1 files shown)


alois-bissuel:
Oups, I had never linked my GitHub account to my W3C one. This is done now, I hope that the checks will pass now.

csharrison:
> Oups, I had never linked my GitHub account to my W3C one. This is done now, I hope that the checks will pass now.

Hm, I tried revalidating the IPR bot and they still think the PR is unacceptable. Can you make sure you also join the WICG? I don't see you [here](https://www.w3.org/community/wicg/participants).

alois-bissuel:
Thanks Charlie, I created a request for joining the WICG, I hope it will be validated soon!

csharrison:
I actually think this technically a "non-substantive" change as we're just fleshing out a non-normative section with a derivation, so we can merge it before your WICG membership goes through.

PR #703, opened 2023-02-15
Default report windows to expiry after expiry is rounded

linnan-github:
Currently the report windows are default to expiry before it is rounded, which may result in report windows greater than expiry.


apasel422 (approved):

apasel422:
The build is currently broken by https://github.com/speced/bikeshed/issues/2471.

PR #711, opened 2023-02-27
Remove concept of source event ID cardinality

apasel422:
It is unused in the implementation and merely complicates things.

***
<a href="https://pr-preview.s3.amazonaws.com/apasel422/attribution-reporting-api/pull/711.html" title="Last updated on Feb 27, 2023, 10:32 PM UTC (dc16561)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/attribution-reporting-api/711/bca95e1...apasel422:dc16561.html" title="Last updated on Feb 27, 2023, 10:32 PM UTC (dc16561)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -512,9 +512,6 @@ However attribution data is inherently cross-site, and operations on storage wou
 
 # Vendor-Specific Values # {#vendor-specific-values}
 
-<dfn>Source event ID cardinality</dfn> is a positive integer that controls the
-maximum value that can be used as an [=attribution source/event id=].
-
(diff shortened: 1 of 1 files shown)


csharrison (commented):
Don't we need to limit this to 64 bits?

apasel422:
> Don't we need to limit this to 64 bits?

The field is already described as a 64-bit integer, and we don't do that for, e.g., [trigger data](https://wicg.github.io/attribution-reporting-api/#parse-event-triggers). We could look into doing that across the board in a separate change.

csharrison (approved):
OK sounds fine to me.

PR #712, opened 2023-02-28
Clarifications in wake of multiple destinations support landing

dmdabbs:
Suggested EVENT explainer tweak for multiple destinations.

### DIFF
--- a/EVENT.md
+++ b/EVENT.md
@@ -185,8 +185,7 @@ header `Attribution-Reporting-Register-Source` of the form:
 ```
 
 - `destination`: an origin whose eTLD+1 is where attribution will be triggered
-for this source. This field can optionally be repeated in a list of some small size
-(e.g. 3-5).
+for this source. The field may also be specified as a list (JSON array) of no more than three elements.
 
 - `source_event_id`: (optional) A string encoding a 64-bit unsigned integer which
 represents the event-level data associated with this source. This will be


dmdabbs:
Related comments/questions:

If you prefer that in the long term all registrations use an array, should the example use a single entry list and the text indicate a single string value is supported for backward compatability with the initial implemenation?
 
A malformed source string silently prevents the registration. Should this result in a `source-unknown-error` source debug data type? That type is referenced in the spec but never generated AFAICT.

Ditto if the value is not a list or the list is empty.

More than three elements also silently prevents the entire registration, if I read the spec correctly. Can't that simply stop at three and accept the registration? If not, should it result in a `source-unknown-error`?

If one of the destination values is malformed that also nixes the entire registration. Should this result in a `source-unknown-error`?

csharrison:
> If you prefer that in the long term all registrations use a list, should the example use a single entry list and the text indicate a string value is supported for backward compatability with the initial implemenation?

I don't feel too strongly about this. I'd expect most usage will have a single destination, and so supporting both seems fairly natural. @apasel422 wdyt?

> A malformed source string (or list member) silently prevents the registration. Should this result in a source-unknown-error source debug data type? That type is referenced in the spec but never generated AFAICT. If not that, since the reason is known, perhaps renaming to source-other-error.
>
> Ditto if the value is not a list or the list is empty.

Currently we don't send any debug reports on parse errors. One reason this is a bit tricky is that we see the debug opt-in during parsing, so we depend on parsing to be successful in order to be confident that the caller wants debug reports enabled. If this is something we feel is critical to add, let's file a separate issue to track since it affects more than just the destination list.

> More than three elements also silently prevents the entire registration, if I read the spec correctly. Can't parsing simply stop at three and accept the registration? If not, should it result in some error?

I'd be cautious about truncating the data vs. failing, as a direct failure is more visible, and truncation might result in hard-to-debug behavior. Still, point taken that failure isn't _that_ visible given this doesn't generate a debug report.

csharrison (approved):

csharrison:
PR LGTM so if you want to continue discussion on some of these items it might be best to hoist to separate issues.

dmdabbs:
Sounds good. Thank you for the feedback.

PR #728, opened 2023-03-17
Add aggregation coordinator identifier to aggregatable report

linnan-github:
This may be used by ad-techs during cloud migration.

***
<a href="https://pr-preview.s3.amazonaws.com/linnan-github/conversion-measurement-api/pull/728.html" title="Last updated on Mar 23, 2023, 6:37 PM UTC (d470398)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/attribution-reporting-api/728/f371279...linnan-github:d470398.html" title="Last updated on Mar 23, 2023, 6:37 PM UTC (d470398)">Diff</a>

### DIFF
--- a/AGGREGATE.md
+++ b/AGGREGATE.md
@@ -237,6 +237,9 @@ The report will be JSON encoded with the following scheme:
     },
   ],
 
+  // the deployment option for the aggregation service.
+  "aggregation_coordinator_identifier": "aws-cloud",
+
   // Optional debugging information (also present in event-level reports),
   // if the cookie `ar_debug` is present.
   "source_debug_key": "[64 bit unsigned integer]",
--- a/AGGREGATE.md
+++ b/AGGREGATE.md
@@ -237,7 +237,7 @@ The report will be JSON encoded with the following scheme:
     },
   ],
 
-  // the deployment option for the aggregation service.
+  // The deployment option for the aggregation service.
   "aggregation_coordinator_identifier": "aws-cloud",
 
   // Optional debugging information (also present in event-level reports),
--- a/index.bs
+++ b/index.bs
@@ -326,7 +326,7 @@ An attribution trigger is a [=struct=] with the following items:
 : <dfn>debug reporting enabled</dfn>
 :: A [=boolean=].
 : <dfn>aggregation coordinator</dfn>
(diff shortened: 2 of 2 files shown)


AGGREGATE.md
```diff
@@ -237,6 +237,9 @@ The report will be JSON encoded with the following scheme:
     },
   ],
 
+  // The deployment option for the aggregation service.
+  "aggregation_coordinator_identifier": "aws-cloud",
```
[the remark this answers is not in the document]
linnan-github: Thanks Alex, I feel it's more clear to use the same field name as in the trigger response. Wdyt?

index.bs
```diff
@@ -636,6 +636,9 @@ delay to deliver an [=aggregatable report=].
 <dfn>Randomized aggregatable report delay</dfn> is a positive [=duration=] that controls the
 random delay to deliver an [=aggregatable report=].
 
+<dfn>Default aggregation coordinator</dfn> is the [=aggregation coordinator=] that controls how to
```
csharrison, line 639: I think it should be fine to just define this constant in the spec vs. leaving it as vendor specified. Otherwise, my preference would be to avoid a default in the first place.

linnan-github: It seems a bit strange if the allowed list is vendor specific but the default is not, so just want to confirm whether we want to define all these constants or just the default.

csharrison: I think it should be OK to pull all these out into constants. If needed in the future we can have an auxiliary list of vendor defined allowed values but I don't think we need to right now.

linnan-github: That makes sense, updated, PTAL, thanks!

AGGREGATE.md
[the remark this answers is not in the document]
alexmturner: Agreed they should be aligned, happy to defer to you whether it makes sense to change both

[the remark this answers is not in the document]
linnan-github: Thanks, as there were a lot of discussions on the field names before, I would keep both names as is.

apasel422 (approved):

index.bs
linnan-github: @csharrison Could you please take another look? Thanks.

PR #729, opened 2023-03-17
Fix a few outdated links in AGGREGATE.md

csharrison:


AGGREGATE.md
```diff
@@ -587,7 +587,8 @@ contribution to any one bucket). Care should be taken to ensure that either:
 * We can support multiple types of contribution bounding for different reporting
   origins without introducing privacy leaks
 
-See [issue 249](https://github.com/WICG/conversion-measurement-api/issues/249)
+See [issue 249](https://github.com/WICG/attribution-reporting
```
linnan-github, line 590: This link seems to be broken, we don't need newline here.

csharrison: Fixed, good catch!

linnan-github (approved):

PR #737, opened 2023-03-28
Add flexible_event_config.md

csharrison:
This document proposes an extension to event-level reports that increases their flexibility.

### DIFF
--- a/flexible_event_config.md
+++ b/flexible_event_config.md
@@ -1,0 +1,303 @@
+# Flexible event-level configurations
+
+_Note: This document describes possible new functionality in the Attribution Reporting API’s event-level reports. It is not necessarily a feature that will ship in the initial iteration of the API, but we are considering it as a largely backwards-compatible fast-follow. While this new functionality is being developed, we still highly encourage testing the existing API functionalities to support core utility and compatibility needs._
+
+The default configuration for event and navigation sources may not be ideal for all use-cases. We can optionally support extended configuration that allows for callers to specify precisely the information they want out of reports, in order to more efficiently extract utility out of the privacy mechanism. The most efficient configuration will differ from use-case to use-case and will depend on a) the parameters of our privacy mechanism and b) the noise level that can be tolerated by the use-case. For instance, this proposal could be used to:
+
+* Vary the frequency of reports by specifying the number of reporting windows — See e.g. [Binary with more frequent reporting](#binary-with-more-frequent-reporting)
+* Vary the trigger data cardinality in a report — See e.g. a config which matches a [default event source](#default-event-sources)
+* Vary the number of attributions per source registration — See e.g. [Reporting trigger counts](#reporting-trigger-counts) 
+* Reduce the amount of total noise by decreasing the above parameters
+* Configure reporting windows rather than using the defaults — See e.g. [Binary with more frequent reporting](#binary-with-more-frequent-reporting)
+
+
+## Goals
+
+In general, the approach here is to more flexibly encode the output of the API for event-level reports:
+
+* Allow flexibly tuning the output space for a particular source, trading off richness of output (e.g. cardinality of `trigger_data`, number of reporting windows, etc.) with noise levels (See issue #99, issue #734, and issue #733)
+* Allow for value-based reporting (via bucketization) in addition to count-based reporting (issue #55)
+* Allow for choosing flexible reporting windows (issue #46 and issue #736)
+* Allow for setting the possible output space for a source based on a _prior probability distribution_ the ad tech has about the types of trigger that are possible, taking advantage of recent research into label differential privacy for [binary classification](https://arxiv.org/abs/2102.06062) and [regression problems](https://arxiv.org/abs/2212.06074).
+
+
+## API changes 
+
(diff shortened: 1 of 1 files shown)


flexible_event_config.md
```diff
@@ -0,0 +1,303 @@
...
+* Vary the number of attributions per source registration — See e.g. [Reporting trigger counts](#reporting-trigger-counts) 
+* Reduce the amount of total noise by decreasing the above parameters
+* Configure reporting windows rather than using the defaults — See e.g. [Binary with more frequent reporting](#binary-with-more-frequent-reporting)
+
+
+## Goals
+
+In general, the approach here is to more flexibly encode the output of the API for event-level reports:
+
+* Allow flexibly tuning the output space for a particular source, trading off richness of output (e.g. cardinality of `trigger_data`, number of reporting windows, etc.) with noise levels (See issue #99, issue #734, and issue #733)
```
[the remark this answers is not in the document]
csharrison: done

flexible_event_config.md
```diff
@@ -0,0 +1,303 @@
...
+  // triggers within the specified event_report_window.
+  // There will be a limit on the number of specs possible to define for a source.
+  "trigger_specs": [{
+    // This spec will only apply to registrations that set one of the given
+    // trigger data values (positive integers) in the list.
+    // trigger_data will still appear in the event-level report.
+    "trigger_data": [<int>, ...]
+
+    // Represents a series of time windows, starting at <start time>.
+    // At the end of each window, a report for this spec will be delivered an
```
arpanah, line 41: Do we want to rephrase this as: a report for this spec will be delivered an hour after the end of each window OR clarify: At the end of each window a report for this spec will be generated that will be delivered an hour later?

flexible_event_config.md
avvarad, line 1: LGTM

flexible_event_config.md
csharrison: Ah great call out. This is actually really badly worded. I used your first suggestion with a minor clarification that multiple reports can be generated for each window.

flexible_event_config.md
```diff
@@ -0,0 +1,303 @@
...
+}
+```
+
+
+## Privacy considerations
+
+The goal of this proposal is to remain largely privacy neutral with respect to the existing event-level reports. We will publish an algorithm which computes the number of output states for a given source registration. From this we will be able to:
+
+* Compute a randomized response algorithm across the entire output space
+* Set the noise level to satisfy a certain epsilon level level via a randomized response mechanism
```
arpanah, line 135: Remove level from ' epsilon level level '

flexible_event_config.md
```diff
@@ -0,0 +1,303 @@
...
+* Verify that the privacy parameters (like information gain) are within a given threshold, and fail registration if they are not
+
+With these pieces we can ensure that this extension does not regress our privacy measures. Additionally, it allows callers to fine-tune the noise added to the API e.g. by specifying different kinds of output domains. For example, `navigation` sources that only need 1 bit of trigger data and 1 reporting window can use the same noise level as `event` sources.
+
+Beyond setting noise levels, we will have some parameter limits to avoid large computation costs and avoid configurations with too many output states (where noise will increase considerably). Here is an example set of restrictions (feedback always welcome):
+
+* Maximum of 20 total reports, globally and per `trigger_data`
+* Maximum of 5 possible reporting windows per `trigger_data`
+* Maximum of 32 for trigger data cardinality
```
arpanah, line 144: 'Maximum of 32 for trigger data cardinality' to 'Maximum of 32 trigger data cardinality'

flexible_event_config.md
```diff
@@ -0,0 +1,303 @@
...
+    "event_report_windows": {
+      "end_times": [<7 days>]
+    },
+    "summary_window_operator": "count",
+    "summary_buckets": [1, 2, 3, 4, 5, 6, 7, 8, 9, 10]
+  }],
+}
+```
+
+Attributed triggers with `trigger_data` set to 0 get counted capped to 10. The trigger value is ignored since `summary_window_operator` is set to “count”. Supposing 4 triggers are registered and attributed to the source, the report would look like this:
```
arpanah, line 242: Clarify:  `trigger_data` set to 0 get **counted capped** to 10

arpanah (changes requested):
left some general comments. LGTM

flexible_event_config.md
csharrison: done

flexible_event_config.md
csharrison: done

csharrison:
Thanks all! Landing now

flexible_event_config.md
csharrison: Oh actually I just realized I missed this comment! I'm a bit confused though since it seems like what you wrote is exactly the same as what's in the doc. Am I missing something?

arpanah: I didn't quite understand it so left a comment to may be clarify the sentence.  Maybe something like - Attributed triggers with `trigger_data` set to 0 are counted and capped at 10 ?

csharrison: Ah got it. Will make that change!

PR #749, opened 2023-04-04
Update explainer for reducing delays for aggregatable reports

### LINKED ISSUE #439: Hide the true number of aggregatable reports

### LINKED ISSUE #738: Reducing delays for aggregatable reports
Currently, aggregatable reports in the API come with a randomized delay of (10 mins, 1 hour), though often due to offline users or inactive browsers the delay can be much larger. This delay serves to protect whether any particular user had an attributed conversion, which is cross site data. However, if aggregatable reports leaked less cross-site data, they could be delivered with reduced delay.

**In this issue I propose reducing the delay in the API to something like ~0-10 minutes instead of ~10-60 minutes.**
 
We can do this by taking ideas from [issue 439](https://github.com/WICG/attribution-reporting-api/issues/439) and introducing null reports for some fraction trigger registrations, to reduce the total amount of cross-site information embedded in a report. Currently, as a lot of the cross site information is embedded in the existing `source_registration_time` field, we are also thinking of making this field optional as part of this change. E.g. in the trigger registration JSON:
```
{
  aggregatable_source_registration_time: "omit" // or "include"
  ...
}
```

If the `source_registration_time` field is present, the null report rate will need be higher. Currently we are thinking of something like ~0.05 reports (in expectation) for triggers which don’t specify the source registration time, and ~0.25 reports if they do.

Note: ideally this level of configuration would be done globally at the reporting origin level rather than the trigger level, but that would require introducing global configuration to the API across users. This is something we might explore in the future, so any feedback where we’d want to _vary_ this across an ad-tech / reporting origin would be useful.

We also think that in the future, embedding the `source_registration_time` inside the encrypted payload would improve the situation here and reduce the amount of null reports. For this reason we think defaulting to omitting this field makes sense.

linnan-github:
Fixes #738 
Fixes #439

### DIFF
--- a/AGGREGATE.md
+++ b/AGGREGATE.md
@@ -413,7 +413,7 @@ registrations. Null reports will not check or affect rate limits.
 As a lot of the cross site information is embedded in the source registration
 time, trigger registration will accept an optional field
 `aggregatable_source_registration_time` to allow developers to specify whether
-to omit or include `source_registration_time` field within `shared_info` in the
+to exclude or include `source_registration_time` field within `shared_info` in the
 aggregatable report. The null report rate will need to be higher if
 `source_registration_time` is present in the aggregatable report.
 
@@ -422,12 +422,12 @@ aggregatable report. The null report rate will need to be higher if
    ...,
    "aggregatable_trigger_data": ...,
    "aggregatable_values": ...,
-   "aggregatable_source_registration_time": "omit" // or "include", defaults to "omit" if not present
+   "aggregatable_source_registration_time": "exclude" // or "include", defaults to "exclude" if not present
 }
 ```
 
 Strawman: There should be ~0.05 reports (in expectation) sent for trigger
-registrations that omit `source_registration_time` field, and ~0.25 reports for
+registrations that exclude `source_registration_time` field, and ~0.25 reports for
 those that include this field.
 
 ## Data processing through a Secure Aggregation Service
--- a/AGGREGATE.md
+++ b/AGGREGATE.md
@@ -16,13 +16,13 @@
     - [Verbose debugging reports](#verbose-debugging-reports)
   - [Contribution bounding and budgeting](#contribution-bounding-and-budgeting)
   - [Storage limits](#storage-limits)
+  - [Hide the true number of attribution reports](#hide-the-true-number-of-attribution-reports)
 - [Data processing through a Secure Aggregation Service](#data-processing-through-a-secure-aggregation-service)
 - [Privacy considerations](#privacy-considerations)
   - [Differential Privacy](#differential-privacy)
 - [Ideas for future iteration](#ideas-for-future-iteration)
   - [Worklet-based aggregation key generation](#worklet-based-aggregation-key-generation)
   - [Custom attribution models](#custom-attribution-models)
-  - [Hide the true number of attribution reports](#hide-the-true-number-of-attribution-reports)
   - [More advanced contribution bounding](#more-advanced-contribution-bounding)
   - [Choosing among aggregation services](#choosing-among-aggregation-services)
 - [Considered alternatives](#considered-alternatives)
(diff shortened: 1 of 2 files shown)


csharrison (approved):

AGGREGATE.md
```diff
@@ -399,6 +398,38 @@ destination site.
...
+{
+   ...,
+   "aggregatable_trigger_data": ...,
+   "aggregatable_values": ...,
+   "aggregatable_source_registration_time": "omit" // or "include", defaults to "omit" if not present
+}
+```
+
+Strawman: There should be ~0.05 reports (in expectation) sent for trigger
+registrations that omit `source_registration_time` field, and ~0.25 reports for
```
apasel422, line 430:
```suggestion
registrations that omit the `source_registration_time` field, and ~0.25 reports for
```

AGGREGATE.md
```diff
@@ -399,6 +398,38 @@ destination site.
...
+to omit or include `source_registration_time` field within `shared_info` in the
+aggregatable report. The null report rate will need to be higher if
+`source_registration_time` is present in the aggregatable report.
+
+```jsonc
+{
+   ...,
+   "aggregatable_trigger_data": ...,
+   "aggregatable_values": ...,
+   "aggregatable_source_registration_time": "omit" // or "include", defaults to "omit" if not present
```
apasel422, line 425: Do we think we'll actually have other options here in addition to `omit` and `include`? If not, should we use a boolean field instead?

report_verification.md
```diff
@@ -151,8 +151,9 @@ registration](https://github.com/WICG/attribution-reporting-api/blob/main/AGGREG
 For reporting origins that are configured to do Private State Token issuance,
 all new requests that have the `Attribution-Reporting-Eligible` request header
 with the `trigger` value will automatically attach a
-`Sec-Attribution-Reporting-Private-State-Token` header with a blinded message
-(which embeds concatenation of the `report_id` and `attribution_destination`).
+`Sec-Attribution-Reporting-Private-State-Token` header with a comma-seperated
```
apasel422, line 154:
```suggestion
`Sec-Attribution-Reporting-Private-State-Token` header with a comma-separated
```

report_verification.md
```diff
@@ -151,8 +151,9 @@ registration](https://github.com/WICG/attribution-reporting-api/blob/main/AGGREG
...
+list of blinded messages (which embeds concatenation of the `report_id` and
```
apasel422, line 155:
```suggestion
list of blinded messages (which embeds the concatenation of the `report_id` and
```

apasel422 (changes requested):

AGGREGATE.md
csharrison: I don't feel too strongly, but I considered a non-boolean enum in case we wanted to extend this to something like "include_coarse" or some variants along those lines which included `source_registration_time` but rounded to a week rather than a day.

apasel422: Using a string is fine, then, but I think it should be `exclude` rather than `omit`.

csharrison: Yep, that's fine.

PR #750, opened 2023-04-05
Spec for introducing null reports and source registration time configuration

### LINKED ISSUE #738: Reducing delays for aggregatable reports

linnan-github:
Fixes #738

***
<a href="https://pr-preview.s3.amazonaws.com/linnan-github/conversion-measurement-api/pull/750.html" title="Last updated on Apr 18, 2023, 8:36 PM UTC (ee06ea5)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/attribution-reporting-api/750/733c2f5...linnan-github:ee06ea5.html" title="Last updated on Apr 18, 2023, 8:36 PM UTC (ee06ea5)">Diff</a>

### DIFF
--- a/header-validator/data.trigger.js
+++ b/header-validator/data.trigger.js
@@ -12,6 +12,7 @@ export const testCases = [
         "filters": {"x": []},
         "not_filters": {"y": []}
       }],
+      "aggregatable_source_registration_time": "include",
       "aggregatable_trigger_data": [{
         "filters": {"a": ["b"]},
         "key_piece": "0x1",
@@ -525,6 +526,23 @@ export const testCases = [
     expectedErrors: [{
       path: ["aggregation_coordinator_identifier"],
       msg: "must match 'aws-cloud' (case-sensitive)",
+    }],
+  },
+
+  {
+    name: "aggregatable-source-registrationt-time-wrong-type",
+    json: `{"aggregatable_source_registration_time": 1}`,
+    expectedErrors: [{
+      path: ["aggregatable_source_registration_time"],
+      msg: "must be a string",
+    }],
+  },
+  {
+    name: "aggregatable-source-registration-time-unknown-value",
+    json: `{"aggregatable_source_registration_time": "OMIT"}`,
+    expectedErrors: [{
+      path: ["aggregatable_source_registration_time"],
+      msg: "must match 'omit' or 'include' (case-sensitive)",
     }],
   },
 ];
--- a/header-validator/data.trigger.js
+++ b/header-validator/data.trigger.js
@@ -539,10 +539,10 @@ export const testCases = [
   },
   {
     name: "aggregatable-source-registration-time-unknown-value",
-    json: `{"aggregatable_source_registration_time": "OMIT"}`,
+    json: `{"aggregatable_source_registration_time": "EXCLUDE"}`,
     expectedErrors: [{
       path: ["aggregatable_source_registration_time"],
-      msg: "must match 'omit' or 'include' (case-sensitive)",
+      msg: "must match 'exclude' or 'include' (case-sensitive)",
     }],
   },
 ];
--- a/header-validator/data.trigger.js
+++ b/header-validator/data.trigger.js
@@ -530,7 +530,7 @@ export const testCases = [
   },
 
   {
-    name: "aggregatable-source-registrationt-time-wrong-type",
+    name: "aggregatable-source-registration-time-wrong-type",
     json: `{"aggregatable_source_registration_time": 1}`,
     expectedErrors: [{
       path: ["aggregatable_source_registration_time"],
--- a/index.bs
+++ b/index.bs
@@ -1824,15 +1824,6 @@ To <dfn>check if an [=attribution source=] can create [=aggregatable contributio
     |remainingAggregatableBudget|, return false.
 1. Return true.
 
-<h3 id="does-trigger-contain-aggregatable-data">Does trigger contain aggregatable data</h3>
-
-To <dfn>check if an [=attribution trigger=] contains aggregatable data</dfn> given an [=attribution trigger=] |trigger|,
-run the following steps:
-
-1. If |trigger|'s [=attribution trigger/aggregatable trigger data=] is not [=list/is empty|empty=], return true.
-1. If |trigger|'s [=attribution trigger/aggregatable values=] is not [=map/is empty|empty=], return true.
-1. Return false.
-
 <h3 id="obtaining-trigger-debug-data">Obtaining debug data on trigger registration</h3>
 
 To <dfn>obtain debug data body on trigger registration</dfn> given a
@@ -2085,6 +2076,13 @@ To <dfn>find matching sources</dfn> given an [=attribution trigger=] |trigger|:
       * |a|'s [=attribution source/priority=] is equal to |b|'s [=attribution source/priority=] and |a|'s
          [=attribution source/source time=] is less than |b|'s [=attribution source/source time=].
 1. Return |matchingSources|.
+
+To <dfn>check if an [=attribution trigger=] contains aggregatable data</dfn> given an [=attribution trigger=] |trigger|,
+run the following steps:
+
+1. If |trigger|'s [=attribution trigger/aggregatable trigger data=] is not [=list/is empty|empty=], return true.
+1. If |trigger|'s [=attribution trigger/aggregatable values=] is not [=map/is empty|empty=], return true.
+1. Return false.
 
 To <dfn noexport>trigger attribution</dfn> given an [=attribution trigger=] |trigger|, run the following steps:
 
--- a/index.bs
+++ b/index.bs
@@ -2365,10 +2365,9 @@ an [=attribution trigger=] |trigger|:
 1. If |reports| [=list/is empty=], return.
 1. Let |privateStateTokens| be |trigger|'s [=attribution trigger/serialized private state tokens=].
 1. If |privateStateTokens| [=list/is empty=], return.
-1. Run [=shuffle a list=] with |reports|.
-1. Run [=shuffle a list=] with |privateStateTokens|.
-1. Let |n| be |reports|'s [=list/size=].
-1. If |privateStateTokens|'s [=list/size=] is less than |n|, set |n| to it.
+1. [=shuffle a list|Shuffle=] |reports|.
+1. [=shuffle a list|Shuffle=] |privateStateTokens|.
+1. Let |n| be the minimum of |reports|'s [=list/size=] and |privateStateTokens|'s [=list/size=].
 1. For each integer |i| between 0 (inclusive) and |n| (exclusive):
     1. Set |reports|[i]'s [=aggregatable report/serialized private state token=] to |privateStateTokens|[i].
 
--- a/index.bs
+++ b/index.bs
@@ -411,8 +411,8 @@ of [=strings=] that specifies which aggregation service deployment to use.
 An <dfn>aggregatable source registration time configuration</dfn> is one of the following:
 
 <dl dfn-for="aggregatable source registration time configuration">
-: "<dfn><code>omit</code></dfn>"
-:: "`source_registration_time`" is omitted from an [=aggregatable report=]'s [=aggregatable report/shared info=].
+: "<dfn><code>exclude</code></dfn>"
+:: "`source_registration_time`" is excluded from an [=aggregatable report=]'s [=aggregatable report/shared info=].
 : "<dfn><code>include</code></dfn>"
 :: "`source_registration_time`" is included in an [=aggregatable report=]'s [=aggregatable report/shared info=].
 
@@ -703,10 +703,10 @@ for triggers that are attributed to an [=attribution source=] whose
 <dfn>Randomized response epsilon</dfn> is a non-negative double that controls
 the randomized response probability of an [=attribution source=].
 
-<dfn>Randomized null report distribution parameter omitting source registration time</dfn> is a
+<dfn>Randomized null report distribution parameter excluding source registration time</dfn> is a
 double between 0 and 1 (both inclusive) that controls the randomized number of null reports
 generated for an [=attribution trigger=] whose [attribution trigger/aggregatable source registration time configuration]
-is "<code>[=aggregatable source registration time configuration/omit=]</code>".
+is "<code>[=aggregatable source registration time configuration/exclude=]</code>".
 
 <dfn>Randomized null report distribution parameter including source registration time</dfn> is a
 double between 0 and 1 (both inclusive) that controls the randomized number of null reports
@@ -1303,7 +1303,7 @@ a [=trigger state=] |triggerState|:
     : [=attribution trigger/serialized private state tokens=]
     :: «»
     : [=attribution trigger/aggregatable source registration time configuration=]
-    :: "<code>[=aggregatable source registration time configuration/omit=]</code>"
+    :: "<code>[=aggregatable source registration time configuration/exclude=]</code>"
 1. Let |fakeReport| be the result of running [=obtain an event-level report=] with |source|,
     |fakeTrigger|, and |fakeConfig|.
 1. Set |fakeReport|'s [=event-level report/report time=] to the result of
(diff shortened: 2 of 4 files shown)


header-validator/data.trigger.js
```diff
@@ -527,4 +528,21 @@ export const testCases = [
       msg: "must match 'aws-cloud' (case-sensitive)",
     }],
   },
+
+  {
+    name: "aggregatable-source-registrationt-time-wrong-type",
```
apasel422, line 533:
```suggestion
    name: "aggregatable-source-registration-time-wrong-type",
```

index.bs
```diff
@@ -1982,7 +2041,7 @@ To <dfn noexport>trigger attribution</dfn> given an [=attribution trigger=] |tri
 1. Let |eventLevelResult| be the result of running [=trigger event-level attribution=]
     with |trigger|, |sourceToAttribute|, and |rateLimitRecord|.
 1. Let |aggregatableResult| be the result of running [=trigger aggregatable attribution=]
-    with |trigger|, |sourceToAttribute|, and |rateLimitRecord|.
+    with |trigger|, |sourceToAttribute|, |rateLimitRecord| and |nullReports|.
```
apasel422, line 2044:
```suggestion
    with |trigger|, |sourceToAttribute|, |rateLimitRecord|, and |nullReports|.
```

index.bs
```diff
@@ -1993,6 +2052,8 @@ To <dfn noexport>trigger attribution</dfn> given an [=attribution trigger=] |tri
         then [=list/append=] |aggregatableDebugData| to |debugDataList|.
 1. If |debugDataList| is not [=list/is empty|empty=], then run [=obtain and deliver a debug report=]
     with |debugDataList| and |trigger|'s [=attribution trigger/reporting endpoint=].
+1. If |aggregatableResult|'s [=triggering result/status=] is "<code>[=triggering status/dropped=]</code>",
+    then run [=assign private state tokens=] with |nullReports| and |trigger|.
```
apasel422, line 2056:
```suggestion
    run [=assign private state tokens=] with |nullReports| and |trigger|.
```

apasel422 (changes requested):

index.bs
```diff
@@ -2129,11 +2189,103 @@ an [=attribution trigger=] |trigger|:
...
+    : [=aggregatable report/source registration time configuration=]
+    :: |trigger|'s [=attribution trigger/aggregatable source registration time configuration=].
+1. Return |report|.
+
+<h3 id="generating-randomized-null-reports">Generating randomized null reports</h3>
+
+To <dfn>obtain a randomized number of null reports</dfn> given a double |distributionParameter|:
+
+1. [=Assert=]: |distributionParameter| is between 0 and 1 ( both inclusive).
+1. Return a random non-negative integer from a geometric distribution with probability parameter
```
csharrison, line 2204: Nit: we should probably specify how to sample from this (at the very least we should say the PMF is `p(1-p)^k`. Feel free to do this in a follow-up.
```suggestion
1. Return a random non-negative integer sampled from the geometric distribution with probability parameter
```

csharrison: Also maybe add an issue about possibly truncating this, although I think that might entail having an extra boolean to this algorithm that says whether the nulls will be accompanied by a real, so we can truncate the sum of real + fakes.

index.bs
```diff
@@ -2129,11 +2189,103 @@ an [=attribution trigger=] |trigger|:
...
+    |distributionParameter|.
+
+To <dfn>obtain a null report</dfn> given an [=attribution trigger=] |trigger| and a [=moment=] |sourceTime|:
```
csharrison, line 2207: Shall we assert that `|sourceTime|` is properly rounded?

index.bs
```diff
@@ -2129,11 +2189,103 @@ an [=attribution trigger=] |trigger|:
...
+To <dfn>shuffle a [=list=]</dfn> |list|, reorder |list|'s elements such that each possible permutation has equal
+probability of appearance.
+
+To <dfn>assign private state tokens</dfn> given a [=list=] of [=aggregatable reports=] |reports| and
+an [=attribution trigger=] |trigger|:
+
+1. If |reports| [=list/is empty=], return.
+1. Let |privateStateTokens| be |trigger|'s [=attribution trigger/serialized private state tokens=].
+1. If |privateStateTokens| [=list/is empty=], return.
+1. Run [=shuffle a list=] with |reports|.
```
csharrison, line 2282:
```suggestion
1. [=shuffle a list|Shuffle=] |reports|.
```

index.bs
```diff
@@ -2129,11 +2189,103 @@ an [=attribution trigger=] |trigger|:
...
+1. Run [=shuffle a list=] with |privateStateTokens|.
```
csharrison, line 2283:
```suggestion
1. [=shuffle a list|Shuffle=] |privateStateTokens|.
```

index.bs
```diff
@@ -2129,11 +2189,103 @@ an [=attribution trigger=] |trigger|:
...
+1. Let |n| be |reports|'s [=list/size=].
+1. If |privateStateTokens|'s [=list/size=] is less than |n|, set |n| to it.
```
csharrison, line 2285:
```suggestion
1. Let |n| be the minimum of |reports|'s [=list/size=] and |privateStateTokens|'s [=list/size=].
```

index.bs
```diff
@@ -2129,11 +2189,103 @@ an [=attribution trigger=] |trigger|:
...
+        1. Add |nullReport| to the [=aggregatable report cache=].
+        1. [=list/Append=] |nullReport| to |nullReports|.
+1. Return |nullReports|.
+
+<h3 id="assign-private-state-tokens-for-aggregatable-reports">Assigning private state tokens for aggregatable reports</h3>
+
+To <dfn>shuffle a [=list=]</dfn> |list|, reorder |list|'s elements such that each possible permutation has equal
+probability of appearance.
+
+To <dfn>assign private state tokens</dfn> given a [=list=] of [=aggregatable reports=] |reports| and
```
csharrison, line 2276: This algorithm also has to assign the proper `report id` to the null reports, otherwise the token will have a mismatched ID.

Given that we don't currently specify how the report ID gets in the token, I think we can just punt this to an ISSUE, but we should document the problem.

index.bs
linnan-github: Do we want to round source time? Currently it's just trigger time + i days.

linnan-github: Actually I think it's probably fine as is, the source time will be rounded during serialization. Wdyt?

csharrison: Oh right, that makes more sense :) I think this is fine then.

index.bs
linnan-github: Yeah I left this out as report ID assignment is not spec'ed yet. Added an issue for that.

csharrison (approved):

index.bs
```diff
@@ -1185,6 +1214,10 @@ a [=trigger state=] |triggerState|:
     :: false
     : [=attribution trigger/aggregation coordinator=]
     :: [=default aggregation coordinator=]
+    : [=attribution trigger/serialized private state tokens=]
+    :: «»
+    : [=attribution trigger/aggregatable source registration time configuration=]
+    :: "<code>[=aggregatable source registration time configuration/exclude=]</code>"
```
apasel422, line 1220: To be clear, this value doesn't matter, right? The fake trigger is never used to create aggregatable reports.

index.bs
```diff
@@ -1514,6 +1547,15 @@ and a [=moment=] |triggerTime|:
     1. If |value|["`aggregation_coordinator_identifier`"] is not a [=string=], return null.
     1. If |value|["`aggregation_coordinator_identifier`"] is not an [=aggregation coordinator=], return null.
     1. Set |aggregationCoordinator| to |value|["`aggregation_coordinator_identifier`"].
+1. Let |aggregatableSourceRegTimeConfig| be "<code>[=aggregatable source registration time configuration/exclude=]</code>".
+1. If |value|["`aggregatable_source_registration_time`"] [=map/exists=]:
+    1. If |value|["`aggregatable_source_registration_time`"] is not a [=string=], return null.
+    1. If |value|["`aggregatable_source_registration_time`"] is not an [=aggregatable source registration time configuration=], return null.
```
apasel422, line 1553:
```suggestion
    1. If |value|["`aggregatable_source_registration_time`"] is not an [=aggregatable source registration time configuration=],
        return null.
```

index.bs
```diff
@@ -1693,6 +1737,15 @@ To <dfn>check if an [=attribution source=] can create [=aggregatable contributio
     |remainingAggregatableBudget|, return false.
 1. Return true.
 
+<h3 id="does-trigger-contain-aggregatable-data">Does trigger contain aggregatable data</h3>
```
apasel422, line 1740: I'm not sure this heading is necessary: I don't think this minor algorithm needs to show up in the table of contents, and it already has an ID as part of the `<dfn>`.

index.bs
```diff
@@ -1947,15 +2003,16 @@ To <dfn>find matching sources</dfn> given an [=attribution trigger=] |trigger|:
 
 To <dfn noexport>trigger attribution</dfn> given an [=attribution trigger=] |trigger|, run the following steps:
 
+1. Let |hasAggregatableData| be the result of [=checking if an attribution trigger contains aggregatable data=].
```
apasel422, line 2006:
```suggestion
1. Let |hasAggregatableData| be the result of [=checking if an attribution trigger contains aggregatable data=]
    with |trigger|.
```

index.bs
```diff
@@ -2129,11 +2189,108 @@ an [=attribution trigger=] |trigger|:
...
+    :: |trigger|'s [=attribution trigger/aggregation coordinator=].
+    : [=aggregatable report/source registration time configuration=]
+    :: |trigger|'s [=attribution trigger/aggregatable source registration time configuration=].
+1. Return |report|.
+
+<h3 id="generating-randomized-null-reports">Generating randomized null reports</h3>
+
+To <dfn>obtain a randomized number of null reports</dfn> given a double |distributionParameter|:
+
+1. [=Assert=]: |distributionParameter| is between 0 and 1 ( both inclusive).
```
apasel422, line 2203:
```suggestion
1. [=Assert=]: |distributionParameter| is between 0 and 1 (both inclusive).
```

index.bs
```diff
@@ -2129,11 +2189,108 @@ an [=attribution trigger=] |trigger|:
...
+1. Return a random non-negative integer sampled from the geometric distribution with probability parameter
+    |distributionParameter|.
+
+Issue: Consider specifying how to sample this from the geometric distribution.
+
+Issue: Consider truncating the total number of reports generated for an [=attribution trigger=].
```
apasel422, line 2209:
```suggestion
Issue: Consider limiting the total number of reports generated for an [=attribution trigger=].
```
?

index.bs
```diff
@@ -2129,11 +2189,108 @@ an [=attribution trigger=] |trigger|:
...
+        [=randomized null report distribution parameter excluding source registration time=].
+    1. Repeat the following steps |numNullReports| times:
+        1. Let |nullReport| be the result of [=obtaining a null report=] with |trigger| and
+            |trigger|'s [=attribution trigger/trigger time=].
+        1. Add |nullReport| to the [=aggregatable report cache=].
+        1. [=list/Append=] |nullReport| to |nullReports|.
+    1. Return |nullReports|.
+1. Let |maxSourceExpiry| be [=max source expiry=].
+1. Round |maxSourceExpiry| away from zero to the nearest day (86400 seconds).
+1. For each integer |day| between 0 (inclusive) and the number of days in |maxSourceExpiry| (inclusive):
```
apasel422, line 2265: > the number of days in |maxSourceExpiry|

This is a bit vague. Can we specify how to calculate that?

index.bs
```diff
@@ -2129,11 +2189,108 @@ an [=attribution trigger=] |trigger|:
...
+To <dfn>generate null reports</dfn> given an [=attribution trigger=] |trigger|:
+
+1. Let |nullReports| be a new [=list/is empty|empty=] [=list=].
+1. If |trigger|'s [=attribution trigger/aggregatable source registration time configuration=] is "<code>[=aggregatable source registration time configuration/exclude=]</code>":
+    1. Let |numNullReports| be the result of [=obtaining a randomized number of null reports=] with
+        [=randomized null report distribution parameter excluding source registration time=].
+    1. Repeat the following steps |numNullReports| times:
+        1. Let |nullReport| be the result of [=obtaining a null report=] with |trigger| and
+            |trigger|'s [=attribution trigger/trigger time=].
+        1. Add |nullReport| to the [=aggregatable report cache=].
```
apasel422, line 2260:
```suggestion
        1. [=set/Append=] |nullReport| to the [=aggregatable report cache=].
```

index.bs
```diff
@@ -2129,11 +2189,108 @@ an [=attribution trigger=] |trigger|:
...
+    1. Return |nullReports|.
+1. Let |maxSourceExpiry| be [=max source expiry=].
+1. Round |maxSourceExpiry| away from zero to the nearest day (86400 seconds).
+1. For each integer |day| between 0 (inclusive) and the number of days in |maxSourceExpiry| (inclusive):
+    1. Let |numNullReports| be the result of [=obtaining a randomized number of null reports=] with
+        [=randomized null report distribution parameter including source registration time=].
+    1. Repeat the following steps |numNullReports| times:
+        1. Let |fakeSourceTime| be |trigger|'s [=attribution trigger/trigger time=] - |day| days.
+        1. Let |nullReport| be the result of [=obtaining a null report=] with |trigger| and |fakeSourceTime|.
+        1. Add |nullReport| to the [=aggregatable report cache=].
```
apasel422, line 2271:
```suggestion
        1. [=set/Append=] |nullReport| to the [=aggregatable report cache=].
```

apasel422 (changes requested):

index.bs
linnan-github: Yes, this is irrelevant, the fake trigger is only used to create fake event-level reports.

index.bs
linnan-github: Yeah I thought about that as well, but this algorithm is used by two other algorithms, so I feel it's easier to organize this way. Wdyt?

index.bs
linnan-github: Discussed offline, maxSourceExpiry is already rounded to the nearest day, so it's fine leaving as is.

index.bs
apasel422: To me this algorithm just belongs in the general trigger section.

index.bs
```diff
@@ -2129,11 +2190,116 @@ an [=attribution trigger=] |trigger|:
...
+1. [=shuffle a list|Shuffle=] |reports|.
+1. [=shuffle a list|Shuffle=] |privateStateTokens|.
+1. Let |n| be the minimum of |reports|'s [=list/size=] and |privateStateTokens|'s [=list/size=].
+1. For each integer |i| between 0 (inclusive) and |n| (exclusive):
+    1. Set |reports|[i]'s [=aggregatable report/serialized private state token=] to |privateStateTokens|[i].
+
+Issue: Assign the ID associated with the private state token to [=aggregatable report/report ID=].
+
+To <dfn>generate null reports and assign private state tokens</dfn> given an [=attribution trigger=] |trigger|
+and an optional [=aggregatable report=] <dfn for="generate null reports and assign private state tokens"><var>report</var></dfn>:
```
apasel422, line 2296:
```suggestion
and an optional [=aggregatable report=] <dfn for="generate null reports and assign private state tokens"><var>report</var></dfn> defaulting to null:
```

csharrison:
one-way flipping algorithm looks good. This will also simplify the report verification integration if we can figure out how to move `source_registration_time` into the encrypted payload. Then the issuer will only ever have to issue a single token.

index.bs
linnan-github: Ok, moved the algorithm to the general trigger section, thanks.

apasel422 (approved):

Issue #756, opened 2023-04-12
Consider making Attribution-Reporting-Eligible header usage consistent across HTML tags and request APIs


arpitjainx:
As per the AR API [spec](https://github.com/WICG/attribution-reporting-api/blob/main/EVENT.md#registering-attribution-sources), for `<a>`, `<img>` and `<script>` tags and `window.open()` calls, the `Attribution-Reporting-Eligible` header is added by the browser when the `attributionsrc` attribute/feature is present and the browser considers the source or the trigger to be eligible for API registration. In contrast, for request APIs such as `window.fetch()` and `XMLHttpRequest` the header needs to be manually added by the publisher / ad tech. This doesn't really give an indication of whether the browser actually considers the event to be eligible for AR API registration.

During Chrome's Origin Trial, the presence of the header has been a useful signal for identifying integration issues and other bugs such as the publisher not rendering the correct Origin Trial token and Chrome on Android not considering new tab clicks eligible for registration: [crbug.com/1394437](http://crbug.com/1394437). This header would be useful for identifying similar issues with events registered using the `window.fetch()` and `XMLHttpRequest` APIs if the browser could be instructed to itself add the header when it considers the event to be eligible for AR API registration.

Also to note is that manually adding the `Attribution-Reporting-Eligible` header to these request APIs appears to trigger [CORS preflight requests](https://developer.mozilla.org/en-US/docs/Glossary/Preflight_request) which doubles(?) the incoming QPS and requires additional work from the ad tech to handle them.

apasel422:
Thanks for posting this. @csharrison posted something similar at https://github.com/WICG/attribution-reporting-api/issues/540#issuecomment-1349644201. I agree that the discrepancy between the API surfaces in this regard is non-ideal.

I don't know how hard it would be to add first-class support for Attribution Reporting to `XmlHttpRequest`, but at least for `fetch`, we could add a field to [`RequestInit`](https://fetch.spec.whatwg.org/#requestinit) corresponding to the eligibility, defaulting to `trigger`.

If we removed manual setting of the `Attribution-Reporting-Eligible` header, we could also [consider merging it](https://github.com/WICG/attribution-reporting-api/issues/540#issuecomment-1216687382) with `Attribution-Reporting-Support`: we chose separate headers is because one was user-specified for `fetch` and one was always browser-specified.

jfggit:
[...]

PR #757, opened 2023-04-12
Replace manual fetch/XHR headers with first-class option

apasel422:
#756

***
<a href="https://pr-preview.s3.amazonaws.com/apasel422/attribution-reporting-api/pull/757.html" title="Last updated on Apr 19, 2023, 4:48 PM UTC (dfaf258)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/attribution-reporting-api/757/a82aaf0...apasel422:dfaf258.html" title="Last updated on Apr 19, 2023, 4:48 PM UTC (dfaf258)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -208,23 +208,23 @@ Issue: Use |attributionSrcUrls| with [=make a background attributionsrc request=
 # Fetch monkeypatches # {#fetch-monkeypatches}
 
 <pre class="idl">
-enum AttributionReportingEligibility {
-  "none",
-  "source",
-  "trigger",
-  "source-and-trigger",
+dictionary AttributionReportingRequestOptions {
+  boolean eventSourceEligible = false;
+  boolean triggerEligible = false;
 };
 
 partial dictionary RequestInit {
-  AttributionReportingEligibility? attributionReportingEligibility;
+  AttributionReportingRequestOptions? attributionReporting;
 };
 
 partial interface XMLHttpRequest {
-  attribute AttributionReportingEligibility? attributionReportingEligibility;
+  attribute AttributionReportingRequestOptions? attributionReporting;
 };
 </pre>
 
-Issue: Describe how this field is used.
+If {{AttributionReportingRequestOptions}} is not set, then no
+`Attribution-Reporting-Eligible` header is added to the request, but the
+request is allowed to register triggers.
 
 # Permissions Policy integration # {#permission-policy-integration}
 
--- a/index.bs
+++ b/index.bs
@@ -218,11 +218,13 @@ enum AttributionReportingEligibility {
 partial dictionary RequestInit {
   AttributionReportingEligibility? attributionReportingEligibility;
 };
+
+partial interface XMLHttpRequest {
+  attribute AttributionReportingEligibility? attributionReportingEligibility;
+};
 </pre>
 
 Issue: Describe how this field is used.
-
-Issue: Determine whether/how to expose this on {{XMLHttpRequest}}.
 
 # Permissions Policy integration # {#permission-policy-integration}
 
--- a/index.bs
+++ b/index.bs
@@ -207,7 +207,22 @@ Issue: Use |attributionSrcUrls| with [=make a background attributionsrc request=
 
 # Fetch monkeypatches # {#fetch-monkeypatches}
 
-Issue: Specify monkeypatches for source/trigger registration.
+<pre class="idl">
+enum AttributionReportingEligibility {
+  "none",
+  "source",
+  "trigger",
+  "source-and-trigger",
+};
+
+partial dictionary RequestInit {
+  AttributionReportingEligibility? attributionReportingEligibility;
+};
+</pre>
+
+Issue: Describe how this field is used.
+
+Issue: Determine whether/how to expose this on {{XMLHttpRequest}}.
 
 # Permissions Policy integration # {#permission-policy-integration}
 
--- a/index.bs
+++ b/index.bs
@@ -223,9 +223,18 @@ partial interface XMLHttpRequest {
 };
 </pre>
 
-If {{AttributionReportingRequestOptions}} is not set, then no
-`Attribution-Reporting-Eligible` header is added to the request, but the
-request is allowed to register triggers.
+To <dfn>mark a request for Attribution Reporting eligibility<dfn> given a
+[=header list=] |headers| and an optional {{AttributionReportingRequestOptions}}
+|options|:
+
+1. If |options| is null, return.
+1. Let |dict| be an [=ordered map=].
+1. If |options|'s {{AttributionReportingRequestOptions/eventSourceEligible}} is
+    true, [=map/set=] |dict|["`event-source`"] to true.
+1. If |options|'s {{AttributionReportingRequestOptions/triggerEligible}} is
+    true, [=map/set=] |dict|["`trigger`"] to true.
+1. [=header list/Set a structured field value=] given
+    ("`Attribution-Reporting-Eligible`", |dict|) in |headers|.
 
 # Permissions Policy integration # {#permission-policy-integration}
 
--- a/index.bs
+++ b/index.bs
@@ -209,16 +209,17 @@ Issue: Use |attributionSrcUrls| with [=make a background attributionsrc request=
 
 <pre class="idl">
 dictionary AttributionReportingRequestOptions {
-  boolean eventSourceEligible = false;
-  boolean triggerEligible = false;
+  required boolean eventSourceEligible;
+  required boolean triggerEligible;
 };
 
 partial dictionary RequestInit {
-  AttributionReportingRequestOptions? attributionReporting;
+  AttributionReportingRequestOptions attributionReporting;
 };
 
 partial interface XMLHttpRequest {
-  attribute AttributionReportingRequestOptions? attributionReporting;
+  [RaisesException, SecureContext]
+  undefined setAttributionReporting(AttributionReportingRequestOptions options);
 };
 </pre>
 
--- a/index.bs
+++ b/index.bs
@@ -999,6 +999,8 @@ Issue: Support header-processing on redirects.
 
 Issue: Check for transient activation with "<code>[=eligibility/navigation-source=]</code>".
 
+Issue: Reformulate [=eligibility=] in terms of a fetch parameter.
+
 To <dfn>process an attributionsrc response</dfn> given a [=suitable origin=]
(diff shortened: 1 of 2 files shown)


index.bs
```diff
@@ -135,7 +135,24 @@ Issue: Monkeypatch `window.open()`.
 
 # Fetch monkeypatches # {#fetch-monkeypatches}
 
-Issue: Specify monkeypatches for source/trigger registration.
+<pre class="idl">
+enum AttributionReportingEligibility {
+  "none",
+  "source",
+  "trigger",
+  "source-and-trigger",
```
csharrison, line 143: The existence of `source-and-trigger` makes me feel like an enum is not quite the right data structure to use here. WDYT about a dictionary (bikeshedding welcome):
```
fetch(<url>, {
  attributionReporting: {
    sourceEligibility: 'event',
    triggerEligibility: true,
  }
}
```
This sort of follows the advice here:
https://www.w3.org/TR/design-principles/#string-constants

"If you need to express a state which is a combination of properties, which might be expressed as a bitmask in another language, use a dictionary object instead. This object can be passed around as easily as a single bitmask value."

An added benefit is that an ARA-specific dictionary allows easier extensibility in the future.

apasel422: I'm fine with that, though I'd prefer to use a boolean for `sourceEligibility` and call it `eventSourceEligible`. Otherwise, we have to deal with values other than `event`.

csharrison: Perfect

csharrison (approved):

index.bs
```diff
@@ -135,7 +135,24 @@ Issue: Monkeypatch `window.open()`.
...
+  boolean eventSourceEligible = false;
+  boolean triggerEligible = false;
+};
+
+partial dictionary RequestInit {
+  AttributionReportingRequestOptions? attributionReporting;
+};
+
+partial interface XMLHttpRequest {
+  attribute AttributionReportingRequestOptions? attributionReporting;
```
apasel422, line 149: Apparently dictionaries cannot be used as attributes, so this won't work.

https://webidl.spec.whatwg.org/#idl-attributes

index.bs
```diff
@@ -135,7 +135,24 @@ Issue: Monkeypatch `window.open()`.
...
 
-Issue: Specify monkeypatches for source/trigger registration.
+<pre class="idl">
+dictionary AttributionReportingRequestOptions {
+  boolean eventSourceEligible = false;
+  boolean triggerEligible = false;
+};
+
+partial dictionary RequestInit {
+  AttributionReportingRequestOptions? attributionReporting;
```
apasel422, line 145:
```
Nullable dictionary type is forbidden as a dictionary member type.
```

https://webidl.spec.whatwg.org/#idl-dictionaries

yoavweiss:
Is XHR a strong use case? Can't folks that care about this move to `fetch()`? (I'd have assumed most already use `fetch()`, but I could be wrong)

apasel422:
> Is XHR a strong use case? Can't folks that care about this move to `fetch()`? (I'd have assumed most already use `fetch()`, but I could be wrong)

I don't feel strongly about supporting XHR (though the Chromium implementation does today via manual setting of the relevant header), but omitting it doesn't solve the optional dictionary problem for `fetch`.

Some other ideas:

1. Remove fine-grained configuration, and just have the following, where `null` means that the `Attribution-Reporting-Eligible` header is omitted but triggers are supported, `false` means that the header is set to the empty string and neither sources nor triggers are supported, and `true` means the header is set to `Attribution-Reporting-Eligible: event-source, trigger` and both are supported:
```webidl
partial dictionary RequestInit {
  boolean? attributionReportingEligible;
}
partial interface XMLHttpRequest {
  attribute boolean? attributionReportingEligible;
}
```
This makes `fetch` and `XHR` able to behave equivalently to ``, with the additional ability to disable registration 
entirely, which `` does not support, but obviously doesn't allow opting in to only triggers or only sources with the header set.

2. Retain the enum version from 3d188ab6a0fea01c43f48b5535ea096d7f287f9b, despite it not being as future-proof.
3. Use some kind of hack to work around the inability to have an optional dictionary, like a union type composed of either a unit type or an `AttributionReportingRequestOptions` dictionary.

Any other ideas?

apasel422:
Based on what Trust Tokens is doing, it seems that for `fetch` we can [make the dictionary non-nullable on `RequestInit`](https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/core/fetch/request_init.idl;l=32;drc=71fadf4673bc87f9a9877eb457ae4adb118670e9) but [dynamically check for its presence](https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/core/fetch/request.cc;l=578;drc=9d9593d7fc3fa8fa70dfd4dc7f5c98dc8b2dbc84).

For XHR we can use [a setter](https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/core/xmlhttprequest/xml_http_request.idl;l=63;drc=71fadf4673bc87f9a9877eb457ae4adb118670e9).

csharrison:
> Is XHR a strong use case? Can't folks that care about this move to `fetch()`? (I'd have assumed most already use `fetch()`, but I could be wrong)

We can ask for feedback, but previously when we tried to omit XHR support from Private State Tokens we heard very strong feedback that it would require major refactoring for partners and they were not interested in moving to `fetch` simultaneously with testing the new API. cc @dvorak42. Note that but PST and ARA integrate tightly with existing legacy flows.

jfggit:
A few questions:

1. Just to check that I'm reading it correctly, would basically look like this? (And are both event and trigger properties actually required to be present?):

```
const attributionOptions = {eventSourceEligible: true, triggerEligible: false};
window.fetch('https://view.url', {attributionReporting: attributionOptions});
```
2. Does this imply you could register trigger event sources from fetch? Would there be any kind of requirement like "happens close to a user-driven interaction event" or anything? I could see that being helpful for registering things like third-party click trackers.

3. What would the estimated timeline look like, and what does the switchover look like? Is it just once the newer release goes out then setting the header is broken and the new option is supported? I assume it would be safe for fetch clients to just do both (i.e. set header and set options) in the interim?

apasel422:
> 1. Just to check that I'm reading it correctly, would basically look like this? (And are both event and trigger properties actually required to be present?):
> 
> ```
> const attributionOptions = {eventSourceEligible: true, triggerEligible: false};
> window.fetch('https://view.url', {attributionReporting: attributionOptions});
> ```

Yes, as shown in `EVENT.md`. If you want to register sources via `fetch` then both options need to be specified. 

> 2. Does this imply you could register trigger event sources from fetch? Would there be any kind of requirement like "happens close to a user-driven interaction event" or anything? I could see that being helpful for registering things like third-party click trackers.

You can already register event sources from `fetch` today, by setting the `Attribution-Reporting-Eligible: event-source` request header in the options parameter. If you look at the commit diffs, you'll see the current way of doing this. This change just makes doing so supported in a more first-class way. There's no interaction requirements with this change or with the existing behavior.

> 3. What would the estimated timeline look like, and what does the switchover look like? Is it just once the newer release goes out then setting the header is broken and the new option is supported? I assume it would be safe for fetch clients to just do both (i.e. set header and set options) in the interim?

TBD, but I assume that there will be an atomic switchover, in which case it is safe to do both, though you could also feature-detect this with something like ` if ('attributionReporting' in XMLHttpRequest.prototype)`.

apasel422:
> Re: 1 - oh, looking at EVENT.md now, don't the new examples each only set one of the two properties? Is that right?

As demonstrated in `EVENT.md`, both properties need to be *present* in the dictionary, but they can contain an arbitrary mixture of `true`/`false`.

csharrison:
> This change seems fine to make to me, if a little late in the process. In any case, it would be good to have the API locked down and an estimate of the timing as soon as possible just to minimize additional complications.

Hey @jfggit , I agree w/ you we want to get the API locked down ASAP given that we want to ship this soon. Unfortunately the feedback from issue #756 came quite recently so we are only able to respond to this now.

In general, during Chromium's Origin Trial we want to make sure we can make changes like this in response to feedback where possible; that's the reason we do OTs in the first place to enable faster iteration on APIs. Once we graduate past OT, we will be much more careful about backwards incompatible changes (in fact, at least in Chromium they will require independent approvals via the intent process).

jfggit:
+1, agreed with the comments in #756 about consistency. And actually, per that comment, do we expect the CORS requirement to remain after this change?

apasel422:
> And actually, per that comment, do we expect the CORS requirement to remain after this change?

I'm not completely sure, but I suspect CORS will no longer be necessary.

jfggit:
> I'm not completely sure, but I suspect CORS will no longer be necessary.

Great, it would good to get clarity on this whenever possible. That would be a potentially significant advantage over the status quo.

apasel422:
> Great, it would good to get clarity on this whenever possible. That would be a potentially significant advantage over the status quo.

I will update this PR with information as I get it. We could have prioritized this long ago, but no one has mentioned the CORS concerns since this was first implemented in M105.

index.bs
```diff
@@ -135,7 +135,25 @@ Issue: Monkeypatch `window.open()`.
...
+  AttributionReportingRequestOptions attributionReporting;
+};
+
+partial interface XMLHttpRequest {
+  [RaisesException, SecureContext]
+  undefined setAttributionReporting(AttributionReportingRequestOptions options);
+};
+</pre>
+
+If {{AttributionReportingRequestOptions}} is not set, then no
```
yoavweiss, line 154: This part should be a monkey patch on the Fetching processing model.
e.g. "In [http network or cache fetch](https://fetch.spec.whatwg.org/#http-network-or-cache-fetch) after step 8.14, call "Append attribution reporting", where "append attribution reporting" is an algorithm defined here which adds the header when conditions are met, and flips on a boolean that enabled trigger registration with the response.

Something similar should be done for XHR

apasel422: Yes, we do need to integrate with `fetch` itself, but that's true even without this change.

yoavweiss: Is that something you're interested in doing as a followup? IME, it's better to start out early (and small) and expand from there.

apasel422: I plan to do it, but we also need to monkeypatch navigation (also even without this change), and I don't think doing so should block getting this landed as soon as possible, though I will take a look at least the fetch part now, while we're still waiting for feedback on this.

yoavweiss (approved):
LGTM

PR #773, opened 2023-04-26
Monkeypatch fetch using Attribution Reporting eligibility

apasel422:
***
<a href="https://pr-preview.s3.amazonaws.com/apasel422/attribution-reporting-api/pull/773.html" title="Last updated on May 3, 2023, 2:35 PM UTC (7624ba5)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/attribution-reporting-api/773/1a814ef...apasel422:7624ba5.html" title="Last updated on May 3, 2023, 2:35 PM UTC (7624ba5)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -242,9 +242,9 @@ given an optional {{AttributionReportingRequestOptions}} |options|:
     : (false, false)
     :: Return "<code>[=eligibility/empty=]</code>".
     : (false, true)
-    :: Return "<code>[=eligibility/event-source=]</code>".
-    : (true, false)
     :: Return "<code>[=eligibility/trigger=]</code>".
+    : (true, false)
+    :: Return "<code>[=eligibility/event-source=]</code>".
     : (true, true)
     :: Return "<code>[=eligibility/event-source-or-trigger=]</code>".
 
--- a/index.bs
+++ b/index.bs
@@ -307,6 +307,8 @@ add the step
 
 Issue: Monkeypatch {{XMLHttpRequest}}.
 
+Issue: Check permissions policy.
+
 # Permissions Policy integration # {#permission-policy-integration}
 
 This specification defines a [=policy-controlled feature=] identified by the string "<code><dfn noexport>attribution-reporting</dfn></code>".
--- a/index.bs
+++ b/index.bs
@@ -132,9 +132,10 @@ Issue: Consider allowing the user agent to limit the size of |tokens|.
 Issue: Specify when to make background attributionsrc requests for <{a}>.
 
 Issue: Monkeypatch <{img}> and <{script}> loading so that the presence of an
-{{HTMLAttributionSrcElementUtils/attributionSrc}} attribute causes the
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -223,20 +225,89 @@ partial interface XMLHttpRequest {
...
+
+1. [=header list/Delete=] "`Attribution-Reporting-Eligible`" from |headers|.
+1. [=header list/Delete=] "`Attribution-Reporting-Support`" from |headers|.
+1. If |eligibility| is "<code>[=eligibility/unset=]</code>", return.
 1. Let |dict| be an [=ordered map=].
-1. If |options|'s {{AttributionReportingRequestOptions/eventSourceEligible}} is
-    true, [=map/set=] |dict|["`event-source`"] to true.
-1. If |options|'s {{AttributionReportingRequestOptions/triggerEligible}} is
-    true, [=map/set=] |dict|["`trigger`"] to true.
+1. If |eligibility| is:
```
csharrison, line 260: Love this OCaml-like complex pattern matching 😃

index.bs
```diff
@@ -223,20 +225,89 @@ partial interface XMLHttpRequest {
 };
 </pre>
 
-To <dfn>mark a request for Attribution Reporting eligibility</dfn> given a
-[=header list=] |headers| and an optional {{AttributionReportingRequestOptions}}
-|options|:
+A [=request=] has an associated
+<dfn for=request>Attribution Reporting eligibility</dfn> (an [=eligibility=]).
```
csharrison, line 229: Is the upper casing of "Attribution Reporting" correct style? I don't see that in other request members.

csharrison (approved):

index.bs
apasel422: We use it capitalized throughout the spec, so I'd rather be consistent. IIUC the spec fields are case-insensitive anyway. We can change this later.

csharrison (approved):

index.bs
```diff
@@ -2857,16 +2881,31 @@ To <dfn noexport>get an OS-registration URL from a header list</dfn> given a
 1. If |value| is not a [=string=], return null.
 1. Return the result of running the [=URL parser=] on |value|.
 
+"<code><dfn>Attribution-Reporting-Support</dfn></code>" is a Dictionary
+Structured Header set on a [=request=] that indicates which registrars, if
+any, the corresponding [=response=] can use. Its values are not specified and
```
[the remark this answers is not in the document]
apasel422: They are booleans right now (always `true`), both in the spec and implementation, but we don't necessarily want reporting origins to interpret the values for maximum compatibility with a hypothetical change in which the values do have meaning. If you think that's overkill, then I can specify them, though we may want to consider just using a List Structured Header instead of a Dictionary one whose keys are permanently `true` (and where `false` would have to be given a meaning).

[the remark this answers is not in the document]
csharrison: OK that's fine.

csharrison (approved):

PR #780, opened 2023-05-04
Allow OS registration headers to specify multiple URLs

### LINKED ISSUE #696: Support Additional Redirects beside 302
Currently, to support registration on Android attribution Reporting API, Chrome will call registerWebSource with a list of attributionSourceURI and each URI’s registration on Android Attribution Reporting API can return a list of redirects. However, as per Android attribution reporting API [documentation](https://developer.android.com/design-for-safety/privacy-sandbox/attribution-app-to-web#apps-attribution-source-registration), it does not respect Attribution-Reporting-Redirects if the source is registered via registerWebSource preventing additional ad techs from registration if they are not part of the 302 redirect chain.

4KXZ2nfeyxS8m7r

However, the above flow for web->app is different from app->app flow supported via Android Attribution Reporting API where AdTechs can also provide a list of redirects via “Attribution-Reporting-Redirects” response header for additional registrations and not go via 302 route. This will create a discrepancy in data available to additional AdTechs (B1/C1) for ads registered via registerWebSource as compared to registerSource. We request that Chrome should fix this behavior to be on parity with the Android Attribution reporting API 

Possible Options:
- Chrome can add an additional header in the source registration response as part of which AdTechs can provide an additional list of redirects for which Chrome will call registerWebSource to Android attribution reporting API to support source registration. This way Chrome can validate the permissions on the website before invoking the registerWebSource.

- Chrome may pass the [permissions policy](https://github.com/WICG/attribution-reporting-api/issues/558) for attribution-reporting via registerWebSource which possibly enables Android to support Attribution-Reporting-Redirects.

apasel422:
This does not add any new functionality to the API, as reporting origins could already redirect to another URL under their control in order to perform multiple registrations, but does obviate that bandwidth-wasting anti-pattern.

Fixes #696

***
<a href="https://pr-preview.s3.amazonaws.com/apasel422/attribution-reporting-api/pull/780.html" title="Last updated on May 5, 2023, 1:58 PM UTC (9986dc2)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/attribution-reporting-api/780/8c8e8a0...apasel422:9986dc2.html" title="Last updated on May 5, 2023, 1:58 PM UTC (9986dc2)">Diff</a>

### DIFF
--- a/app_to_web.md
+++ b/app_to_web.md
@@ -54,7 +54,7 @@ background requests will be made and the browser will not set
 `Attribution-Reporting-Eligible` header on `<a>`, `window.open`, `<img>`, or
 `<script>` requests.
 
-If the `Attribution-Reporting-Support` header indicates OS support, the reporting origin can optionally respond to the request with a [list structured header](https://www.rfc-editor.org/rfc/rfc8941.html#name-lists) containing one or more URLs that indicates a desire to use the OS's attribution API instead of the browser's. Note that the API also allows browsers to only support OS-level attribution if they choose.
+If the `Attribution-Reporting-Support` header indicates OS support, the reporting origin can optionally respond to the request with a [list structured header](https://httpwg.org/specs/rfc8941.html#list) containing one or more URLs that indicates a desire to use the OS's attribution API instead of the browser's. Note that the API also allows browsers to only support OS-level attribution if they choose.
 ```http
 // Registers a source against a native OS attribution API
 Attribution-Reporting-Register-OS-Source: "https://adtech.example/register", "https://other-adtech.example/register"
--- a/app_to_web.md
+++ b/app_to_web.md
@@ -54,17 +54,16 @@ background requests will be made and the browser will not set
 `Attribution-Reporting-Eligible` header on `<a>`, `window.open`, `<img>`, or
 `<script>` requests.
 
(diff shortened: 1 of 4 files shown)


app_to_web.md
```diff
@@ -54,17 +54,16 @@ background requests will be made and the browser will not set
 `Attribution-Reporting-Eligible` header on `<a>`, `window.open`, `<img>`, or
 `<script>` requests.
 
-If the `Attribution-Reporting-Support` header indicates OS support, the reporting origin can optionally respond to the request with a [string structured header](https://datatracker.ietf.org/doc/html/draft-ietf-httpbis-header-structure-15#section-3.3.3) that indicates a desire to use the OS's attribution API instead of the browser's. Note that the API also allows browsers to only support OS-level attribution if they choose.
-```
+If the `Attribution-Reporting-Support` header indicates OS support, the reporting origin can optionally respond to the request with a [list structured header](https://httpwg.org/specs/rfc8941.html#list) containing one or more URLs that indicates a desire to use the OS's attribution API instead of the browser's. Note that the API also allows browsers to only support OS-level attribution if they choose.
+```http
```
csharrison, line 58: This kind of makes the `//comments` a bit ugly but I guess it's nbd

csharrison (approved):

app_to_web.md
apasel422: The comments don't add anything beyond the prose, so I've just removed them.

PR #784, opened 2023-05-05
Specify defer trigger attribution logic

agarant:
Goal is to mitigate a race condition where a user clicks on an ad which initiates a source registration and navigates to a page that initiates a trigger registration that matches the source. 

If the trigger registration completes before the source, the attribution would be dropped.

***
<a href="https://pr-preview.s3.amazonaws.com/agarant/attribution-reporting-api/pull/784.html" title="Last updated on May 30, 2023, 2:19 PM UTC (a7967b5)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/attribution-reporting-api/784/9debabb...agarant:a7967b5.html" title="Last updated on May 30, 2023, 2:19 PM UTC (a7967b5)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1267,8 +1267,7 @@ To <dfn>process an attributionsrc response</dfn> given a [=suitable origin=]
                 |destinationSite|, |reportingOrigin|, |privateStateTokens|, and
                 |context|'s [=current wall time=].
             1. If |trigger| is not null:
-                1. [=Maybe defer trigger attribution=] with |trigger|.
-                1. [=Trigger attribution=] with |trigger|.
+                1. [=Maybe defer and then complete trigger attribution=] with |trigger|.
         1. If |osSourceURLs| is not null and the user agent supports OS
             registrations, process |osSourceURLs| according to an
             [=implementation-defined=] algorithm.
@@ -2618,13 +2617,15 @@ and an optional [=aggregatable report=] <dfn for="generate null reports and assi
 
 <h3 id="deferring-trigger-attribution">Deferring trigger attribution</h3>
 
-To <dfn>maybe defer trigger attribution</dfn> given an [=attribution trigger=] |trigger|:
+To <dfn>maybe defer and then complete trigger attribution</dfn> given an [=attribution trigger=] |trigger|
+, run the following steps [=in parallel=]:
 
-1. Let |navigation| be the navigation that landed on the document from which |trigger|'s registration was initiated.
-1. If |navigation| is empty, return.
-1. Let |sources| be source registrations originating from background attributionsrc requests initiated by |navigation|.
+1. Let |navigation| be the navigation that landed on the [=document=] from which |trigger|'s registration was initiated.
+1. If |navigation| is null, return.
+1. Let |sources| be all source registrations originating from background attributionsrc requests initiated by |navigation|.
 1. If |sources| is empty, return.
 1. Wait until all |sources| are [=process an attribution source|processed=].
+1. [=Trigger attribution=] with |trigger|.
 
 Issue: Specify this in terms of <a href="https://html.spec.whatwg.org/multipage/browsing-the-web.html#navigating-across-documents">Navigation</a>
 
--- a/index.bs
+++ b/index.bs
@@ -1267,8 +1267,8 @@ To <dfn>process an attributionsrc response</dfn> given a [=suitable origin=]
                 |destinationSite|, |reportingOrigin|, |privateStateTokens|, and
                 |context|'s [=current wall time=].
             1. If |trigger| is not null:
-                1. [=maybe defer trigger attribution=] with |trigger|.
-                1. [=trigger attribution=] with |trigger|.
+                1. [=Maybe defer trigger attribution=] with |trigger|.
+                1. [=Trigger attribution=] with |trigger|.
         1. If |osSourceURLs| is not null and the user agent supports OS
             registrations, process |osSourceURLs| according to an
             [=implementation-defined=] algorithm.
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -1266,7 +1266,9 @@ To <dfn>process an attributionsrc response</dfn> given a [=suitable origin=]
                 [=create an attribution trigger=] with |triggerHeader|
                 |destinationSite|, |reportingOrigin|, |privateStateTokens|, and
                 |context|'s [=current wall time=].
-            1. If |trigger| is not null, [=trigger attribution=] with |trigger|.
+            1. If |trigger| is not null:
+                1. [=maybe defer trigger attribution=] with |trigger|.
```
apasel422, line 1270:
```suggestion
                1. [=Maybe defer trigger attribution=] with |trigger|.
```

index.bs
```diff
@@ -1266,7 +1266,9 @@ To <dfn>process an attributionsrc response</dfn> given a [=suitable origin=]
...
+                1. [=trigger attribution=] with |trigger|.
```
apasel422, line 1271:
```suggestion
                1. [=Trigger attribution=] with |trigger|.
```

index.bs
```diff
@@ -2614,6 +2616,18 @@ and an optional [=aggregatable report=] <dfn for="generate null reports and assi
     1. [=list/Append=] |report| to |reports|.
 1. Run [=assign private state tokens=] with |reports| and |trigger|.
 
+<h3 id="deferring-trigger-attribution">Deferring trigger attribution</h3>
+
+To <dfn>maybe defer trigger attribution</dfn> given an [=attribution trigger=] |trigger|:
+
+1. Let |navigation| be the navigation that landed on the document from which |trigger|'s registration was initiated.
+1. If |navigation| is empty, Return.
```
apasel422, line 2624:
```suggestion
1. If |navigation| is empty, return.
```

index.bs
```diff
@@ -2614,6 +2616,18 @@ and an optional [=aggregatable report=] <dfn for="generate null reports and assi
...
+1. Let |sources| be all source registrations initiated alongside |navigation|.
+1. If |sources| is empty, Return.
```
apasel422, line 2626:
```suggestion
1. If |sources| is empty, return.
```

index.bs
```diff
@@ -2614,6 +2616,18 @@ and an optional [=aggregatable report=] <dfn for="generate null reports and assi
...
+1. Wait until all |sources| are [=process an attribution source|processed=].
+
+Issue:  Specify this in terms of <a href="https://html.spec.whatwg.org/multipage/browsing-the-web.html#navigating-across-documents">Navigation</a>
```
apasel422, line 2629:
```suggestion
Issue: Specify this in terms of <a href="https://html.spec.whatwg.org/multipage/browsing-the-web.html#navigating-across-documents">Navigation</a>.
```

index.bs
```diff
@@ -2614,6 +2616,18 @@ and an optional [=aggregatable report=] <dfn for="generate null reports and assi
     1. [=list/Append=] |report| to |reports|.
 1. Run [=assign private state tokens=] with |reports| and |trigger|.
 
+<h3 id="deferring-trigger-attribution">Deferring trigger attribution</h3>
+
+To <dfn>maybe defer trigger attribution</dfn> given an [=attribution trigger=] |trigger|:
+
+1. Let |navigation| be the navigation that landed on the document from which |trigger|'s registration was initiated.
+1. If |navigation| is empty, Return.
+1. Let |sources| be all source registrations initiated alongside |navigation|.
```
apasel422, line 2625: Maybe clarify that the source registrations were ones originating from background attributionsrc requests that were initiated by `|navigation|`. IIUC foreground ones don't need to be handled specially here.

apasel422 (changes requested):

index.bs
agarant: Other algorithms defined are lowercased, I'd be inclined to keep it lowercased.

index.bs
```diff
@@ -1266,7 +1266,9 @@ To <dfn>process an attributionsrc response</dfn> given a [=suitable origin=]
                 [=create an attribution trigger=] with |triggerHeader|
                 |destinationSite|, |reportingOrigin|, |privateStateTokens|, and
                 |context|'s [=current wall time=].
-            1. If |trigger| is not null, [=trigger attribution=] with |trigger|.
+            1. If |trigger| is not null:
+                1. [=Maybe defer trigger attribution=] with |trigger|.
```
apasel422, line 1270: We probably need to specify that this needs to be done asynchronously somehow. Otherwise, waiting here will block all other processing, right?

index.bs
```diff
@@ -2614,6 +2616,18 @@ and an optional [=aggregatable report=] <dfn for="generate null reports and assi
     1. [=list/Append=] |report| to |reports|.
 1. Run [=assign private state tokens=] with |reports| and |trigger|.
 
+<h3 id="deferring-trigger-attribution">Deferring trigger attribution</h3>
+
+To <dfn>maybe defer trigger attribution</dfn> given an [=attribution trigger=] |trigger|:
+
+1. Let |navigation| be the navigation that landed on the document from which |trigger|'s registration was initiated.
+1. If |navigation| is empty, return.
```
apasel422, line 2624: Not a big deal since "the navigation" isn't even a concept we can reference, but it might be clearer to use "null" instead of "empty".

index.bs
```diff
@@ -2614,6 +2616,18 @@ and an optional [=aggregatable report=] <dfn for="generate null reports and assi
...
+1. Let |sources| be source registrations originating from background attributionsrc requests initiated by |navigation|.
```
apasel422, line 2625: Maybe:
```suggestion
1. Let |sources| be all ongoing source registrations originating from background attributionsrc requests initiated by |navigation|.
```

apasel422 (changes requested):

index.bs
agarant: Considering the next line specifies how we wait for the sources to be processed, I updated to `all` instead of `all ongoing`.

apasel422 (approved):

PR #799, opened 2023-05-15
Clear attribution caches based on reporting origin

linnan-github:


index.bs
```diff
@@ -3038,10 +3038,8 @@ TODO
 <h3 id="clearing-attribution-storage">Clearing attribution storage</h3>
 
 A user agent's [=attribution caches=] contain data about a user's web activity. When a user agent clears an origin's storage,
```
apasel422, line 3040: Should this say "clears an origin's storage on the user's behalf"? We already have https://wicg.github.io/attribution-reporting-api/#clear-site-data-integration for when it's site-initiated.

linnan-github:
The user data control can be implementation-defined and doesn't have to be in the spec.

apasel422 (approved):

PR #804, opened 2023-05-16
Add JSON schemas for registrations and reports

apasel422:


csharrison (approved):
Mostly RS LGTM, I mostly skimmed / spot checked. LMK if you want a more careful review from both of us.

csharrison (approved):
Mostly RS LGTM, I mostly skimmed / spot checked. LMK if you want a more careful review from both of us.

schema/aggregatable_report.schema.json
```diff
@@ -0,0 +1,53 @@
...
+  "properties": {
+    "aggregation_coordinator_identifier": {
+      "type": "string"
+    },
+    "aggregation_service_payloads": {
+      "type": "array",
+      "items": {
+        "type": "object",
+        "properties": {
+          "debug_cleartext_payload": {
```
linnan-github, line 16: This field is optional, shall we add default null?

schema/source.schema.json
```diff
@@ -0,0 +1,99 @@
...
+  "title": "Source",
+  "description": "An Attribution Reporting source registration",
+  "type": "object",
+  "properties": {
+    "aggregatable_report_window": {
+      "$ref": "#/$defs/signed_duration_in_seconds"
+    },
+    "aggregation_keys": {
+      "type": "object",
+      "additionalProperties": {
```
linnan-github, line 13: It's currently not in the spec, but shall we add the bytes limit in the schema? Similar for filters.

apasel422: It's not in the spec right now, so no.

PR #813, opened 2023-05-17
Replace certain vendor-specific values with normative ones

apasel422:
- Max source expiry
- Max entries per filter data
- Max values per filter data entry
- Attribution rate-limit window


csharrison (approved):
Arguably you could also include:
- Max aggregation keys per attribution
- Allowed aggregatable budget per source

PR #817, opened 2023-05-18
Set report requests to be 'same-origin'

### LINKED ISSUE #546: Create a report request causes fetch to blow up (maybe)?

csharrison:
This obsoletes #547. The consequences for this change is that the report request will fail if it redirects across origins.

Fixes #546.

***
<a href="https://pr-preview.s3.amazonaws.com/WICG/attribution-reporting-api/pull/817.html" title="Last updated on May 18, 2023, 2:00 AM UTC (71d3942)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/attribution-reporting-api/817/4b138b0...71d3942.html" title="Last updated on May 18, 2023, 2:00 AM UTC (71d3942)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2978,6 +2978,8 @@ and a [=header list=] |newHeaders| (defaults to an empty [=list=]):
     :: "`no-referrer`"
     :   [=request/client=]
     ::  `null`
+    :   [=request/origin=]
+    ::  |url|'s [=url/origin=]
     :   [=request/window=]
     ::  "`no-window`"
     :   [=request/service-workers mode=]
@@ -2985,7 +2987,7 @@ and a [=header list=] |newHeaders| (defaults to an empty [=list=]):
     :   [=request/initiator=]
     ::  ""
     :   [=request/mode=]
-    ::  "`cors`"
+    ::  "`same-origin`"
     :   [=request/unsafe-request flag=]
     ::  set
     :   [=request/credentials mode=]


csharrison:
cc @domfarolino would you please review, given the discussion in #547?

apasel422 (approved):

domfarolino:
Were the implementation changes made alongside this? Any web platform tests available to assert this behavior?

csharrison:
We haven't made the implementation changes yet, no, although I think it should be fairly straightforward and it's easier to vet the implementation with a spec PR for references.

domfarolino:
Got it. Is there a web platform test that asserts the behavior in this change? Or at least one that we can change the expectations of soon?

csharrison:
> Got it. Is there a web platform test that asserts the behavior in this change? Or at least one that we can change the expectations of soon?

Not that I can tell from quick glance. Happy to add those before landing this though, whatever you prefer.

domfarolino:
I think that would be nice if it's not too much trouble, just to make sure we're synchronizing the spec and test expectations and implementation, so that neither get lost along the way.

csharrison:
Hey @domfarolino , I added a WPT here:
https://chromium-review.googlesource.com/c/chromium/src/+/4549301

domfarolino (approved):
Great, LGTM. I'd probably lean towards making the tests assertions match the behavior here, and adding WPT expectations that fail until the impl is updated, but up to you.

csharrison:
> Great, LGTM. I'd probably lean towards making the tests assertions match the behavior here, and adding WPT expectations that fail until the impl is updated, but up to you.

Good idea, will do it!

PR #829, opened 2023-05-31
Use aggregation coordinator origin instead of identifier

linnan-github:
To avoid a translation layer of vendor-specific identifier to origin.

***
<a href="https://pr-preview.s3.amazonaws.com/linnan-github/conversion-measurement-api/pull/829.html" title="Last updated on May 31, 2023, 1:54 PM UTC (7cf1bde)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/attribution-reporting-api/829/93010b0...linnan-github:7cf1bde.html" title="Last updated on May 31, 2023, 1:54 PM UTC (7cf1bde)">Diff</a>

### DIFF
--- a/AGGREGATE.md
+++ b/AGGREGATE.md
@@ -304,7 +304,7 @@ shared.
 The encryption will use public keys specified by the aggregation service. The
 browser will encrypt payloads just before the report is sent by fetching the
 public key endpoint (the aggregation service coordinator origin at the path
- `.well-known/aggregation-service/publick-keys`) with an un-credentialed request. The processing origin will
+ `/.well-known/aggregation-service/public-keys`) with an un-credentialed request. The processing origin will
 respond with a set of keys which will be stored according to standard HTTP
 caching rules, i.e. using Cache-Control headers to dictate how long to store the
 keys for (e.g. following the [freshness
@@ -458,8 +458,8 @@ Currently the aggregation service can be deployed on Amazon Web Services (AWS). 
 
 Trigger registration will accept an optional string field `aggregation_coordinator_origin`
 to allow developers to specify the deployment option for the aggregation service
-supported by the browser, e.g. the origin for the aggregation service deloyed on
-AWS, and GCP and other platforms in the future.
+supported by the browser, e.g. the origin for the aggregation service deployed on
+AWS, GCP, and other platforms in the future.
 
 ```jsonc
 {
--- a/AGGREGATE.md
+++ b/AGGREGATE.md
@@ -238,7 +238,7 @@ The report will be JSON encoded with the following scheme:
   ],
 
   // The deployment option for the aggregation service.
-  "aggregation_coordinator_identifier": "aws-cloud",
+  "aggregation_coordinator_origin": "https://publickeyservice.aws.privacysandboxservices.com",
 
   // Optional debugging information (also present in event-level reports),
   // if the cookie `ar_debug` is present.
@@ -303,7 +303,8 @@ shared.
 
 The encryption will use public keys specified by the aggregation service. The
 browser will encrypt payloads just before the report is sent by fetching the
-public key endpoint with an un-credentialed request. The processing origin will
+public key endpoint (the aggregation service coordinator origin at the path
+ `.well-known/aggregation-service/publick-keys`) with an un-credentialed request. The processing origin will
 respond with a set of keys which will be stored according to standard HTTP
 caching rules, i.e. using Cache-Control headers to dictate how long to store the
 keys for (e.g. following the [freshness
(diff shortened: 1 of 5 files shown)


index.bs
```diff
@@ -691,7 +691,7 @@ An event-level trigger configuration is a [=struct=] with the following items:
 <h3 id="aggregation-coordinator-header">Aggregation coordinator</h3>
 
 An <dfn>aggregation coordinator</dfn> is one of a user-agent-determined [=set=]
```
apasel422, line 693: Does this still need to be described as a set in the specification, or does it suffice to say that an aggregation coordinator is an origin?

index.bs
```diff
@@ -2900,9 +2902,17 @@ of running the following steps:
 
 <h3 id="obtain-aggregatable-report-aggregation-service-payloads">Obtaining an aggregatable report's aggregation service payloads</h3>
 
-To <dfn>obtain the public key for encryption</dfn> given an [=aggregation coordinator=]
-<var ignore=''>aggregationCoordinator</var>, asynchronously return a user-agent-determined public key
-or an error in the event that the user agent failed to obtain the public key.
+To <dfn>obtain the public key for encryption</dfn> given an [=aggregation coordinator=] |aggregationCoordinator|:
+
+1. [=Assert=]: |aggregationCoordinator| is not an [=opaque origin=].
```
apasel422, line 2907: Is this assertion necessary? Can we instead specify that `aggregation coordinator` is a non-opaque origin?

index.bs
```diff
@@ -2900,9 +2902,17 @@ of running the following steps:
...
+1. Let |url| be a new [=URL record=].
+1. Set |url|'s [=url/scheme=] to |aggregationCoordinator|'s [=origin/scheme=].
+1. Set |url|'s [=url/host=] to |aggregationCoordinator|'s [=origin/host=].
+1. Set |url|'s [=url/port=] to |aggregationCoordinator|'s [=origin/port=].
+1. Let |path| be «"`.well-known`", "`aggregation-service`", "`public-keys`"».
```
apasel422, line 2912: Is the `|path|` variable necessary? It seems like we can just do `Set |url|'s [=url/path=] to...`

index.bs
```diff
@@ -2900,9 +2902,17 @@ of running the following steps:
...
+1. Set |url|'s [=url/path=] to |path|.
+1. Asynchronously return a user-agent-determined public key from |url| or an error in the event that the user
+    agent failed to obtain the public key from |url|.
```
apasel422, line 2915: This is vague. Is it really user-agent-determined, or should it reference `[=fetch=]` in some way?

And what does it mean for this to be asynchronous? I think it would be better to let the caller of this algorithm decide to invoke it asynchronously, if needed, rather than inherently require the algorithm to be asynchronous.

apasel422 (dismissed):
Please also update the header validator and the source-registration schema.

index.bs
linnan-github: I think it's still useful to describe it as a set as only an origin in the allowlist will be valid?

index.bs
linnan-github: Hmmm that's what we had before, we can think about how to rephrase it in a follow up?

apasel422: In that case, please add issues for implementing this in terms of fetch and for reorganizing the asynchronous part.

linnan-github: Actually rephrased this step a bit. I think the idea is that the public key may be cached in the storage, so it doesn't always have to be fetch.

apasel422: If it can be cached in storage, then we don't even need to construct `url` at all in some cases. I think we will ultimately want to specify fetching the key as one algorithm, and whether to fetch it at all (e.g. how it can be cached) in a different one.

linnan-github: Yeah, that makes sense. Added an issue to specify in terms of fetch. Thanks!

apasel422 (changes requested):
LGTM other than the header validator and source schema changes.

header-validator/validate-json.js
```diff
@@ -192,11 +192,11 @@ const destination = string((state, url) => {
     state.warn('contains a fragment that will be ignored')
   }
 })
-const destinationList = list(destination, 3, 1)
+const destinationList = list(suitableUrl, 3, 1)
```
apasel422, line 195:
```suggestion

const destinationList = list(suitableUrl, 3, 1)
```

linnan-github:
> Please also update the header validator and the source-registration schema.

Updated, thanks!

apasel422 (approved):

apasel422 (approved):

csharrison (approved):

linnan-github:
@hostirosti, @ruclohani, could you PTAL, thanks!

hostirosti (approved):

ruclohani (approved):

AGGREGATE.md
```diff
@@ -303,7 +303,8 @@ shared.
 
 The encryption will use public keys specified by the aggregation service. The
 browser will encrypt payloads just before the report is sent by fetching the
-public key endpoint with an un-credentialed request. The processing origin will
+public key endpoint (the aggregation service coordinator origin at the path
+ `.well-known/aggregation-service/publick-keys`) with an un-credentialed request. The processing origin will
```
dmdabbs: Should be 

`.well-known/aggregation-service/public-keys`

apasel422: This was already fixed in 6e676b61dc067f5fb472afbec847e8706d553d7a.

PR #843, opened 2023-06-08
Add WebDriver extension to spec

johnivdel:
***
<a href="https://pr-preview.s3.amazonaws.com/WICG/attribution-reporting-api/pull/843.html" title="Last updated on Jun 27, 2023, 2:20 PM UTC (1da3719)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/attribution-reporting-api/843/0d79905...1da3719.html" title="Last updated on Jun 27, 2023, 2:20 PM UTC (1da3719)">Diff</a>


csharrison (commented):
High level feedback: can we avoid the "debug" name which conflicts with existing concepts? How about "developer testing mode" or something?

index.bs
```diff
@@ -3261,6 +3271,49 @@ To <dfn noexport>set an OS-support header</dfn> given a [=header list=]
 1. [=header list/Set a structured field value=] given
     ("<code>[=Attribution-Reporting-Support=]</code>", |dict|) in |headers|.
 
+# User Agent Automation # {#automation}
+
+For the purposes of user agent automation and website testing, this document
```
apasel422, line 3276:
```suggestion
# User-Agent Automation # {#automation}

For the purposes of user-agent automation and website testing, this document
```

index.bs
```diff
@@ -3261,6 +3271,49 @@ To <dfn noexport>set an OS-support header</dfn> given a [=header list=]
...
+1. If |parameters| is not a JSON [[ECMASCRIPT#sec-json-object|Object]], return a
+     [=error|WebDriver error=] with [=error code=] [=invalid argument=].
+
+1. Let |enabled| be the result of [=getting a property=] named `"enabled"` from
+    |parameters|.
+
+1. If |enabled| is {{undefined}} or is not a boolean, return a [=error|WebDriver error=]
+    with [=error code=] [=invalid argument=].
+
+1. If |enabled| is true, enables the [=automation debug mode=]; otherwise,
```
apasel422, line 3310:
```suggestion
1. If |enabled| is true, enable [=automation debug mode=]; otherwise,
```

index.bs
```diff
@@ -3261,6 +3271,49 @@ To <dfn noexport>set an OS-support header</dfn> given a [=header list=]
...
+    disables it.
```
apasel422, line 3311:
```suggestion
    disable it.
```

index.bs
```diff
@@ -3261,6 +3271,49 @@ To <dfn noexport>set an OS-support header</dfn> given a [=header list=]
...
+
+1. Return [=success=] with data `null`.
+
+Note: Without this, reports would be subject to noise and delays making testing difficult.
```
apasel422, line 3315:
```suggestion
Note: Without this, reports would be subject to noise and delays, making testing difficult.
```

apasel422 (changes requested):

index.bs
```diff
@@ -3261,6 +3271,49 @@ To <dfn noexport>set an OS-support header</dfn> given a [=header list=]
...
+        <thead>
+            <tr>
+                <th>HTTP Method</th>
+                <th>URI Template</th>
+            </tr>
+        </thead>
+        <tbody>
+            <tr>
+                <td>POST</td>
+                <td>`/session/{session id}/ara/`<dfn>`setdebugmodeenabled`</dfn></td>
```
apasel422, line 3293: Based on other commands in the WebDriver spec, maybe the final component of the URI template should be `debugmode` rather than `setdebugmodeenabled`, i.e. a noun rather than a command?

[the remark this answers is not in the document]
johnivdel: I agree it's a bit confusing. I don't think it implies that this is session sepecific.

For example, https://w3c.github.io/webdriver/#get-current-url has session id in the path, but performs an action that is opaque to which session invoked it.

Ultimately, the rest of the spec is oblivious to what a WebDriver session is, so this seems like the best we can do.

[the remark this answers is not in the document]
apasel422: From https://w3c.github.io/webdriver/#sessions:

> A [session](https://w3c.github.io/webdriver/#dfn-session) is equivalent to a single instantiation of a particular user agent, including all its child browsers.

I'm just wondering what would/should happen if two tests are run simultaneously, one disabling ARA debug mode and one enabling it.

johnivdel:
> High level feedback: can we avoid the "debug" name which conflicts with existing concepts? How about "developer testing mode" or something?

Went with "local testing", let me know if you would prefer something else.

apasel422 (changes requested):

index.bs
```diff
@@ -1013,17 +1018,20 @@ for triggers that are attributed to an [=attribution source=] whose
 0 <= [=event-level trigger configuration/trigger data=] < [=event-source trigger data cardinality=].
 
 <dfn>Randomized response epsilon</dfn> is a non-negative double that controls
-the randomized response probability of an [=attribution source=].
+the randomized response probability of an [=attribution source=]. If <dfn>automation local testing mode</dfn> is enabled,
```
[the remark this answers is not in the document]
johnivdel: Done, also switched to a boolean rather than this bespoke mode type

apasel422 (approved):

PR #848, opened 2023-06-14
Monkeypatch fenced frame beacon sending

linnan-github:
***
<a href="https://pr-preview.s3.amazonaws.com/linnan-github/conversion-measurement-api/pull/848.html" title="Last updated on Jun 22, 2023, 2:05 AM UTC (71732b5)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/attribution-reporting-api/848/aef3a7b...linnan-github:71732b5.html" title="Last updated on Jun 22, 2023, 2:05 AM UTC (71732b5)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1452,9 +1452,7 @@ To <dfn>make background attributionsrc requests</dfn> given an
         [=iteration/continue=]. Otherwise, let |url| be the resulting
         [=URL record=].
     1. Run [=make a background attributionsrc request=] with |url|,
-        |contextOrigin|, |eligibility|, and |element|'s [=Node/node document=].
-
-Issue: Set |contextOrigin| properly.
+        |element|'s [=node/context origin=], |eligibility|, and |element|'s [=Node/node document=].
 
 Issue: Consider allowing the user agent to limit the size of |tokens|.
 
--- a/index.bs
+++ b/index.bs
@@ -1401,13 +1401,14 @@ To <dfn>validate a background attributionsrc eligibility</dfn> given an
     "<code>[=eligibility/navigation-source=]</code>" or
     "<code>[=eligibility/event-source-or-trigger=]</code>".
 
-To <dfn>make a background attributionsrc request</dfn> given a [=URL=] |url|, a
-[=suitable origin=] |contextOrigin|, an [=eligibility=] |eligibility|, and a
+To <dfn>make a background attributionsrc request</dfn> given a [=URL=] |url|, an
+[=origin=] |contextOrigin|, an [=eligibility=] |eligibility|, and a
 {{Document}} |document|:
 
 1. [=Validate a background attributionsrc eligibility|Validate=] |eligibility|.
 1. If |url|'s [=url/scheme=] is not [=check if a scheme is suitable|suitable=],
     return.
+1. If |contextOrigin| is not [=check if an origin is suitable|suitable=], return.
 1. Let |context| be |document|'s [=relevant settings object=].
 1. If |context| is not a [=secure context=], return.
 1. If the "<code>[=attribution-reporting=]</code>" feature
--- a/index.bs
+++ b/index.bs
@@ -530,20 +530,19 @@ and [=this=]'s [=relevant global object=]'s [=associated Document=].
 
 Modify <a spec=fenced-frame>report an event</a> as follows:
 
+<div algorithm="report-an-event-monkeypatch">
+
 Add a [=Document=] parameter called |document|.
 
 After the step
 
-> If <var ignore>reporting map</var>[|destination|] does not [=map/exist=], return.
+> If |reporting map|[|destination|] does not [=map/exist=], return.
 
 add the steps
 
-<div algorithm="report-an-event-monkeypatch">
-1. Let <var ignore>attributionReportingEnabled</var> be the result of determining whether |document| is
+1. Let |attributionReportingEnabled| be the result of determining whether |document| is
     [=allowed to use=] the "<code>[=attribution-reporting=]</code>" feature.
-1. Let <var ignore>contextOrigin</var> be |document|'s [=node/context origin=].
-
-</div>
+1. Let |contextOrigin| be |document|'s [=node/context origin=].
 
 In the step
 
@@ -558,13 +557,15 @@ add the properties
 
 Replace the step
 
-> <a spec=fenced-frame>Send a beacon</a> with <var ignore>reporting map</var>[|destination|],
+> <a spec=fenced-frame>Send a beacon</a> with |reporting map|[|destination|],
 > |eventType|, and |eventData|.
 
 with
 
-<a spec=fenced-frame>Send a beacon</a> with <var ignore>reporting map</var>[|destination|],
+<a spec=fenced-frame>Send a beacon</a> with |reporting map|[|destination|],
 |eventType|, |eventData|, |attributionReportingEnabled|, and |contextOrigin|.
+
+</div>
 
 Modify [=finalize a reporting destination=] as follows:
 
@@ -586,7 +587,7 @@ parameter called |contextOrigin|.
 
 After the step
 
-> Let <var ignore>destination url</var> be <var ignore>destination map</var>[|eventType|].
+> Let <var ignore>destination url</var> be <var ignore>destination map</var>[<var ignore>eventType</var>].
 
 add the steps
 
--- a/index.bs
+++ b/index.bs
@@ -515,18 +515,22 @@ Add the following items to [=fencedframetype/pending event=]:
 
 Modify {{Fence/reportEvent()}} as follows:
 
+<div algorithm="report-event-monkeypatch">
+
 Replace the step
 
-> Run <a spec=fenced-frame>report an event</a> using <var ignore>instance</var>'s
+> Run <a spec=fenced-frame>report an event</a> using |instance|'s
 > [=fenced frame config instance/fenced frame reporter=] with |destination|,
 > |event|'s {{FenceEvent/eventType}}, and |event|'s {{FenceEvent/eventData}}.
 
 with
 
-Run <a spec=fenced-frame>report an event</a> using <var ignore>instance</var>'s
+Run <a spec=fenced-frame>report an event</a> using |instance|'s
 [=fenced frame config instance/fenced frame reporter=] with |destination|,
 |event|'s {{FenceEvent/eventType}}, |event|'s {{FenceEvent/eventData}},
 and [=this=]'s [=relevant global object=]'s [=associated Document=].
+
+</div>
 
 Modify <a spec=fenced-frame>report an event</a> as follows:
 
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -498,7 +498,69 @@ Add the step:
...
+Modify <a spec=fenced-frame>report an event</a> as follows:
+
+Add a [=Document=] parameter called |document|.
+
+Replace the step
+
+> <a spec=fenced-frame>Send a beacon</a> with <var ignore>reporting map</var>[|destination|],
+> |eventType|, and |eventData|.
+
+With
```
apasel422, line 525:
```suggestion
with
```

index.bs
```diff
@@ -498,7 +498,69 @@ Add the step:
...
+Modify [=finalize a reporting destination=] as follows:
+
+Add a [=Document=] parameter called |document|.
+
+Replace the step
+
+> <a spec=fenced-frame>Send a beacon</a> with |destination map|, |pending event|'s [=pending event/eventType=],
+> and |pending event|'s [=pending event/eventData=].
+
+With
```
apasel422, line 539:
```suggestion
with
```

index.bs
```diff
@@ -498,7 +498,69 @@ Add the step:
...
+
+1. Let |attributionReportingEligibility| be "<code>[=eligibility/unset=]</code>".
+1. If the "`attribution-reporting`" feature <a href="https://w3c.github.io/webappsec-permissions-policy/#algo-is-feature-enabled">is enabled</a>
+     in |document| with |document|'s [=origin=] and the result of [=getting supported registrars=]
+     is not [=set/is empty|empty=]:
+         1. Set |attributionReportingEligibility| to "<code>[=eligibility/event-source=]</code>".
+
+         Issue: Allow "<code>[=eligibility/navigation-source=]</code>" when automatic beacon is specified.
+
+Issue: Set [=request=]'s [=request/Attribution Reporting eligibility=] to |attributionReportingEligibility|
```
apasel422, line 562: Please link directly to https://wicg.github.io/fenced-frame/#issue-c137ca33

index.bs
```diff
@@ -498,7 +498,69 @@ Add the step:
...
+
+add the steps
+
+1. Let |attributionReportingEligibility| be "<code>[=eligibility/unset=]</code>".
+1. If the "`attribution-reporting`" feature <a href="https://w3c.github.io/webappsec-permissions-policy/#algo-is-feature-enabled">is enabled</a>
+     in |document| with |document|'s [=origin=] and the result of [=getting supported registrars=]
+     is not [=set/is empty|empty=]:
+         1. Set |attributionReportingEligibility| to "<code>[=eligibility/event-source=]</code>".
+
+         Issue: Allow "<code>[=eligibility/navigation-source=]</code>" when automatic beacon is specified.
```
apasel422, line 560: Is there an issue on the Fenced Frame spec we can link to instead?

apasel422 (changes requested):

index.bs
linnan-github: Not yet, chatted with fenced-frame team earlier to add an issue for that.

linnan-github:
@shivanigithub, would you mind taking a look at the spec changes for ARA integration with ad beacons? Thanks.

linnan-github:
> LGTM but could you get someone from the Fenced Frame side to take a quick look as well?

Thanks, updated with an additional check that context origin should be suitable. Could you please take another look?

index.bs
```diff
@@ -1306,6 +1307,11 @@ is true and whose same-site-flag is "`None`"; there is no way to prevent the
 last-access-time from being modified; and the return value is a string that
 would have to be further processed to check for the "`ar_debug`" cookie.
 
+<h3 algorithm id="obtaining-context-origin">Obtaining context origin</h3>
+
+To obtain the <dfn for=node>context origin</dfn> of a [=node=] |node|, return |node|'s [=node navigable=]'s
+<a spec=fenced-frame for=navigable>unfenced container document</a>'s [=origin=].
```
apasel422, line 1313: We use the term `context origin` elsewhere in the spec. We may want to use something more specifically tied to fenced frame if this dfn only applies to the fenced-frame logic.

linnan-github: Yeah I see context origin used in the spec, and I think this algorithm applies to those cases as well (we need the unfenced top-level origin for all API surfaces), therefore adding this as an algorithm.

apasel422: In that case, I think we should make that clear in this PR by updating the other algorithms to use this new dfn.

linnan-github: Updated and removed the issue, PTAL, thanks.

linnan-github: Actually need to check whether it's suitable. Please wait a bit.

linnan-github: Updated, PTAL, thanks!

apasel422 (approved):

domfarolino:
Can we just modify the fenced frame spec directly to send this header? Is there any reason not to do that?

linnan-github:
> Can we just modify the fenced frame spec directly to send this header? Is there any reason not to do that?

I don't feel strongly, but slightly lean towards keeping ARA specific logic in ARA spec.

shivanigithub:
> @shivanigithub, would you mind taking a look at the spec changes for ARA integration with ad beacons? Thanks.

Will defer this to @domfarolino as Dom closely knows the FF spec.

index.bs
```diff
@@ -498,7 +498,70 @@ Add the step:
...
+Add a [=Document=] parameter called |document|.
+
+After the step
+
+> Let <var ignore>destination url</var> be <var ignore>destination map</var>[|eventType|].
+
+add the steps
+
+1. Let |attributionReportingEligibility| be "<code>[=eligibility/unset=]</code>".
+1. If the "`attribution-reporting`" feature <a href="https://w3c.github.io/webappsec-permissions-policy/#algo-is-feature-enabled">is enabled</a>
```
domfarolino, line 555: Can you link to "is enabled" directly instead of using a manual anchor? This isn't the best practice for such a widely referenced algorithm like permissions policy ones.

index.bs
```diff
@@ -498,7 +498,70 @@ Add the step:
...
+
+> <a spec=fenced-frame>Send a beacon</a> with <var ignore>reporting map</var>[|destination|],
+> |eventType|, and |eventData|.
+
+with
+
+<a spec=fenced-frame>Send a beacon</a> with <var ignore>reporting map</var>[|destination|],
+|eventType|, |eventData|, and |document|.
+
+Modify [=finalize a reporting destination=] as follows:
```
domfarolino, line 530: I think [this algorithm](https://wicg.github.io/fenced-frame/#finalize-a-reporting-destination) is meant to be called from in parallel, so we shouldn't really have direct access to a Document object here. @gtanzer is much more familiar with this part of the config-generating API integration though so let me loop him in here.

domfarolino (commented):
In general I'd recommend moving. My only concern is with adding the document parameter to the "finalize [...]" algorithm in the FF spec. I think that algorithm is meant to be called in parallel so we might not have access to a document to pass into it. Maybe we should instead accept individual origins that are snapshotted at an earlier time, but let me loop gtanzer in to confirm.

index.bs
gtanzer: The implementation doesn't take a document parameter, so I'm not sure why the spec would need to. But I haven't looked into the details yet.
https://source.chromium.org/chromium/chromium/src/+/refs/heads/main:content/browser/fenced_frame/fenced_frame_reporter.h;drc=3e67c0491457f1a05fb7816346fc6d9676bb7a16;bpv=1;bpt=1;l=124?q=fenced_frame_reporter&ss=chromium

linnan-github: Updated to add the permission policy value and the context origin to pending event for later use.

index.bs

index.bs
```diff
@@ -503,7 +505,102 @@ Add the step:
...
+parameter called |contextOrigin|.
+
+After the step
+
+> Let <var ignore>destination url</var> be <var ignore>destination map</var>[|eventType|].
+
+add the steps
+
+1. Let |attributionReportingEligibility| be "<code>[=eligibility/unset=]</code>".
+1. If |attributionReportingEnabled| is true and the result of [=getting supported registrars=]
```
apasel422, line 596: Why do we need to check the registrars being empty here? Shouldn't that be handled universally in https://wicg.github.io/attribution-reporting-api/#set-attribution-reporting-headers ?

apasel422 (changes requested):

index.bs
linnan-github: Ideally we should handle this universally, but I don't think the implementation does this right now, therefore would prefer addressing this separately.

Also the checking here also allows setting process response handler properly when fetch is specified.

apasel422: > Also the checking here also allows setting process response handler properly when fetch is specified.

Could you elaborate on this? I'm not following.

apasel422 (approved):

index.bs
```diff
@@ -522,6 +534,30 @@ Modify <a spec=fenced-frame>report an event</a> as follows:
...
 Add a [=Document=] parameter called |document|.
 
+After the step
+
+> If <var ignore>reporting map</var>[|destination|] does not [=map/exist=], return.
+
+add the steps
+
+<div algorithm="report-an-event-monkeypatch">
+1. Let <var ignore>attributionReportingEnabled</var> be the result of determining whether the
```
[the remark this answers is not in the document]
linnan-github: Updated, thanks!

index.bs
```diff
@@ -40,6 +40,8 @@ spec: idl; type: dfn; urlPrefix: https://webidl.spec.whatwg.org/
     text: throw; url: #dfn-throw
 spec: uuid; type: dfn; urlPrefix: https://wicg.github.io/uuid/
     text: generate a random UUID; url: #dfn-generate-a-random-uuid
+spec: permissions-policy; type: dfn; urlPrefix: https://w3c.github.io/webappsec-permissions-policy/
+    text: is feature enabled; url: #algo-is-feature-enabled
```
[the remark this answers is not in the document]
linnan-github: Done.

index.bs
```diff
@@ -522,6 +534,30 @@ Modify <a spec=fenced-frame>report an event</a> as follows:
...
+</div>
+
+In the step
+
+> Let <var ignore>newEvent</var> be a new [=fencedframetype/pending event=] with the following...
+
+add the properties
+
+: [=pending event/attributionReportingEnabled=]
+:: |attributionReportingEnabled|
```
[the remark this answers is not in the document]
linnan-github: Thanks, expanded the div scope.

domfarolino (commented):
I think this looks good modulo the remaining question about plumbing permissions policy (which might not match the implementation?)

index.bs
linnan-github: The context origin is the top-level origin (https://source.chromium.org/chromium/chromium/src/+/main:content/browser/attribution_reporting/attribution_host.cc;l=444).

I think plumbing the permissions policy value matches the implementation as well, which indicates whether the feature is allowed in the initiator document (https://source.chromium.org/chromium/chromium/src/+/main:content/browser/attribution_reporting/attribution_host.cc;l=436).

domfarolino: Great, OK that makes sense!

domfarolino (approved):

index.bs
linnan-github: Thanks for the reviews!

PR #851, opened 2023-06-15
Move source deactivation step after top level filter matching

### LINKED ISSUE #842: Source deactivation should occur before filter checking in triggering event-level attribution flow
Currently, sources are deactivated after filter matching, rate limit/noise checks, and report generation takes place. However, since a noised source might not lead into a report,  adtechs can have the ability to determine if the source was noised, breaking privacy guarantees. An example of this is when, given S to be a source, T to be a trigger, and E as an event report:

S2 -> S1 (noised) -> T1 -> S1 expires -> T2
- S1 can win and generate a report
- S1 + T1 -> No report generated and S2 is not deactivated because S1 is noised
- S1 expires
- S2 + T2 -> E1

S2 → S1 → T1 → S1 expires → T2 
- S1 can win and generate a report
- S1 + T1 -> E1 generated and S2 deactivated
- S1 expires
- T2 -> No report generated.

Source deactivation should not be dependent on whether the winning source was noised nor if a certain rate limit is reached. It also should not be affected by report generation. To remedy this, source deactivation should occur immediately after the trigger is matched to a source. With this, the behavior of the example above would be as follows:

S2 -> S1 (noised) -> T1 -> S1 expires -> T2
- S1 can win and generate a report
- S1 + T1 -> No report generated and S2 is deactivated
- S1 expires
- T2 -> No report generated.

S2 → S1 → T1 → S1 expires → T2 
- S1 can win and generate a report
- S1 + T1 -> E1 generated and S2 deactivated
- S1 expires
- T2 -> No report generated.

ThomasQuesadilla:
Resolves #842. Renamed to reflect agreed upon changes.

***
<a href="https://pr-preview.s3.amazonaws.com/WICG/attribution-reporting-api/pull/851.html" title="Last updated on Jun 27, 2023, 5:30 PM UTC (c77b36f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/attribution-reporting-api/851/d8e02e8...c77b36f.html" title="Last updated on Jun 27, 2023, 5:30 PM UTC (c77b36f)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2557,8 +2557,6 @@ To <dfn noexport>trigger attribution</dfn> given an [=attribution trigger=] |tri
     1. Return.
 1. Let |sourceToAttribute| be |matchingSources|[0].
 1. [=list/Remove=] |sourceToAttribute| from |matchingSources|.
-1. [=list/iterate|For each=] |item| of |matchingSources|:
-    1. [=set/Remove=] |item| from the [=attribution source cache=].
 1. If the result of running
     [=match an attribution source's filter data against filters and negated filters=] with
     |sourceToAttribute|, |trigger|'s [=attribution trigger/filters=], and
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -2556,6 +2556,9 @@ To <dfn noexport>trigger attribution</dfn> given an [=attribution trigger=] |tri
         with |trigger| and [=generate null reports and assign private state tokens/report=] set to null.
     1. Return.
 1. Let |sourceToAttribute| be |matchingSources|[0].
+1. [=list/Remove=] |sourceToAttribute| from |matchingSources|.
+1. For each |item| of |matchingSources|:
```
apasel422, line 2560:
```suggestion
1. [=list/iterate|For each=] |item| of |matchingSources|:
```

csharrison (approved):
This LGTM, but when implementing this we should think carefully through the breakage, and ideally try to measure.

csharrison:
LGTM but can you rename the PR to say that this is after "top level" filter matching?

johnivdel (approved):

PR #852, opened 2023-06-16
Add explainer on summary reports with key discovery extension.

hidayetaksu:


aggregate_key_discovery.md
```diff
@@ -0,0 +1,90 @@
...
+
+* declare a set of buckets to be included in the output. This can be an empty set. This declaration matches the existing query option in the API.
+* declare a key-mask that specifies the set of key bits ( e.g., ‘0…0001111111111’ to represent the rightmost 10 bits of keyspace) that should be included in the output in addition to explicitly pre-declared buckets.
+* set a threshold, so that only (non pre-declared) buckets with values exceeding the threshold are included in the output.
+* supply multiple key masks with a threshold for each.
+
+The algorithm for computing summary reports with this new mechanism is as follows:
+
+* If a bucket is pre-declared — namely, if the adtech explicitly requested it — it is always present in the output. If that bucket happens to not be present in any input aggregatable report, the aggregation service will still output a noised aggregated value (zero aggregated value + noise) for that bucket. Noise will be directly drawn from the statistical noise distribution mentioned [here](#bookmark=id.nat6mofklxgj).
```
csharrison, line 36:
```suggestion
* If a bucket is pre-declared — namely, if the ad tech explicitly requested it — it is always present in the output. If that bucket happens to not be present in any input aggregatable report, the aggregation service will still output a noised aggregated value (zero aggregated value + noise) for that bucket. Noise will be directly drawn from the statistical noise distribution mentioned [here](#bookmark=id.nat6mofklxgj).
```

aggregate_key_discovery.md
```diff
@@ -0,0 +1,90 @@
...
+* For any other (non pre-declared) bucket matching the key-mask provided by the adtech:
```
csharrison, line 37:
```suggestion
* For any other (non pre-declared) bucket matching the key-mask provided by the ad tech:
```

aggregate_key_discovery.md
```diff
@@ -0,0 +1,90 @@
...
+    * first, noise is added to its aggregated value (the value being zero if the bucket is not present in any input aggregatable report)
+    * then, the bucket is checked against the threshold the ad tech declared. The aggregation service will only include buckets in the output whose noisy values exceed that threshold. Note that this means some buckets may be present in the resulting summary report that have no associated real user contributions.
+
+#### Precision/recall tradeoff
+
+To ensure privacy, noise will be added to all requested buckets before thresholding, even if they are not part of the aggregatable reports. Therefore, buckets without any input may have non-zero values and appear similar to buckets that had input. Thus, a threshold will be chosen that minimizes the number of buckets in the output that are pure noise values (high precision) and maximizes the number of buckets with input (high recall).
+
+The adtech-specified threshold works as a lever to control a trade off between the precision and recall of the output.
```
csharrison, line 45:
```suggestion
The ad-tech-specified threshold works as a lever to control a trade off between the precision and recall of the output. 
```

aggregate_key_discovery.md
```diff
@@ -0,0 +1,90 @@
...
+
+Here precision is the proportion of buckets with user contribution in the output, while recall is the proportion of buckets with user contribution that are correctly reported in the output among all buckets with user contribution.
+
+Both higher precision and higher recall are better, but they are inversely related. Adtech can set thresholds for balancing accuracy/recall with each query. By default the threshold is set to a value that provides 100% precision (`L1 + L1/epsilon*ln(1/delta)`) though at that threshold value recall may be low, as the threshold is set to a relatively high value. Assuming the values of epsilon=10, delta=10^-8, and L1=2^16, then the threshold will be 186257 (2.84*L1) and expected recall will be ~82% for a [sample ads dataset](https://ailab.criteo.com/criteo-sponsored-search-conversion-log-dataset).
```
csharrison, line 52:
```suggestion
Both higher precision and higher recall are better, but they are inversely related. Ad tech can set thresholds for balancing precision/recall with each query. By default the threshold is set to a value that provides 100% precision (`L1 + L1/epsilon*ln(1/delta)`) though at that threshold value recall may be low, as the threshold is set to a relatively high value. Assuming the values of epsilon=10, delta=10^-8, and L1=2^16, then the threshold will be 186257 (2.84*L1) and expected recall will be ~82% for a [sample ads dataset](https://ailab.criteo.com/criteo-sponsored-search-conversion-log-dataset).
```

aggregate_key_discovery.md
```diff
@@ -0,0 +1,90 @@
...
+
+
+## Threshold selection
+
+We intend to publish a tool which demonstrates the impact of threshold selection for precision/recall tradeoff. With this tool, we intend for adtechs to be able to:
```
csharrison, line 57:
```suggestion
We intend to publish a tool which demonstrates the impact of threshold selection for precision/recall tradeoff. With this tool, we intend for ad techs to be able to:
```

aggregate_key_discovery.md
```diff
@@ -0,0 +1,90 @@
...
+
+* Explore impact of threshold on various key-mask, epsilon, delta, with simulated datasets.
+* Find the threshold level that satisfies adtech’s expectations.
```
csharrison, line 60:
```suggestion
* Find threshold values that satisfy their expectations.
```

aggregate_key_discovery.md
```diff
@@ -0,0 +1,90 @@
...
+def legacy_aggregate(aggregatable_reports, pre_declared_labels):
+  return aggregate_with_key_discovery(aggregatable_reports, declared_labels)
+```
+
+Note: the truncated Laplace noise mechanism will apply to all API calls.
+
+
+##  Example call with this new proposal
+
+Assuming an adtech uses the rightmost 42 bits of keyspace to encode attribution keys, and has found that a threshold of L1*2.5 provides a good precision/recall tradeoff for their use case. In this scenario, the adtech's call to the aggregation service would look as follows:
```
csharrison, line 78:
```suggestion
Assuming an ad tech uses the rightmost 42 bits of keyspace to encode attribution keys, and has found that a threshold of L1*2.5 provides a good precision/recall tradeoff for their use case. In this scenario, the ad tech's call to the aggregation service would look as follows:
```

aggregate_key_discovery.md
```diff
@@ -0,0 +1,90 @@
...
+
+
+#### Pre-declaring aggregation buckets, key-mask, and threshold
+
+The Attribution Reporting API and Private Aggregation API client describes a [128 bit key space](https://en.wikipedia.org/wiki/Key_size) for assigning _aggregation keys_, which are referred to as _buckets_ when associated with a value. This large space gives adtechs flexibility in how they define and use aggregated metrics. However, at this size, it's infeasible for the aggregation service to output a value for every possible bucket. To solve this, we are introducing new mechanisms to an ad tech’s query to limit the output size. At query time, ad techs can:
+
+* declare a set of buckets to be included in the output. This can be an empty set. This declaration matches the existing query option in the API.
+* declare a key-mask that specifies the set of key bits ( e.g., ‘0…0001111111111’ to represent the rightmost 10 bits of keyspace) that should be included in the output in addition to explicitly pre-declared buckets.
```
csharrison, line 30:
```suggestion
* declare a key-mask that specifies the set of key bits ( e.g., `"0…0001111111111"` to represent the rightmost 10 bits of keyspace) that should be included in the output in addition to explicitly pre-declared buckets.
```

aggregate_key_discovery.md
```diff
@@ -0,0 +1,90 @@
...
+
+The adtech-specified threshold works as a lever to control a trade off between the precision and recall of the output. 
+
+![Precision recall tradeoff diagram](precision-recall-tradeoff.svg)
+_Fig 1: The chart shows two overlapping distributions, one in blue for false conversions (no user contribution) and one in orange for true conversions (have user contribution). Two distributions are divided by a threshold value, buckets that exceed the threshold (positive) are reported while the buckets that do not exceed the threshold (negative) are not reported in the output._
+
+Here precision is the proportion of buckets with user contribution in the output, while recall is the proportion of buckets with user contribution that are correctly reported in the output among all buckets with user contribution.
```
csharrison, line 50:
```suggestion
- **precision** is the proportion of buckets with true user contributions in the final output. 
- **recall** is the proportion of buckets with true user contributions that are correctly reported in the output (i.e. not dropped during thresholding) among all buckets with user contribution.
```

aggregate_key_discovery.md
```diff
@@ -0,0 +1,90 @@
...
+
+Both higher precision and higher recall are better, but they are inversely related. Adtech can set thresholds for balancing accuracy/recall with each query. By default the threshold is set to a value that provides 100% precision (`L1 + L1/epsilon*ln(1/delta)`) though at that threshold value recall may be low, as the threshold is set to a relatively high value. Assuming the values of epsilon=10, delta=10^-8, and L1=2^16, then the threshold will be 186257 (2.84*L1) and expected recall will be ~82% for a [sample ads dataset](https://ailab.criteo.com/criteo-sponsored-search-conversion-log-dataset).
+
+
+## Threshold selection
+
+We intend to publish a tool which demonstrates the impact of threshold selection for precision/recall tradeoff. With this tool, we intend for adtechs to be able to:
+
+* Explore impact of threshold on various key-mask, epsilon, delta, with simulated datasets.
```
csharrison, line 59:
```suggestion
* Explore the impact of threshold on various values of the key-mask, epsilon, and delta with simulated datasets.
```

aggregate_key_discovery.md
```diff
@@ -0,0 +1,90 @@
...
+* Find the threshold level that satisfies adtech’s expectations.
+* Learn the estimated output size for user-defined settings.
+
+
+## Backward compatible API call
+
+Key-mask by default is all zero (meaning that the default threshold will never kick in) which would emulate the legacy aggregate API call where only buckets in pre_declared_labels are present in the output.
```
csharrison, line 66: What does it mean to "kick in"? Maybe "no bucket with true user contribution will be thresholded"?

hidayetaksu: threshold never **kick in** here means threshold would not be applied to any bucket as no key would match all-zero default key-mask. 
to improve readability, I am going to place it with following text. does that sound better?
"The key-mask is all zeros by default, so no extra keys will be discovered and the default threshold will never be triggered. This means that the output will only contain buckets that are in pre_declared_labels, which is similar to the legacy aggregate API call."

csharrison: LGTM!

csharrison (approved):

PR #855, opened 2023-06-21
Refactor event-level reporting windows

csharrison:
This PR should not induce any behavior change. It just restructures things so that follow-up spec changes can be made to more easily pre-configure reporting windows associated with individual sources.

***
<a href="https://pr-preview.s3.amazonaws.com/WICG/attribution-reporting-api/pull/855.html" title="Last updated on Jun 22, 2023, 1:56 PM UTC (d0d7065)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/attribution-reporting-api/855/b631440...d0d7065.html" title="Last updated on Jun 22, 2023, 1:56 PM UTC (d0d7065)">Diff</a>

Issue #849, opened 2023-06-14
Alternative VTC Event Level Report Configurations


akashnadan:
Currently view through conversions (VTC) for Event Level Reports support 1 conversion, 1 bit of trigger data, and 1 reporting window per ad event.

We are seeking additional feedback on whether another configuration is optimal for your use-cases, specifically on the number of reports and reporting windows for views. Please share any feedback you have on the following:

- Number of conversions per view
- Number of reporting windows
    - If multiple reporting windows, what would be the optimal timing (i.e. 1hr, 1 day, 2 days, etc.) for the windows?
- Impact of noise
    - How much noise would be acceptable for your use cases? For example, assuming an epsilon of 14, the configurations below would have these percentages of noise (calculated using the [flexible_event_privacy.py](https://github.com/WICG/attribution-reporting-api/blob/main/flexible-event/flexible_event_privacy.py) script)
    - Configurations
        - 1 conversion, 1 bit, 1 reporting window = 0.00025% noise
        - 1 conversion, 1 bit, 3 reporting windows = 0.00058% noise
        - 2 conversions, 1 bit, 3 reporting windows = 0.00232% noise
        - 3 conversions, 1 bit, 3 reporting windows = 0.00698% noise

We are considering how to make VTC configurations more flexible to better satisfy ad tech use cases and plan to run tests based on the feedback we receive.

akashnadan:
This has been addressed through the implementation of phase 1 flexible event configurations (ability to configure the number of reports and reporting windows): https://github.com/WICG/attribution-reporting-api/blob/main/EVENT.md#optional-varying-frequency-and-number-of-reports

PR #856, opened 2023-06-22
Add configurable windows and maximum # of reports to spec

### LINKED ISSUE #730: Testing Alternative Reporting Windows
We have heard feedback on reporting delays (reports being received after the expected amount of time) and we are considering testing new values for the 1st and 2nd reporting windows for event level reports (clicks) and testing reduced delays for aggregate reports. We are planning on initially testing the following reporting windows on non-stable channels in Chrome:
- For event level reports (clicks) we plan to test each of the following times separately as the 1st and 2nd reporting windows: [1 hour, 12 hours], [1 hour, 1 day], [1 hour, 2 days], and [6 hours, 7 days]
- For aggregate reports we plan to test each of the following times separately as the new expected reporting delay range: [0 min to 5 min], [0 min to 10 min], [0 min to 20 min], and [0 min to 30 min] <br /><br />

We are seeking feedback on what reporting windows and ranges would be most useful to test based on your use cases:
- For event level reports (clicks): what values for the 1st and 2nd reporting windows should be tested
  - Example: 2 hours, 12 hours, 1 day, etc.
  - Current 1st and 2nd reporting windows for event level reports (for clicks) are 2 days and 7 days

- For aggregate reports: what expected reporting delay range should be tested
  - Example: [X minutes to X minutes]
  - Current expected reporting delay range is [10 minutes to 60 minutes]

We are also seeking feedback on what event level reporting windows for views would be most useful to test as well.

### LINKED ISSUE #736: Updating Reporting Windows for Event-level reports in ARA
**Problem**
The event-level reports in ARA are subject to transmission loss and excessive delays: some conversions registered by the Event API may not be reported or may be delayed. Transmission loss can occur when, for example, the user’s device is not active at the time of the reporting window or if the user clears their cache and deletes the API reports. 

Early OT data and associated debug reports suggest that the total magnitude of transmission loss is not ignorable. We find that, on average, the longer a conversion sits in the browser waiting for the next reporting window, the higher the chance of transmission loss through the Event API.

**Proposal**
We propose to configure the Event API with shorter reporting windows to reduce transmission loss. Currently, Click-Through Conversions are endowed with three windows set to **(2 days, 7 days, source expiry)**. The benefit of shorter windows is that early-arriving conversions will be subject to lower transmission loss rates. However, there are also costs of shorter windows. Conversions that happen after the shorter first window will now be subject to higher loss rates due to a longer gap between the shorter first window and the default 7-day second window, hence we propose changing the second window to a shorter value. We propose a new window configuration of **(1 hour, 2 days, source expiry)** as a new default.

image

**Analysis framework**
Evaluating alternate window configurations requires a target metric. In our analysis, we measure the expected conversion loss as a function of a window configuration W=(w1, w2, w3).

image

Measuring expected conversion loss requires two inputs. The first is a **conversion profile**, which tracks the number of conversions observed (ObsConv) as a function of the conversion delay time (conversion time minus click time). Conversion delay time is represented by the t index in the equation above. The conversion profile is directly observable in the logs for any Adtech. The second is a **transmission loss profile**, which measures the share of reports not sent as a function of the sitting in browser time (reporting window minus conversion time). The transmission loss profile can be learned from OT data.

csharrison:
This PR depends on https://github.com/WICG/attribution-reporting-api/pull/855. This PR specifies a partial version of https://github.com/WICG/attribution-reporting-api/blob/main/flexible_event_config.md, i.e. everything except the `trigger_specs` behavior.

In particular, this allows for a wider range of configurability in the event-level reports, in the form of:
- Allows specifying anywhere between 1 and 5 reporting windows, with (mostly) arbitrary times
- Allows specifying anywhere between 1 and 20 total reports

Fixes #736
Fixes #730 
Addresses #849 % channel capacity limits

***
<a href="https://pr-preview.s3.amazonaws.com/WICG/attribution-reporting-api/pull/856.html" title="Last updated on Jul 17, 2023, 4:10 PM UTC (51dbcd8)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/attribution-reporting-api/856/b5d8bdf...51dbcd8.html" title="Last updated on Jul 17, 2023, 4:10 PM UTC (51dbcd8)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1798,13 +1798,14 @@ and a [=report window list=] |default|:
 1. If |map|["`event_report_windows`"] does not [=map/exist=], return |default|.
 1. Let |values| be |map|["`event_report_windows`"].
 1. If |values| is not a [=map=], return an error.
+1. Let |startDuration| be the result of running [=parse a duration=] with |values|, "`start_time`", and (0, infinity).
+1. If |startDuration| is greater than |defaultEndDuration|, return an error.
 1. If |values|["`end_times`"] does not [=map/exist=] or is not a [=list=], return an error.
 1. Let |endDurations| be |values|["`end_times`"].
 1. If the [=list/size=] of |endDurations| is greater than [=max settable event-level report windows=], return an error.
 1. If |endDurations| [=list/is empty=], return an error.
 1. Let |windows| be an [=list/is empty|empty=] [=list=].
 1. Let |defaultEndDuration| be the [=duration from=] |sourceTime| and the [=report window/end=] of the last element of |default|.
-1. Let |startDuration| be 0 seconds.
 1. [=list/iterate|For each=] |end| of |endDurations|:
     1. If |end| is not an integer, return an error.
     1. Let |endDuration| be |end| seconds.
--- a/index.bs
+++ b/index.bs
@@ -1865,11 +1865,9 @@ To <dfn noexport>parse source-registration JSON</dfn> given a [=byte sequence=]
     1. Round |expiry| away from zero to the nearest day (86400 seconds).
     1. Set |triggerDataCardinality| to the user agent's
         [=event-source trigger data cardinality=].
-1. Let |maxAttributionsPerSource| be the result of running
-    [=parse an optional 64-bit signed integer=] with |value|, "`max_event_level_reports`", and
-    [=default event-level attributions per source=][|sourceType|].
-1. If |maxAttributionsPerSource| is an error, return null.
-1. If |maxAttributionsPerSource| is greater than [=max settable event-level attributions per source=], return null.
+1. Let |maxAttributionsPerSource| be [=default event-level attributions per source=][|sourceType|].
+1. Set |maxAttributionsPerSource| to |value|["`max_event_level_reports`"] if it [=map/exists=].
+1. If |maxAttributionsPerSource| is not a positive integer, or is greater than [=max settable event-level attributions per source=], return null.
 1. Let |eventReportWindow| be the result of running [=parse a duration=] with
     |value|, "`event_report_window`", and ([=min report window=], |expiry|).
 1. If |eventReportWindow| is an error, return null.
--- a/index.bs
+++ b/index.bs
@@ -1797,11 +1797,12 @@ and a [=list=] of [=event-level report windows=] |default|:
 1. If |start| is less than 0, return an error.
 1. Let |startDuration| be |start| seconds.
 1. If |startDuration| is greater than |defaultEndDuration|, return an error.
-1. If |values|["`end_times`"] does not [=map/exist=], return an error.
+1. If |values|["`end_times`"] does not [=map/exist=] or is not a [=list=], return an error.
 1. Let |endDurations| be |values|["`end_times`"].
 1. If the [=list/size=] of |endDurations| is greater than [=max settable event-level report windows=], return an error.
 1. Let |windows| be an [=list/is empty|empty=] [=list=].
 1. [=list/iterate|For each=] |elt| of |endDurations|:
+    1. If |elt| is not a [=string=], return an error.
     1. Let |end| be the result of applying the <a spec="html">rules for parsing integers</a> to |elt|.
     1. If |end| is an error, return an error.
     1. Let |endDuration| be |end| seconds.
--- a/index.bs
+++ b/index.bs
@@ -1800,10 +1800,7 @@ and an [=report window list=] |default|:
 1. If |map|["`event_report_windows`"] does not [=map/exist=], return |default|.
 1. Let |values| be |map|["`event_report_windows`"].
 1. If |values| is not a [=map=], return an error.
-1. Let |start| be the result of running [=parse an optional 64-bit signed integer=] with |values|, "`start_time`", and 0.
-1. If |start| is an error, return an error.
-1. If |start| is less than 0, return an error.
-1. Let |startDuration| be |start| seconds.
+1. Let |startDuration| be the result of running [=parse a duration=] with |values|, "`start_time`", and (0, infinity).
 1. If |startDuration| is greater than |defaultEndDuration|, return an error.
 1. If |values|["`end_times`"] does not [=map/exist=] or is not a [=list=], return an error.
 1. Let |endDurations| be |values|["`end_times`"].
--- a/index.bs
+++ b/index.bs
@@ -1793,9 +1793,8 @@ a [=moment=] |sourceTime|, and a [=duration=] |eventReportWindow|:
 1. Return |windows|.
 
 To <dfn>parse report windows</dfn> given a [=map=] |map|, a [=moment=] |sourceTime|,
-and an [=report window list=] |default|:
+and a [=report window list=] |default|:
 
-1. [=Assert=]: |default| is not [=list/is empty|empty=].
 1. Let |defaultEndDuration| be the [=duration from=] |sourceTime| and the [=report window/end=] of the last element of |default|.
 1. If |map|["`event_report_windows`"] does not [=map/exist=], return |default|.
 1. Let |values| be |map|["`event_report_windows`"].
--- a/index.bs
+++ b/index.bs
@@ -736,11 +736,11 @@ A <dfn>report window</dfn> is a [=struct=] with the following items:
 </dl>
 
 
-An <dfn>event-level report window list</dfn> is a [=list=] of [=event-level report windows=].
+A <dfn>report window list</dfn> is a [=list=] of [=report windows=].
 It has the following constraints:
 
-* Elements are in ascending order based on their [=event-level report window/start=].
-* Every element's [=event-level report window/start=] is equal to the previous element's [=event-level report window/end=], if it exists.
+* Elements are in ascending order based on their [=report window/start=].
+* Every element's [=report window/start=] is equal to the previous element's [=report window/end=], if it exists.
 * There is at least one element in the list.
 
 <h3 dfn-type=dfn>Attribution source</h3>
@@ -763,7 +763,7 @@ An attribution source is a [=struct=] with the following items:
 : <dfn>expiry</dfn>
 :: A [=duration=].
 : <dfn>event-level report windows</dfn>
-:: An [=event-level report window list=].
+:: A [=report window list=].
 : <dfn>aggregatable report window</dfn>
 :: A [=report window=].
 : <dfn>priority</dfn>
@@ -1793,10 +1793,10 @@ a [=moment=] |sourceTime|, and a [=duration=] |eventReportWindow|:
 1. Return |windows|.
 
 To <dfn>parse report windows</dfn> given a [=map=] |map|, a [=moment=] |sourceTime|,
-and an [=event-level report window list=] |default|:
+and an [=report window list=] |default|:
 
 1. [=Assert=]: |default| is not [=list/is empty|empty=].
-1. Let |defaultEndDuration| be the [=duration from=] |sourceTime| and the [=event-level report window/end=] of the last element of |default|.
+1. Let |defaultEndDuration| be the [=duration from=] |sourceTime| and the [=report window/end=] of the last element of |default|.
 1. If |map|["`event_report_windows`"] does not [=map/exist=], return |default|.
 1. Let |values| be |map|["`event_report_windows`"].
 1. If |values| is not a [=map=], return an error.
@@ -1818,11 +1818,11 @@ and an [=event-level report window list=] |default|:
     1. If |endDuration| is greater than |defaultEndDuration|, set |endDuration| to |defaultEndDuration|.
     1. If |endDuration| is less than 1 hour, set |endDuration| to 1 hour.
     1. If |endDuration| less than or equal to |startDuration|, return an error.
-    1. Let |window| be a new [=event-level report window=] whose items are
+    1. Let |window| be a new [=report window=] whose items are
 
-        : [=event-level report window/start=]
+        : [=report window/start=]
         :: |sourceTime| + |startDuration|
-        : [=event-level report window/end=]
+        : [=report window/end=]
         :: |sourceTime| + |endDuration|
 
     1. [=list/Append=] |window| to |windows|.
@@ -2911,8 +2911,8 @@ To <dfn noexport>trigger attribution</dfn> given an [=attribution trigger=] |tri
 To <dfn>check whether a moment falls within a window</dfn> given a [=moment=] |moment| and
 a [=report window=] |window|:
 
-1. Return whether |moment| is greater than or equal to [=event-level report window/start=]
-    and |moment| is less than |window|'s [=event-level report window/end=].
+1. Return whether |moment| is greater than or equal to [=report window/start=]
+    and |moment| is less than |window|'s [=report window/end=].
 
 To <dfn>obtain a report time from a window</dfn> given a [=moment=] |sourceTime| and
 a [=report window=] |window|:
--- a/index.bs
+++ b/index.bs
@@ -1792,7 +1792,7 @@ a [=moment=] |sourceTime|, and a [=duration=] |eventReportWindow|:
 1. Return |windows|.
 
 To <dfn>parse report windows</dfn> given a [=map=] |map|, a [=moment=] |sourceTime|,
-and a [=event-level report window list=] |default|:
+and an [=event-level report window list=] |default|:
 
 1. [=Assert=]: |default| is not [=list/is empty|empty=].
 1. Let |defaultEndDuration| be the [=duration from=] |sourceTime| and the [=event-level report window/end=] of the last element of |default|.
--- a/index.bs
+++ b/index.bs
@@ -1219,11 +1219,11 @@ many [=aggregatable reports=] can be in the [=aggregatable report cache=] per
 
 <dfn>Default event-level attributions per source</dfn> is a [=map=] that
 controls how many times a single [=attribution source=] can create an [=event-level report=] by default.
-The keys are «[=source type/navigation=],[=source type/event=]».
+The keys are «[=source type/navigation=], [=source type/event=]». The values are non-negative integers.
 
 <dfn>Max event-level channel capacity per source</dfn> is a [=map=] that
 controls how many bits of information can be exposed associated with a single [=attribution source=].
-The keys are «[=source type/navigation=],[=source type/event=]».
+The keys are «[=source type/navigation=], [=source type/event=]». The values are non-negative integers.
 
 <dfn>Max aggregatable reports per source</dfn> is a positive integer that controls how many [=aggregatable reports=]
 can be created by [=attribution triggers=] attributed to a single [=attribution source=].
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -1701,6 +1712,41 @@ a [=moment=] |sourceTime|, and a [=duration=] |eventReportWindow|:
...
+1. If |values|["`end_times`"] does not [=map/exist=], return error.
+1. Let |endDurations| be |values|["`end_times`"].
+1. If the [=list/size=] of |endDurations| is greater than 5, return error.
+1. Let |windows| be an [=list/is empty=] [=list=].
+1. [=list/iterate|For each=] |elt| of |endDurations|:
+    1. Let |end| be the result of applying the <a spec="html">rules for parsing integers</a> to |elt|.
+    1. If |end| is an error, return error
+    1. Let |endDuration| be |end| seconds.
+    1. If |endDuration| is greater than |defaultEndDuration|, set |endDuration| to |defaultEndDuration|.
+    1. If |endDuration| <= |startDuration|, return error.
```
apasel422, line 1738:
```suggestion
1. If |values| is not a [=map=], return an error.
1. Let |startDuration| be 0 seconds.
1. If |values|["`start_time`"] [=map/exists=]:
    1. Set |start| to the result of applying the <a spec="html">rules for parsing integers</a> with |values|["`start_time`"].
    1. If |start| is an error, return an error.
    1. If |start| is less than 0, return an error.
    1. Set |startDuration| to |start| seconds.
    1. If |startDuration| is greater than |defaultEndDuration|, return an error.
1. If |values|["`end_times`"] does not [=map/exist=], return an error.
1. Let |endDurations| be |values|["`end_times`"].
1. If the [=list/size=] of |endDurations| is greater than 5, return an error.
1. Let |windows| be an [=list/is empty=] [=list=].
1. [=list/iterate|For each=] |elt| of |endDurations|:
    1. Let |end| be the result of applying the <a spec="html">rules for parsing integers</a> to |elt|.
    1. If |end| is an error, return an error.
    1. Let |endDuration| be |end| seconds.
    1. If |endDuration| is greater than |defaultEndDuration|, set |endDuration| to |defaultEndDuration|.
    1. If |endDuration| <= |startDuration|, return an error.
```

index.bs
```diff
@@ -1701,6 +1712,41 @@ a [=moment=] |sourceTime|, and a [=duration=] |eventReportWindow|:
...
+
+1. Let |defaultEndDuration| be the [=duration from=] |sourceTime| and the [=event-level report window/end=] of the last element of |default|.
+1. If |map|["`event_report_windows`"] does not [=map/exist=], return |default|.
+1. Let |values| be |map|["`event_report_windows`"].
+1. If |values| is not a [=map=], return error.
+1. Let |startDuration| be 0 seconds.
+1. If |values|["`start_time`"] [=map/exists=]:
+    1. Set |start| to the result of applying the <a spec="html">rules for parsing integers</a> with |values|["`start_time`"].
+    1. If |start| is an error, return error.
+    1. If |start| is less than 0, return error.
```
apasel422, line 1726: Can this use https://wicg.github.io/attribution-reporting-api/#parse-an-optional-64-bit-signed-integer? As is, it's unclear to me whether the JSON field needs to be parsed as an arbitrary-precision integer.

index.bs
```diff
@@ -1701,6 +1712,41 @@ a [=moment=] |sourceTime|, and a [=duration=] |eventReportWindow|:
...
+    1. Set |startDuration| to |start| seconds.
+    1. If |startDuration| is greater than |defaultEndDuration|, return error.
+1. If |values|["`end_times`"] does not [=map/exist=], return error.
+1. Let |endDurations| be |values|["`end_times`"].
+1. If the [=list/size=] of |endDurations| is greater than 5, return error.
+1. Let |windows| be an [=list/is empty=] [=list=].
```
apasel422, line 1732:
```suggestion
1. Let |windows| be an [=list/is empty|empty=] [=list=].
```

index.bs
```diff
@@ -1701,6 +1712,41 @@ a [=moment=] |sourceTime|, and a [=duration=] |eventReportWindow|:
...
+1. [=list/iterate|For each=] |elt| of |endDurations|:
+    1. Let |end| be the result of applying the <a spec="html">rules for parsing integers</a> to |elt|.
```
apasel422, line 1734: Likewise, we may want to reuse one of our existing parsing algorithms. We could consider refactoring https://wicg.github.io/attribution-reporting-api/#parse-an-expiry in a separate change.

index.bs
apasel422, line 1738: We should be consistent about using `greater than` and `<=` (text vs symbol).

index.bs
```diff
@@ -1701,6 +1712,41 @@ a [=moment=] |sourceTime|, and a [=duration=] |eventReportWindow|:
...
+    1. If |endDuration| is greater than |defaultEndDuration|, set |endDuration| to |defaultEndDuration|.
+    1. If |endDuration| <= |startDuration|, return error.
+    1. Let |window| be a new [=event-level report window=] whose items are
+
+        : [=event-level report window/start=]
+        :: |sourceTime| + |startDuration|
+        : [=event-level report window/end=]
+        :: |sourceTime| + |endDuration|
+
+    1. [=list/append=] |window| to |windows|.
```
apasel422, line 1746:
```suggestion
    1. [=list/Append=] |window| to |windows|.
```

index.bs
```diff
@@ -1701,6 +1712,41 @@ a [=moment=] |sourceTime|, and a [=duration=] |eventReportWindow|:
...
+    1. Set |startDuration| to |endDuration|.
+1. Return |windows|
```
apasel422, line 1748:
```suggestion
1. Return |windows|.
```

index.bs
```diff
@@ -1740,14 +1786,15 @@ To <dfn noexport>parse source-registration JSON</dfn> given a [=byte sequence=]
...
     1. Round |expiry| away from zero to the nearest day (86400 seconds).
     1. Set |triggerDataCardinality| to the user agent's
         [=event-source trigger data cardinality=].
-    1. Set |maxAttributionsPerSource| to the user agent's
-        [=max attributions per event source=].
+1. Let |maxAttributionsPerSource| be the result of running
+    [=parse an optional 64-bit signed integer=] with |value|, "`max_event_level_reports`", and
+    [=default event-level attributions per source=][|sourceType|].
+1. If |maxAttributionsPerSource| is an error, return null.
+1. If |maxAttributionsPerSource| is greater than 20, return null.
```
apasel422, line 1797: Let's extract 20 into a constant, per my previous comment.

index.bs
csharrison: Done

index.bs
csharrison: I thought about reusing parse-an-expiry but it doesn't really fit the use-case. We could consider a more generic clamping algorithm but I think we can do that in a follow-up.

index.bs
```diff
@@ -1701,6 +1721,40 @@ a [=moment=] |sourceTime|, and a [=duration=] |eventReportWindow|:
     1. Set |lastEnd| to |lastEnd| + |deadline|.
 1. Return |windows|.
 
+To <dfn>parse report windows</dfn> given a [=map=] |map|, a [=moment=] |sourceTime|,
+and a [=list=] of [=event-level report windows=] |default|:
+
+1. [=Assert=]: |default| is not [=list/is empty|empty=].
+1. Let |defaultEndDuration| be the [=duration from=] |sourceTime| and the [=event-level report window/end=] of the last element of |default|.
```
apasel422, line 1728: Do we need to ensure that `default` is sorted? It may make sense to introduce a "wrapper type" like `event-level report window list` that is defined as a list of event-level report windows sorted by X.

index.bs
```diff
@@ -1701,6 +1721,40 @@ a [=moment=] |sourceTime|, and a [=duration=] |eventReportWindow|:
...
+1. If |map|["`event_report_windows`"] does not [=map/exist=], return |default|.
+1. Let |values| be |map|["`event_report_windows`"].
+1. If |values| is not a [=map=], return an error.
+1. Let |start| be the result of running [=parse an optional 64-bit signed integer=] with |values|, "`start_time`", and 0.
+1. If |start| is an error, return an error.
+1. If |start| is less than 0, return an error.
+1. Let |startDuration| be |start| seconds.
+1. If |startDuration| is greater than |defaultEndDuration|, return an error.
+1. If |values|["`end_times`"] does not [=map/exist=], return an error.
```
apasel422, line 1737:
```suggestion
1. If |values|["`end_times`"] does not [=map/exist=] or is not a [=list=], return an error.
```

index.bs
```diff
@@ -1701,6 +1721,40 @@ a [=moment=] |sourceTime|, and a [=duration=] |eventReportWindow|:
...
+1. Let |endDurations| be |values|["`end_times`"].
+1. If the [=list/size=] of |endDurations| is greater than [=max settable event-level report windows=], return an error.
+1. Let |windows| be an [=list/is empty|empty=] [=list=].
+1. [=list/iterate|For each=] |elt| of |endDurations|:
+    1. Let |end| be the result of applying the <a spec="html">rules for parsing integers</a> to |elt|.
```
apasel422, line 1742:
```suggestion
    1. If |elt| is not a [=string=], return an error.
    1. Let |end| be the result of applying the <a spec="html">rules for parsing integers</a> to |elt|.
```

index.bs
```diff
@@ -1758,7 +1813,10 @@ To <dfn noexport>parse source-registration JSON</dfn> given a [=byte sequence=]
 1. If |value|["`debug_reporting`"] [=map/exists=] and is a [=boolean=], set
     |debugReportingEnabled| to |value|["`debug_reporting`"].
 
-1. Let |eventReportWindows| be the result of [=obtaining effective windows=] given |sourceType|, |sourceTime|, and |eventReportWindow|.
+1. Let |eventReportWindows| be the result of [=obtaining default effective windows=] given |sourceType|, |sourceTime|, and |eventReportWindow|.
+1. If |value|["`event_report_window`"] does not [=map/exist=]:
+    1. Set |eventReportWindows| to the result of [=parsing report windows=] with |value|, |sourceTime|, and |eventReportWindows|.
```
apasel422, line 1818: Should it be an error for `event_report_window` and `event_report_windows` to be set in the same registration?

Alternatively, should we reuse the `event_report_window` name and allow it to be either a single value as today *or* the flex lite stuff added in this PR?

apasel422 (changes requested):

index.bs
csharrison: Wrapper type SGTM, let me add one momentarily.

index.bs
csharrison: Using the name `event_report_window` for multiple windows feels very unintuitive to me so I'd prefer making this an error.

apasel422: Agreed on the unintuitiveness, but we did something similar for multi-destination. I don't feel strongly either way, but agree that it should be an error if we do not reuse.

csharrison: Using `destination` to refer to a set of destinations feels less egregious to me than `window` referring to a set of windows. I think this might be because "destination" is a more abstract term.

index.bs
```diff
@@ -1701,6 +1728,43 @@ a [=moment=] |sourceTime|, and a [=duration=] |eventReportWindow|:
     1. Set |lastEnd| to |lastEnd| + |deadline|.
 1. Return |windows|.
 
+To <dfn>parse report windows</dfn> given a [=map=] |map|, a [=moment=] |sourceTime|,
+and a [=event-level report window list=] |default|:
```
apasel422, line 1732:
```suggestion
and an [=event-level report window list=] |default|:
```

csharrison:
> We should also update the header validator and the source JSON schema, though you or I can do that in a followup if you'd prefer.

I'll try to do this in a follow-up. We can also resolve the issue about setting the minimum end duration.

index.bs
```diff
@@ -1701,6 +1728,43 @@ a [=moment=] |sourceTime|, and a [=duration=] |eventReportWindow|:
...
+        : [=event-level report window/start=]
+        :: |sourceTime| + |startDuration|
+        : [=event-level report window/end=]
+        :: |sourceTime| + |endDuration|
+
+    Issue: Set a stricter minimum for |endDuration|.
+
+    1. [=list/Append=] |window| to |windows|.
+    1. Set |startDuration| to |endDuration|.
+1. Return |windows|.
```
apasel422, line 1766: What is supposed to happen if `windows` is empty, e.g. because the `end_times` JSON list was empty?

apasel422 (changes requested):

index.bs
csharrison: 1. I added a constraint to the wrapper type that it needs at least one element.
2. I added an error if |endDurations| is empty.

index.bs
```diff
@@ -1701,6 +1729,44 @@ a [=moment=] |sourceTime|, and a [=duration=] |eventReportWindow|:
...
+1. If |values| is not a [=map=], return an error.
+1. Let |start| be the result of running [=parse an optional 64-bit signed integer=] with |values|, "`start_time`", and 0.
+1. If |start| is an error, return an error.
+1. If |start| is less than 0, return an error.
+1. Let |startDuration| be |start| seconds.
+1. If |startDuration| is greater than |defaultEndDuration|, return an error.
+1. If |values|["`end_times`"] does not [=map/exist=] or is not a [=list=], return an error.
+1. Let |endDurations| be |values|["`end_times`"].
+1. If the [=list/size=] of |endDurations| is greater than [=max settable event-level report windows=], return an error.
+1. If |endDurations| is [=list/empty=], return an error.
```
apasel422, line 1748:
```suggestion
1. If |endDurations| [=list/is empty=], return an error.
```

apasel422 (approved):

csharrison:
Thanks for the thorough review, Andrew 🕺

csharrison:
@vikassahu29 can you please review?

vikassahu29:
> @vikassahu29 can you please review?

Reviewed. LGTM!

csharrison:
cc @yanzhangnyc

apasel422 (changes requested):
We intend to remove the start time, right?

csharrison:
> We intend to remove the start time, right?

Done

apasel422 (approved):

index.bs
```diff
@@ -1765,6 +1792,39 @@ a [=moment=] |sourceTime|, and a [=duration=] |eventReportWindow|:
     1. Set |lastEnd| to |lastEnd| + |deadline|.
 1. Return |windows|.
 
+To <dfn>parse report windows</dfn> given a [=map=] |map|, a [=moment=] |sourceTime|,
+and an [=report window list=] |default|:
```
apasel422, line 1796:
```suggestion
and a [=report window list=] |default|:
```

index.bs
```diff
@@ -1765,6 +1792,39 @@ a [=moment=] |sourceTime|, and a [=duration=] |eventReportWindow|:
...
+
+1. [=Assert=]: |default| is not [=list/is empty|empty=].
+1. Let |defaultEndDuration| be the [=duration from=] |sourceTime| and the [=report window/end=] of the last element of |default|.
```
apasel422, line 1799: Let's move this step closer to where this variable is first used, i.e. before iterating over `|endDurations|`.

index.bs
```diff
@@ -1765,6 +1792,39 @@ a [=moment=] |sourceTime|, and a [=duration=] |eventReportWindow|:
...
+1. If the [=list/size=] of |endDurations| is greater than [=max settable event-level report windows=], return an error.
+1. If |endDurations| [=list/is empty=], return an error.
+1. Let |windows| be an [=list/is empty|empty=] [=list=].
+1. [=list/iterate|For each=] |elt| of |endDurations|:
+    1. If |elt| is not a [=string=], return an error.
+    1. Let |end| be the result of applying the <a spec="html">rules for parsing integers</a> to |elt|.
+    1. If |end| is an error, return an error.
+    1. Let |endDuration| be |end| seconds.
+    1. If |endDuration| is greater than |defaultEndDuration|, set |endDuration| to |defaultEndDuration|.
+    1. If |endDuration| is less than 1 hour, set |endDuration| to 1 hour.
```
apasel422, line 1815: Instead of 1 hour, should this be referencing https://wicg.github.io/attribution-reporting-api/#min-report-window?

apasel422 (changes requested):

index.bs
```diff
@@ -1803,14 +1862,15 @@ To <dfn noexport>parse source-registration JSON</dfn> given a [=byte sequence=]
...
-1. Let |maxAttributionsPerSource| be the user agent's
-    [=max attributions per navigation source=].
 1. If |sourceType| is "<code>[=source type/event=]</code>":
     1. Round |expiry| away from zero to the nearest day (86400 seconds).
     1. Set |triggerDataCardinality| to the user agent's
         [=event-source trigger data cardinality=].
-    1. Set |maxAttributionsPerSource| to the user agent's
-        [=max attributions per event source=].
+1. Let |maxAttributionsPerSource| be the result of running
+    [=parse an optional 64-bit signed integer=] with |value|, "`max_event_level_reports`", and
```
apasel422, line 1870: Seems like a previous comment of mine was lost here: This differs from the explainer, which says this field is an integer, not a string: https://github.com/WICG/attribution-reporting-api/blob/main/flexible_event_config.md#api-changes

csharrison: Done.

csharrison: (Treated as an integer)

apasel422: Isn't this still requiring `max_event_level_reports` to be a string?

apasel422 (changes requested):

index.bs
csharrison: Yes sorry, forgot to push the recent change. I'll upload a combined fix which also re-adds `start_time` today

csharrison: PTAQL

apasel422 (approved):

index.bs
```diff
@@ -1765,6 +1792,37 @@ a [=moment=] |sourceTime|, and a [=duration=] |eventReportWindow|:
...
+1. Let |startDuration| be the result of running [=parse a duration=] with |values|, "`start_time`", and (0, infinity).
+1. If |startDuration| is greater than |defaultEndDuration|, return an error.
+1. If |values|["`end_times`"] does not [=map/exist=] or is not a [=list=], return an error.
+1. Let |endDurations| be |values|["`end_times`"].
+1. If the [=list/size=] of |endDurations| is greater than [=max settable event-level report windows=], return an error.
+1. If |endDurations| [=list/is empty=], return an error.
+1. Let |windows| be an [=list/is empty|empty=] [=list=].
+1. Let |defaultEndDuration| be the [=duration from=] |sourceTime| and the [=report window/end=] of the last element of |default|.
+1. [=list/iterate|For each=] |end| of |endDurations|:
+    1. If |end| is not an integer, return an error.
```
apasel422, line 1810: Here we're saying that each element of `end_times` is an integer, but above we are parsing `start_time` as a string.

csharrison: Good catch, fixed.

index.bs
```diff
@@ -1803,14 +1861,13 @@ To <dfn noexport>parse source-registration JSON</dfn> given a [=byte sequence=]
...
-    [=max attributions per navigation source=].
 1. If |sourceType| is "<code>[=source type/event=]</code>":
     1. Round |expiry| away from zero to the nearest day (86400 seconds).
     1. Set |triggerDataCardinality| to the user agent's
         [=event-source trigger data cardinality=].
-    1. Set |maxAttributionsPerSource| to the user agent's
-        [=max attributions per event source=].
+1. Let |maxAttributionsPerSource| be [=default event-level attributions per source=][|sourceType|].
+1. Set |maxAttributionsPerSource| to |value|["`max_event_level_reports`"] if it [=map/exists=].
+1. If |maxAttributionsPerSource| is not a positive integer, or is greater than [=max settable event-level attributions per source=], return null.
```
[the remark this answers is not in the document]
csharrison: Good idea, done/

apasel422 (changes requested):

index.bs
```diff
@@ -1765,6 +1792,39 @@ a [=moment=] |sourceTime|, and a [=duration=] |eventReportWindow|:
     1. Set |lastEnd| to |lastEnd| + |deadline|.
 1. Return |windows|.
 
+To <dfn>parse report windows</dfn> given a [=map=] |map|, a [=moment=] |sourceTime|,
+and a [=report window list=] |default|:
+
+1. If |map|["`event_report_windows`"] does not [=map/exist=], return |default|.
+1. Let |values| be |map|["`event_report_windows`"].
+1. If |values| is not a [=map=], return an error.
+1. Let |start| be |map|["`start_time`"]
```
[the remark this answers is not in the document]
csharrison: Done

apasel422 (approved):

PR #858, opened 2023-06-22
Use current wall time instead of the one on environment settings object

linnan-github:
The implementation uses base::Time::Now() which doesn't consider cross origin isolated contexts as well.

***
<a href="https://pr-preview.s3.amazonaws.com/linnan-github/conversion-measurement-api/pull/858.html" title="Last updated on Jun 22, 2023, 12:53 PM UTC (451d320)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/attribution-reporting-api/858/d91215b...linnan-github:451d320.html" title="Last updated on Jun 22, 2023, 12:53 PM UTC (451d320)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1469,8 +1469,7 @@ To <dfn>make a background attributionsrc request</dfn> given a [=URL=] |url|, an
     :   [=request/Attribution Reporting eligibility=]
     ::  |eligibility|
 1. [=Fetch=] |request| with [=fetch/processResponse=] being
-    [=process an attributionsrc response=] with |contextOrigin|, |eligibility|,
-    and |context|.
+    [=process an attributionsrc response=] with |contextOrigin| and |eligibility|.
 
 Issue: Audit other properties on |request| and set them properly.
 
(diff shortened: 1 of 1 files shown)


yoavweiss:
> The implementation uses base::Time::Now()

It seems fine to move to more coarsened times by default, especially given that it seems the requirement time granularity you require here is very coarse. It's worthwhile to verify the implementation is similarly coarsening the timestamps correctly.

Can you think of ways to write a WPT that tests that? (not a blocker, but would be great to have)

yoavweiss (commented):
LGTM

yoavweiss (approved):

linnan-github:
> > The implementation uses base::Time::Now()
> 
> It seems fine to move to more coarsened times by default, especially given that it seems the requirement time granularity you require here is very coarse. It's worthwhile to verify the implementation is similarly coarsening the timestamps correctly.
> 
> Can you think of ways to write a WPT that tests that? (not a blocker, but would be great to have)

Thanks, we will look into the implementation to see whether the time is similarly coarsened.

PR #865, opened 2023-06-22
Reducing event_report_window minimum

akashnadan:
Reducing the event_report_window minimum from 86400 seconds (1 day) to 3600 seconds (1 hour)


csharrison (approved):

apasel422 (commented):
Do we also have to update https://wicg.github.io/attribution-reporting-api/#parse-source-registration-json ?

csharrison:
> Do we also have to update https://wicg.github.io/attribution-reporting-api/#parse-source-registration-json ?

Yeah, I can do it in a follow-up

Issue #793, opened 2023-05-12
Reject registrations with invalid input instead of setting silent defaults


giladbarkan-github:
Our primary concern is that silent defaults make it impossible in some cases for ARA consumers to know about an issue. 

Let's say an ad-tech is using an expiry: 86400 instead of expiry: "86400". ARA would extend it to 30 days without notifying them and change their attribution logic.

In case of rejection, 
- They won't receive source success reports
- They'll also not receive event/aggregate report which will force them to look into the issue during integration.

Specific places we noticed:

#### Source registration
source_event_id
expiry
event_report_window
aggregatable_report_window
priority

We would argue not to reject a registration if debug_key is invalid.

#### Trigger registration
event_trigger_data -> trigger_data
event_trigger_data -> priority
Aggregatable_deduplication_keys -> dedupKey -> deduplication_key

csharrison:
This seems reasonable to me. @giladbarkan-github would you be interested in discussing this in our call on Monday? It's at 8am PT.

giladbarkan-github:
Discussing this in the call on Monday at 8am PT sounds good to me! Please let me know the details.

csharrison:
I am supportive of this, especially given that we have the success reports right now which will indicate failures to the ad-tech. Something like #801 could be follow-up work to consider when success reports go away.

apasel422:
> We would argue not to reject a registration if debug_key is invalid.

I'm fine with this, but what is the reasoning?

apasel422:
> expiry
> event_report_window
> aggregatable_report_window

Could you elaborate on what is considered an error here? Certainly it is an error for the field to have a type other than string or for the string to fail parsing as a base-10 signed integer, but if the value is out of range, should it be considered an error or clamped to the allowed range?

apasel422:
What about the `debug_reporting` fields? Should a registration be rejected if the field is present but isn't a boolean?

johnivdel:
The argument for not rejecting based on debug_key seems to be that debug_key being ignored does not change the reporting origins expectation for how attribution is done by the browser. I think `debug_reporting` should be treated similarly to debug_key for that reason.

That being said, this seems to be a fairly weak protection for developers, and it would be nice if we didn't need to special case any fields in the registration.

apasel422:
> does not change the reporting origins expectation for how attribution is done by the browser

Wouldn't this also apply to clamping source expiry and reporting windows?

giladbarkan-github:
[...]

PR #876, opened 2023-06-29
Set the min event-report-window to 1 hour

### LINKED ISSUE #561: Consider hour-level expiry within a one-day conversion

csharrison:
Fixes #561. Because this change only addresses the report windows and not the expiry, the privacy attack outlined in https://github.com/WICG/attribution-reporting-api/issues/561#issuecomment-1262347679 is not possible.

This PR also makes one behavior change: if durations are passed in with negative values, we error out rather than clamping.

***
<a href="https://pr-preview.s3.amazonaws.com/WICG/attribution-reporting-api/pull/876.html" title="Last updated on Jun 29, 2023, 2:34 PM UTC (558b39b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/attribution-reporting-api/876/fe4f2a4...558b39b.html" title="Last updated on Jun 29, 2023, 2:34 PM UTC (558b39b)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1690,19 +1690,18 @@ To <dfn>parse attribution destinations</dfn> from a [=map=] |map|:
     [=max destinations per source=], return null.
 1. Return |result|.
 
-To <dfn>parse an expiry</dfn> given a [=map=] |map|, a [=string=] |key|, and a
-positive [=duration=] |default|:
+To <dfn>parse a duration</dfn> given a [=map=] |map|, a [=string=] |key|, a
+tuple of [=durations=] (|clampStart|, |clampEnd|):
 
-1. [=Assert=]: |default| is greater than or equal to 1 day.
-1. [=Assert=]: |default| is less than or equal to the [=max source expiry=].
 1. Let |seconds| be the result of running
     [=parse an optional 64-bit signed integer=] with |map|, |key|, and null.
 1. If |seconds| is an error, return an error.
-1. If |seconds| is null, return |default|.
-1. Let |expiry| be |seconds| seconds.
-1. If |expiry| is less than 1 day, return 1 day.
-1. If |expiry| is greater than |default|, return |default|.
-1. Return |expiry|.
+1. If |seconds| is null, return |clampEnd|.
+1. If |seconds| is less than 0, return an error.
+1. Let |duration| be the [=duration=] of |seconds| seconds.
+1. If |duration| is less than |clampStart|, return |clampStart|.
+1. If |duration| is greater than |clampEnd|, return |clampEnd|.
+1. Return |duration|.
 
 Issue: Consider rejecting out-of-bounds values instead of silently clamping.
 
@@ -1757,8 +1756,8 @@ To <dfn noexport>parse source-registration JSON</dfn> given a [=byte sequence=]
     [=parse an optional 64-bit unsigned integer=] with |value|,
     "`source_event_id`", and 0.
 1. If |sourceEventId| is an error, return null.
-1. Let |expiry| be the result of running [=parse an expiry=] with |value|,
-    "`expiry`", and the [=max source expiry=].
+1. Let |expiry| be the result of running [=parse a duration=] with |value|,
+    "`expiry`", and (1 day, [=max source expiry=]).
 1. If |expiry| is an error, return null.
 1. Let |priority| be the result of running
     [=parse an optional 64-bit signed integer=] with |value|, "`priority`", and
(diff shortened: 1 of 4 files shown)


index.bs
```diff
@@ -1690,19 +1690,18 @@ To <dfn>parse attribution destinations</dfn> from a [=map=] |map|:
     [=max destinations per source=], return null.
 1. Return |result|.
 
-To <dfn>parse an expiry</dfn> given a [=map=] |map|, a [=string=] |key|, and a
-positive [=duration=] |default|:
+To <dfn>parse a duration</dfn> given a [=map=] |map|, a [=string=] |key|, a
```
apasel422, line 1693:
```suggestion
To <dfn>parse a duration</dfn> given a [=map=] |map|, a [=string=] |key|, and a
```

index.bs
```diff
@@ -1690,19 +1690,18 @@ To <dfn>parse attribution destinations</dfn> from a [=map=] |map|:
...
+tuple of [=durations=] (|clampStart|, |clampEnd|):
 
-1. [=Assert=]: |default| is greater than or equal to 1 day.
-1. [=Assert=]: |default| is less than or equal to the [=max source expiry=].
 1. Let |seconds| be the result of running
```
apasel422, line 1696:
```suggestion
1. [=Assert=]: |clampStart| < |clampEnd|.
1. Let |seconds| be the result of running
```

index.bs
```diff
@@ -1757,8 +1756,8 @@ To <dfn noexport>parse source-registration JSON</dfn> given a [=byte sequence=]
     [=parse an optional 64-bit unsigned integer=] with |value|,
     "`source_event_id`", and 0.
 1. If |sourceEventId| is an error, return null.
-1. Let |expiry| be the result of running [=parse an expiry=] with |value|,
-    "`expiry`", and the [=max source expiry=].
+1. Let |expiry| be the result of running [=parse a duration=] with |value|,
```
apasel422, line 1759: Let's extract the 1 day, 1 hour, and 1 hour clamp-start values into constants for ease of reference.

apasel422:
> This PR also makes one behavior change: if durations are passed in with negative values, we error out rather than clamping.

CC #793, @giladbarkan-github

index.bs
```diff
@@ -1690,19 +1690,18 @@ To <dfn>parse attribution destinations</dfn> from a [=map=] |map|:
...
 
-To <dfn>parse an expiry</dfn> given a [=map=] |map|, a [=string=] |key|, and a
-positive [=duration=] |default|:
+To <dfn>parse a duration</dfn> given a [=map=] |map|, a [=string=] |key|, a
+tuple of [=durations=] (|clampStart|, |clampEnd|):
 
-1. [=Assert=]: |default| is greater than or equal to 1 day.
-1. [=Assert=]: |default| is less than or equal to the [=max source expiry=].
 1. Let |seconds| be the result of running
     [=parse an optional 64-bit signed integer=] with |map|, |key|, and null.
```
apasel422, line 1697: Should we just parse this as an unsigned integer if we're going to prohibit negative values anyway? There are slightly different considerations for the spec and the implementation (not a big deal given that the clamp range easily fits in both types), but I don't see a benefit to having the spec parse the value as signed only to error out manually.

apasel422 (changes requested):
Please also update the header validator and source-registration schema.

index.bs
giladbarkan-github: Would that mean a negative number returns an error rather than be clamped? That would seem breaking.

apasel422: That's exactly what this PR does anyway a few lines later.

giladbarkan-github: Oh, thanks for pointing that out! (And reminding me to read more before commenting!)

apasel422 (approved):
LGTM, but do we need to do anything to address the breaking part of this?

csharrison:
> LGTM, but do we need to do anything to address the breaking part of this?

Given that we don't expect this to cause any actual breakage, let's handle that question on the implementation side.

giladbarkan-github (approved):

PR #886, opened 2023-07-10
Minor fixes and clarifications to flex explainer

csharrison:
Fixes some examples which incorrectly indicated a batching mechanism. Also beefs up some comments.

### DIFF
--- a/flexible_event_config.md
+++ b/flexible_event_config.md
@@ -63,6 +63,11 @@ We will add two optional parameters to the JSON in `Attribution-Reporting-Regist
     // e.g. [5, 10, 100] encodes the following ranges:
     // [[0, 4], [5, 9], [10, 99], [100, MAX_INT]]
     //
+    // At the end of each reporting window, triggers will be summarized into an
+    // integer which slots into one of these ranges. Reports will be sent for
+    // every new range boundary that is crossed. Reports will never be sent for
+    // the range that includes 0, as every source is initialized in this range.
+    //
     // If omitted, then represents a trivial mapping
     // [1, 2, ... , MAX_INT]
     // With MAX_INT being the maximum possible int value defined by the browser.
@@ -195,10 +200,10 @@ This example configuration supports a developer who wants to optimize for value 
   {
     "trigger_data": [0],
     "event_report_windows": {
-      "end_times": [<7 days>]
+      "end_times": [<7 days>, <14 days>]
     },
     "summary_window_operator": "value_sum",
-    "summary_buckets": [5, 10, 100]
+    "summary_buckets": [1, 5, 10, 100]
   }],
 }
 ```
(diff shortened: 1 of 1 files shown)


csharrison:
@avvarad @vikassahu29

flexible_event_config.md
```diff
@@ -63,6 +63,11 @@ We will add two optional parameters to the JSON in `Attribution-Reporting-Regist
     // e.g. [5, 10, 100] encodes the following ranges:
     // [[0, 4], [5, 9], [10, 99], [100, MAX_INT]]
     //
+    // At the end of each reporting window, triggers will be summarized into an
+    // integer which slots into one of these ranges. Reports will be sent for
+    // every new range boundary that is crossed. Reports will never be sent for
```
apasel422, line 68: What does `new` mean in this context?

flexible_event_config.md
```diff
@@ -212,15 +217,46 @@ Triggers could be registered with the value field set, which are summed up and b
 { "event_trigger_data": [{"trigger_data": "0", "value": 4}] }
 ```
 
-The values are summed (to 8) and reported as bucket [5, 9]
+The values are summed (to 8) and reported in the following reports after 7 days + 1 hour:
```
apasel422, line 220: I'm still having a bit of difficulty understanding the mapping between triggers and reports. To help clarify, let's say each of the above triggers has a different `debug_key`. What would the reports look like?

flexible_event_config.md
csharrison: "new" relative to the ranges you've previously visited. Here is the high level algorithm:
- At the beginning of a window, mark the range you are in $R_b$. For the first window, this will always be the "null" range with 0 in it.
- At the end of a window, mark the range you are in $R_e$
- Emit reports for all bucket ranges in $(R_b, R_e]$

LMK if you can think of a better way of briefly describing this.

flexible_event_config.md
csharrison: Hm I think we will need to spend some extra thought specifying how `debug_key` will compose with flexible event. This is because we lose the 1:1 mapping from triggers --> reports. With this proposal, multiple triggers can contribute to a single report, as they "aggregate" together within a particular summary bucket.

We could do a few things:
- keep track of all triggers that contributed to a report and emit all of their debug keys in a list.
- Emit the debug key of the last trigger which 
- etc

apasel422: I think that a list of all them is the most natural.

Separate from the keys themselves, though, we will have to consider the format, semantics, and benefit of [attribution-success debug reports](https://github.com/WICG/attribution-reporting-api/blob/main/EVENT.md#attribution-success-debugging-reports).

apasel422 (approved):

flexible_event_config.md
yanzhangnyc, line 220: There are some ongoing work to set different delay for the reporting time. Do we need to make some change?

csharrison: The report time is determined by the `end_times` field here so it shouldn't be related to any changes in defaults, etc.

flexible_event_config.md
```diff
@@ -195,10 +200,10 @@ This example configuration supports a developer who wants to optimize for value
   {
     "trigger_data": [0],
     "event_report_windows": {
-      "end_times": [<7 days>]
+      "end_times": [<7 days>, <14 days>]
     },
     "summary_window_operator": "value_sum",
-    "summary_buckets": [5, 10, 100]
+    "summary_buckets": [1, 5, 10, 100]
```
avvarad, line 206: Why add the "1" here? Would be nice to have one example where the buckets don't start at 1

csharrison: Let me remove it. At some point I had a good reason to add this but after I added the other triggers I don't anymore.

avvarad (approved):

PR #889, opened 2023-07-13
Remove erroneous 30-day report window from default navigation config

apasel422:
Presumably this 30-day value was meant to represent the reporting window corresponding to the source's expiry, but that window is implicit, as indicated in the correctly defined default configuration for event sources .

### DIFF
--- a/flexible_event_config.md
+++ b/flexible_event_config.md
@@ -175,7 +175,7 @@ Here are the default configurations for event and navigation sources. Especially
   {
     "trigger_data": [0, 1, 2, 3, 4, 5, 6, 7],
     "event_report_windows": {
-      "end_times": [<2 days>, <7 days>, <30 days>]
+      "end_times": [<2 days>, <7 days>]
     }
   }],
   "max_event_level_reports": 3


csharrison (approved):
Looks good. Yeah I would assume these windows follow a similar algorithm as https://github.com/WICG/attribution-reporting-api/pull/856

yanzhangnyc:
IMO, whether keep the "30 days" or not doesn't make any difference, why it is important to remove it. In flexible event API, it is common to have several different setting are actually the same for the backend. Explicitly set 30 days may be clear than implicitly to fall to expiry.

apasel422:
> IMO, whether keep the "30 days" or not doesn't make any difference, why it is important to remove it.

If you want to keep it, shouldn't we also add it to [the default event-level configuration](https://github.com/WICG/attribution-reporting-api/blob/main/flexible_event_config.md#default-event-sources)? The inconsistency is itself a problem.

> Explicitly set 30 days may be clear than implicitly to fall to expiry.

Could you clarify this? 30 days is not actually the default. The final report window is based on the `event_report_window` which is itself based on the `expiry`, which defaults to 30 days. 30 days is not always the default value even today. Please see #856 for details.

To be clear here, this PR changes a document that is describing the equivalent registration configuration JSON when no `trigger_specs` field is provided. It is *not* describing the *actual* report windows that the implementation uses, which are described in an algorithm in the spec. In other words, the final reporting window is implicit and never part of the trigger specification itself.

csharrison:
Actually isn't #856 compatible with this? In that pr we don't append a default window.

apasel422:
> For example, if expiry set as 30 days, and end times are [2 day, 7 days], there are technically 2 report windows or 3 report windows?

There is always at least 1 report window, which is when the source expires.

> I suggest to keep all report windows explicit in end_times, and expiry is only serve as a guardrail to void the registration if it is earlier than last window.

Do  you agree, then, that https://github.com/WICG/attribution-reporting-api/blob/main/flexible_event_config.md#default-event-sources should have an additional value of 30 days at the end of *iIts* list?

apasel422:
Closing in favor of a different PR that clarifies the existing examples.

PR #789, opened 2023-05-10
Add a section for the specification of flexible event report API

yanzhangnyc:
***
<a href="https://pr-preview.s3.amazonaws.com/WICG/attribution-reporting-api/pull/789.html" title="Last updated on May 15, 2023, 2:46 AM UTC (8fe097d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/attribution-reporting-api/789/9044fce...8fe097d.html" title="Last updated on May 15, 2023, 2:46 AM UTC (8fe097d)">Diff</a>

PR #890, opened 2023-07-13
Add non-normative section for flex-event specification

apasel422:
I would add even more definitions here for the trigger spec structure itself, but a lot of that is dependent on choices for fixed-width arithmetic and reusing parts of #856.

***
<a href="https://pr-preview.s3.amazonaws.com/apasel422/attribution-reporting-api/pull/890.html" title="Last updated on Jul 13, 2023, 4:31 PM UTC (9d9f44e)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/attribution-reporting-api/890/892ff4d...apasel422:9d9f44e.html" title="Last updated on Jul 13, 2023, 4:31 PM UTC (9d9f44e)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -3810,3 +3810,23 @@ Possible mitigations:
     presence from the reporting origin. Compared to the previous mitigation, the
     proxy server could itself handle the [=event-level report/trigger priority=]
     functionality, at the cost of increased complexity in the proxy.
+
+# Flexible event-level configurations # {#flexible-event-level-configurations}
+
+*This section is non-normative.*
+
+This is a more formal description of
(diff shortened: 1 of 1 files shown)


csharrison:
What is the relation between this and https://github.com/WICG/attribution-reporting-api/pull/789? Are we going to land these piecemeal or do you want to keep it in a PR while it is fleshed out?

apasel422:
> What is the relation between this and #789? Are we going to land these piecemeal or do you want to keep it in a PR while it is fleshed out?

That PR seems to be abandoned, but regardless, I think some amount of iterative work will be required due to the sheer size of the proposal. I will add a few more substantive commits to this PR before merging, with the caveat that there will be a number of inline issues about integers vs strings and integer sizes, and I would still like #856 to land first.

PR #896, opened 2023-07-18
Add basic tests for JSON schemas

apasel422:


csharrison (approved):
Generally RS LGTM, but how can we avoid duplicating work across the header validator and this is still an open question to me, and I don't want to end up in a state where we have to maintain essentially two copies of every test.

apasel422:
> Generally RS LGTM, but how can we avoid duplicating work across the header validator and this is still an open question to me, and I don't want to end up in a state where we have to maintain essentially two copies of every test.

It is possible to use the schemas to drive the interactive header validator, but I need to read up on how to effectively produce a combined JS script that can be loaded in the browser environment.

@maudnals may be able to help us with this.

PR #898, opened 2023-07-18
Flex lite schema and header validator

csharrison:


apasel422 (approved):
Nice! I would like to land #896 first, though, so we get CI on this.

PR #899, opened 2023-07-19
Emphasize the strict-increments for end_times and summary_bucket

yanzhangnyc:


flexible_event_config.md
```diff
@@ -143,24 +143,26 @@ In addition to the parameters that were added in Phase 1, we will add one additi
     // Represents a series of time windows, starting at the source registration time.
     // Reports for this spec will be delivered an hour after the end of each window.
     // Time is encoded as seconds after source registration.
+    // end_times should consist of strictly increasing integers without repetitions.
```
apasel422, line 146: If we're adding this at all, I would say `must` instead of `should` (and same below): https://datatracker.ietf.org/doc/html/rfc2119

However, I am curious why we need to require this property, when the end times are effectively just a set of positive integers.

flexible_event_config.md
```diff
@@ -143,24 +143,26 @@ In addition to the parameters that were added in Phase 1, we will add one additi
...
     // value_sum: sum of the value of triggers within a window
     // The summary is reported as an index into a bucketization scheme.
     // Defaults to "count"
-    "summary_window_operator": <one of "count" or "value_sum">
+    "summary_window_operator": <one of "count" or "value_sum">, 
 
     // Represents a bucketization of the integers from [0, MAX_INT], encoded as
     // a list of integers where new buckets begin (excluding 0 which is
     // implicitly included).
+    // It should consist of strictly increasing positive integers without repetitions.
```
apasel422, line 165: Similar to the `end_times` field, what is the benefit of requiring this? It seems like more work for the implementation to enforce the strict-increasing part when the bucket endpoints are already effectively a set.

flexible_event_config.md
apasel422: The values `[1, 2, 3]` and `[3, 1, 2, 3]` are effectively the same. They say that the report window end times are 1, 2, and 3, regardless of their order within the JSON list. There's no reason to prevent duplicates or require values to be sorted during parsing.

flexible_event_config.md
yanzhangnyc: Major motivation has been mention in previous comments. Implementation-wise, we need pre-processing anyway, either to process it to reasonable window or just return error.

flexible_event_config.md
apasel422: We can discuss loosening this requirement separately, but please `s/should/must/` here and below.

flexible_event_config.md
apasel422: > Major motivation

I disagree that the lack of a business use-case for duplicates or unordered values is a major motivation. We don't currently prevent duplicates in other areas where they are irrelevant, e.g. the source's `destination` field or `filter_data`'s list elements.

flexible_event_config.md
yanzhangnyc: The values [1, 2, 3] and [3, 1, 2, 3] are effectively the same --- If we define it the same, we also need to provide extra explanation the detail behavior. I don't think  [1, 2, 3] and [3, 1, 2, 3] are the same. At least the potential confusion is when bucket/window to go 3, will it generate 2 duplicate report or 1 report. 
Generally, I would like the keep rules for the config as strict as possible unless there is valid business use case to provide additional flexibility. Vague/Unconventional input like [3, 1, 2, 3] will create additional corner case and make it difficult for implementation and communication with developer.

apasel422: Should we also prevent duplicates in the trigger spec's `trigger_data` field then?

As I mentioned in https://github.com/WICG/attribution-reporting-api/pull/899#discussion_r1268380645, if you feel strongly about this, please also consider filing an issue for rejecting:

1. Sources with duplicate sites in their `destination` field
2. Sources with duplicate values in their `filter_data` lists
3. Triggers with duplicate values in their `filters` lists
4. Triggers with duplicate values in their `aggregatable_trigger_data` `source_keys` lists

flexible_event_config.md
yanzhangnyc: Although implemented as list, both destination and filter_data are technically hash-table style and each element performs independently. However, end_times and summary_buckets are list, and relationship between the element are the key to it function.

flexible_event_config.md
apasel422, line 146: `without repetition` is redundant with `strictly increasing`

apasel422 (changes requested):
LGTM for now modulo changing `should` to `must`

flexible_event_config.md
yanzhangnyc: Should we also prevent duplicates in the trigger spec's trigger_data field then? -- You are right. I am just drafting that PR. The duplicates in trigger_data is confusing because one trigger_data only need one config. 

For filter, each element work independently and no confusion for duplicated filter, I tend to reject the duplicate but open to any option. 

But back to the essential question, if no valid business use case, is there any benefit to allow vague config?

apasel422: > But back to the essential question, if no valid business use case, is there any benefit to allow vague config?

1. It's more flexible for the implementation
2. It's more flexible for reporting origins

But it's fine for now.

flexible_event_config.md
```diff
@@ -143,24 +143,26 @@ In addition to the parameters that were added in Phase 1, we will add one additi
...
     // value_sum: sum of the value of triggers within a window
     // The summary is reported as an index into a bucketization scheme.
     // Defaults to "count"
-    "summary_window_operator": <one of "count" or "value_sum">
+    "summary_window_operator": <one of "count" or "value_sum">, 
 
     // Represents a bucketization of the integers from [0, MAX_INT], encoded as
     // a list of integers where new buckets begin (excluding 0 which is
     // implicitly included).
+    // It must consist of strictly increasing positive integerss.
```
apasel422, line 165:
```suggestion
    // It must consist of strictly increasing positive integers. 
```

apasel422 (approved):

Issue #769, opened 2023-04-20
Consider having browser provided filters for lookback window


johnivdel:
The current filtering mechanism in the API supports custom trigger-based lookback windows using the list intersection logic. For example, if you have conversion type = 0 with a 7 day lookback, and conversion_type = 1 with a 3 day lookback, you can encode this by doing: 

source
```
{
...
  "filter_data": {"lookback": "day5"}
}
```

trigger
```
{
   ...
  "event_trigger_data": [
     {
     "trigger_data": 0
     "filters": {"lookback": ["day1", "day2", "day3", "day4", "day5", "day6", "day7"]}
     },
     {
     "trigger_data": 0
     "filters": {"lookback": ["day5", "day6", "day7"]}
     }
]
}
```

While this works, it is not a dense encoding, and supporting granularity beyond days would require a very large increase in size.

The browser could provide a custom filter which would have it's own behavior:

source -> no changes
trigger
```
{
   ...
  "event_trigger_data": [
     {
     "trigger_data": 0
     "lookback": 7 // days, but could support a finer unit
     },
     {
     "trigger_data": 0
     "lookback": 3 
     }
]
}
```
The browser would include the lookback field and perform it's own comparison using "trigger time - source time" when deciding if a filter matches.

This is more flexible, but introduces more complexity to the filtering logic / registration schema.

i-bansal:
The above proposed solution would not work with OR support that was added for filters in [issue](https://github.com/WICG/attribution-reporting-api/issues/615) as described in the example below without breaking the OR into individual event blocks. 

Before lookback support -
```html
{
 ...
  "event_trigger_data": [
    {
      "trigger_data": "2",
       "filters": [
           {"source_type": ["view"], "lookback": ["day1", "day2"]},  // OR
           {"source_type": ["navigation"], "lookback": ["day1", "day2", "day3", "day4", "day5"]}
        ]
    },
]}
``` 
After lookback support -
```html
  {
   ...
  "event_trigger_data": [
     {
     "trigger_data": 2
     "lookback": 2
     "filters": [{"source_type": ["view"]]}]
     },
     {
     "trigger_data": 2
     "lookback": 5
     "filters": [{"source_type": ["navigation"]]}]
     },
]}
```

apasel422:
[...]

PR #914, opened 2023-07-25
Specify lookback window

agarant:
Fix https://github.com/WICG/attribution-reporting-api/issues/769

Support lookback window by adding an optional reserved keyword "_lookback_window" to trigger's filters.

When set, the field is parsed as a duration in seconds.

To decide if a filter matches, we first check the lookback window. If available, the duration since the source was registered must be smaller or equal to the lookback window duration.

**Backward incompatibility**:  By introducing a new reserved keyword, we break source registrations which are currently using the "_lookback_window" key.

***
<a href="https://pr-preview.s3.amazonaws.com/agarant/attribution-reporting-api/pull/914.html" title="Last updated on Aug 2, 2023, 5:43 PM UTC (da3c1bf)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/attribution-reporting-api/914/cad9329...agarant:da3c1bf.html" title="Last updated on Aug 2, 2023, 5:43 PM UTC (da3c1bf)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -713,7 +713,7 @@ A <dfn>filter config</dfn> is a [=struct=] with the following items:
 : <dfn>map</dfn>
 :: A [=filter map=].
 : <dfn>lookback window</dfn>
-:: A [=duration=].
+:: Null or a positive [=duration=].
 
 </dl>
 
@@ -1408,12 +1408,10 @@ To <dfn>parse filter config</dfn> given a |value|:
 
 1. If |value| is not a [=map=], return null.
 1. Let |lookbackWindow| be null.
-1. [=map/iterate|For each=] |filter| → |data| of |value|:
-    1. If |filter| is not equal to "`lookback-window`", [=iteration/continue=].
-    1. If |value| is not a non-negative integer, return null.
-    1. Set |lookbackWindow| to |value|.
-    1. [=map/Remove=] |value|[|filter|].
-    1. [=iteration/Break=].
+1. If |value|["`lookback-window`"] [=map/exists=]:
+    1. If |value|["`lookback_window`""] is not a non-negative integer, return null.
+    1. Let |lookbackWindow| be |value|["`lookback_window`""].
+    1. [=map/Remove=] |value|["`lookback_window`""].
 1. Let |map| be the result of running [=parse filter values=] with |value|.
 1. If |map| is null, return null.
 1. Let |filter| be a [=filter config=] with the items:
@@ -2465,12 +2463,12 @@ To <dfn>match [=filter values=] with negation</dfn> given a [=filter value=] |a|
 1. Return true.
 
 To <dfn>match an attribution source's filter data against a filter config</dfn> given an
-[=attribution source=] |source|, a [=filter config=] |filter|, and a [=boolean=]
+[=attribution source=] |source|, a [=filter config=] |filter|, a [=moment=] moment, and a [=boolean=]
 <dfn for="match an attribution source's filter data against a filter config"><var>isNegated</var></dfn>:
 
 1. Let |lookbackWindow| be |filter|'s [=filter config/lookback window=].
 1. If |lookbackWindow| is not null:
-    1. If [=current wall time=] - |source|'s [=attribution source/source time=] is greater than |lookbackWindow|:
+    1. If |moment| - |source|'s [=attribution source/source time=] is greater than |lookbackWindow|:
         1. If |isNegated|, return true. Otherwise, return false. 
 
 1. Let |filterMap| be |filter|'s [=filter config/map=].
@@ -2491,21 +2489,21 @@ To <dfn>match an attribution source's filter data against a filter config</dfn> 
 1. Return true.
 
 To <dfn>match an attribution source's filter data against filters</dfn> given an
-[=attribution source=] |source|, a [=list=] of [=filter configs=] |filters|, and a [=boolean=]
+[=attribution source=] |source|, a [=list=] of [=filter configs=] |filters|, a [=moment=] |moment|, and a [=boolean=]
 <dfn for="match an attribution source's filter data against filters"><var>isNegated</var></dfn>:
 
 1. If |filters| [=list/is empty=], return true.
 1. [=list/iterate|For each=] |filter| of |filters|:
-    1. If the result of running [=match an attribution source's filter data against a filter config=] with |source|, |filter| and |isNegated| is true, return true.
+    1. If the result of running [=match an attribution source's filter data against a filter config=] with |source|, |filter|, |moment|, and |isNegated| is true, return true.
 1. Return false.
 
 To <dfn>match an attribution source's filter data against filters and negated filters</dfn> given an
-[=attribution source=] |source|, a [=list=] of [=filter configs=] |filters|, and a [=list=] of [=filter configs=] |notFilters|:
+[=attribution source=] |source|, a [=list=] of [=filter configs=] |filters|, a [=list=] of [=filter configs=] |notFilters|, and a [=moment=] |moment|:
 
 1. If the result of running [=match an attribution source's filter data against filters=] with
-    |source|, |filters|, and [=match an attribution source's filter data against filters/isNegated=] set to false is false, return false.
+    |source|, |filters|, |moment|, and [=match an attribution source's filter data against filters/isNegated=] set to false is false, return false.
 1. If the result of running [=match an attribution source's filter data against filters=] with
-    |source|, |notFilters|, and [=match an attribution source's filter data against filters/isNegated=] set to true is false, return false.
+    |source|, |notFilters|, |moment|, and [=match an attribution source's filter data against filters/isNegated=] set to true is false, return false.
 1. Return true.
 
 <h3 dfn id="should-block-attribution-for-attribution-limit">Should attribution be blocked by attribution rate limit</h3>
@@ -2576,9 +2574,10 @@ To <dfn>create [=aggregatable contributions=]</dfn> given an [=attribution sourc
 1. Let |aggregationKeys| be the result of [=map/clone|cloning=] |source|'s [=attribution source/aggregation keys=].
 1. [=list/iterate|For each=] |triggerData| of |trigger|'s [=attribution trigger/aggregatable trigger data=]:
     1. If the result of running [=match an attribution source's filter data against filters and negated filters=] with
-        |source|, |triggerData|'s [=aggregatable trigger data/filters=], and
-        |triggerData|'s [=aggregatable trigger data/negated filters=]
-        is false, [=iteration/continue=].
+        |source|, |triggerData|'s [=aggregatable trigger data/filters=],
+        |triggerData|'s [=aggregatable trigger data/negated filters=], and
+        |trigger|'s [=attribution trigger/trigger time=]
+        is false, [=iteration/continue=]:
     1. [=set/iterate|For each=] |sourceKey| of |triggerData|'s [=aggregatable trigger data/source keys=]:
         1. If |aggregationKeys|[|sourceKey|] does not [=map/exist=], [=iteration/continue=].
         1. [=map/Set=] |aggregationKeys|[|sourceKey|] to |aggregationKeys|[|sourceKey|] bitwise-OR |triggerData|'s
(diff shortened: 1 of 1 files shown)


apasel422 (commented):
Please reference #769 in the PR description.

Also, could you provide a high-level explanation of what types of operations this supports and how it's used syntactically?

index.bs
```diff
@@ -1394,21 +1404,40 @@ To <dfn>parse filter data</dfn> given a |value|:
 Issue: Determine whether to limit [=string/length=] or
 [=string/code point length=] for |filter| and |d| above.
 
+To <dfn>parse filter config</dfn> given a |value|:
+
+1. If |value| is not a [=map=], return null.
+1. Let |lookbackWindow| be null.
+1. [=map/iterate|For each=] |filter| → |data| of |value|:
+    1. If |filter| is not equal to "`lookback-window`", [=iteration/continue=].
```
apasel422, line 1412: The `-` here is inconsistent with the other special filter key `source_type`.

index.bs
```diff
@@ -1394,21 +1404,40 @@ To <dfn>parse filter data</dfn> given a |value|:
...
+    1. Set |lookbackWindow| to |value|.
+    1. [=map/Remove=] |value|[|filter|].
+    1. [=iteration/Break=].
+1. Let |map| be the result of running [=parse filter values=] with |value|.
+1. If |map| is null, return null.
+1. Let |filter| be a [=filter config=] with the items:
+    : [=filter config/map=]
+    :: |map|
+    : [=filter config/lookback window=]
+    :: |lookbackWindow|
```
apasel422, line 1423: Shouldn't this be something like `|lookbackWindow| seconds`?

apasel422 (changes requested):

index.bs
```diff
@@ -2435,12 +2464,18 @@ To <dfn>match [=filter values=] with negation</dfn> given a [=filter value=] |a|
 1. If |i| is not [=set/is empty|empty=], then return false.
 1. Return true.
 
-To <dfn>match an attribution source's filter data against a filter map</dfn> given an
-[=attribution source=] |source|, a [=filter map=] |filter|, and a [=boolean=]
-<dfn for="match an attribution source's filter data against a filter map"><var>isNegated</var></dfn>:
+To <dfn>match an attribution source's filter data against a filter config</dfn> given an
```
apasel422, line 2467: Do we want to distinguish the existing filter-match failure from lookback failure in the verbose debug reports?

As written, both types will use the existing `trigger-no-matching-filter-data` / `trigger-event-no-matching-configurations` types.

dmdabbs: Clearer tracing debug signals are helpful.

index.bs
```diff
@@ -1394,21 +1404,38 @@ To <dfn>parse filter data</dfn> given a |value|:
 Issue: Determine whether to limit [=string/length=] or
 [=string/code point length=] for |filter| and |d| above.
 
+To <dfn>parse filter config</dfn> given a |value|:
+
+1. If |value| is not a [=map=], return null.
+1. Let |lookbackWindow| be null.
+1. If |value|["`lookback-window`"] [=map/exists=]:
```
apasel422, line 1411:
```suggestion
1. If |value|["`lookback_window`"] [=map/exists=]:
```

index.bs
```diff
@@ -1394,21 +1404,38 @@ To <dfn>parse filter data</dfn> given a |value|:
...
+    1. If |value|["`lookback_window`""] is not a non-negative integer, return null.
+    1. Let |lookbackWindow| be |value|["`lookback_window`""].
```
apasel422, line 1413:
```suggestion
    1. Set |lookbackWindow| to |value|["`lookback_window`"].
```

index.bs
```diff
@@ -1394,21 +1404,38 @@ To <dfn>parse filter data</dfn> given a |value|:
 Issue: Determine whether to limit [=string/length=] or
 [=string/code point length=] for |filter| and |d| above.
 
+To <dfn>parse filter config</dfn> given a |value|:
+
+1. If |value| is not a [=map=], return null.
+1. Let |lookbackWindow| be null.
+1. If |value|["`lookback-window`"] [=map/exists=]:
+    1. If |value|["`lookback_window`""] is not a non-negative integer, return null.
```
apasel422, line 1412:
```suggestion
    1. If |value|["`lookback_window`"] is not a non-negative integer, return null.
```

index.bs
```diff
@@ -2435,12 +2462,18 @@ To <dfn>match [=filter values=] with negation</dfn> given a [=filter value=] |a|
 1. If |i| is not [=set/is empty|empty=], then return false.
 1. Return true.
 
-To <dfn>match an attribution source's filter data against a filter map</dfn> given an
-[=attribution source=] |source|, a [=filter map=] |filter|, and a [=boolean=]
-<dfn for="match an attribution source's filter data against a filter map"><var>isNegated</var></dfn>:
+To <dfn>match an attribution source's filter data against a filter config</dfn> given an
+[=attribution source=] |source|, a [=filter config=] |filter|, a [=moment=] moment, and a [=boolean=]
```
apasel422, line 2466:
```suggestion
[=attribution source=] |source|, a [=filter config=] |filter|, a [=moment=] |moment|, and a [=boolean=]
```

apasel422 (changes requested):

index.bs
```diff
@@ -1394,21 +1404,38 @@ To <dfn>parse filter data</dfn> given a |value|:
...
 [=string/code point length=] for |filter| and |d| above.
 
+To <dfn>parse filter config</dfn> given a |value|:
+
+1. If |value| is not a [=map=], return null.
+1. Let |lookbackWindow| be null.
+1. If |value|["`lookback_window`"] [=map/exists=]:
+    1. If |value|["`lookback_window`"] is not a non-negative integer, return null.
+    1. Set |lookbackWindow| to the [=duration=] of |value|["`lookback_window`"] seconds.
+    1. [=map/Remove=] |value|["`lookback_window`""].
```
apasel422, line 1414: This is elegant, but we have to decide if we're OK breaking any registrations that are already using the `lookback_window` key.

Related to that, we currently fail source registrations if their filter data contains a `source_type` key. Should we do something similar here?

index.bs
```diff
@@ -2435,12 +2462,18 @@ To <dfn>match [=filter values=] with negation</dfn> given a [=filter value=] |a|
...
-To <dfn>match an attribution source's filter data against a filter map</dfn> given an
-[=attribution source=] |source|, a [=filter map=] |filter|, and a [=boolean=]
-<dfn for="match an attribution source's filter data against a filter map"><var>isNegated</var></dfn>:
+To <dfn>match an attribution source's filter data against a filter config</dfn> given an
+[=attribution source=] |source|, a [=filter config=] |filter|, a [=moment=] |moment|, and a [=boolean=]
+<dfn for="match an attribution source's filter data against a filter config"><var>isNegated</var></dfn>:
+
+1. Let |lookbackWindow| be |filter|'s [=filter config/lookback window=].
+1. If |lookbackWindow| is not null:
+    1. If |moment| - |source|'s [=attribution source/source time=] is greater than |lookbackWindow|:
```
apasel422, line 2471: Let's use https://w3c.github.io/hr-time/#dfn-duration-from, as we do elsewhere in the spec, instead of `-` here.

index.bs
```diff
@@ -1394,21 +1404,38 @@ To <dfn>parse filter data</dfn> given a |value|:
 Issue: Determine whether to limit [=string/length=] or
 [=string/code point length=] for |filter| and |d| above.
 
+To <dfn>parse filter config</dfn> given a |value|:
+
+1. If |value| is not a [=map=], return null.
+1. Let |lookbackWindow| be null.
+1. If |value|["`lookback_window`"] [=map/exists=]:
+    1. If |value|["`lookback_window`"] is not a non-negative integer, return null.
```
apasel422, line 1412: Is there a reason to allow 0 here? Should we require the integer to be positive instead? As written, this is incompatible with the `filter config` definition, which says the duration is positive.

index.bs
apasel422, line 1414:
```suggestion
    1. [=map/Remove=] |value|["`lookback_window`"].
```

apasel422 (changes requested):

index.bs
agarant: If we do break registrations that are using the `lookback_window` key, it does makes sense to fail source registrations if their filter data contains a `lookback_window` key, it will prevent confusion down the line.

However, whether this is fine to do is a good question. Was there a similar decision when `source_type` was added ?

apasel422: IIRC `source_type` was added at the same time as filtering in general, so we were able to prevent that from the start, at least in the spec (#468).

index.bs
agarant: One issue with reporting different failure types is that both could occur at the same time. e.g. First filter is outside the lookback window and the second has no match. We could:

1. Keep reporting one type of failure that encompasses both types.
2. Report a different type of failure by using a heuristic (lookback first, first seen, etc) to pick the type of failure to report on when both are applicable.
3. Report multiple failure types in a single verbose debug report.

I think 2 would be a reasonable option to give visibility on lookback window failures without the complexity associated with supporting multiple failure types.

index.bs
agarant: I added a note in the description on the backward incompatibility. We can also mitigate the impact for currently stored sources by having a transition release that supports test lookback_window key.

index.bs
vikassahu29: I don't think we should introduce new failure types as it is still a `filter-data` & `no-matching-configuration` failure along with being a `lookback_window` failure or a `source_type` failure

Maybe we can provide additional info about the failure along with the type i.e. sub-type/error_info within the verbose report.

index.bs
```diff
@@ -707,6 +707,16 @@ A <dfn>filter value</dfn> is an [=ordered set=] of [=strings=].
 A <dfn>filter map</dfn> is an [=ordered map=] whose [=map/key|keys=] are [=strings=] and whose
 [=map/value|values=] are [=filter values=].
 
+A <dfn>filter config</dfn> is a [=struct=] with the following items:
+
+<dl dfn-for="filter config">
+: <dfn>map</dfn>
+:: A [=filter map=].
+: <dfn>lookback window</dfn>
```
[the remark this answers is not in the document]
apasel422: Yes, this a spec structure.

index.bs
apasel422: Based on https://github.com/WICG/attribution-reporting-api/pull/914#discussion_r1275459945 and https://github.com/WICG/attribution-reporting-api/pull/914#discussion_r1275528357 I think it's fine to use the existing debug report types. In today's debug reports, we don't say which filter field(s) failed to match, so if lookback is being implemented as a special type of filter field, I think it's consistent to keep things as they are.

apasel422 (approved):
Spec changes LGTM, but I'm withholding any judgment on the compatibility issue.

vikassahu29 (approved):

index.bs
```diff
@@ -1394,21 +1404,38 @@ To <dfn>parse filter data</dfn> given a |value|:
 Issue: Determine whether to limit [=string/length=] or
 [=string/code point length=] for |filter| and |d| above.
 
+To <dfn>parse filter config</dfn> given a |value|:
+
+1. If |value| is not a [=map=], return null.
+1. Let |lookbackWindow| be null.
+1. If |value|["`lookback_window`"] [=map/exists=]:
```
johnivdel, line 1411: It's worth noting that in the Protected Audiences extending private aggregation explainer, they prefix "special" keys with "reserved." (https://github.com/WICG/turtledove/blob/main/FLEDGE_extended_PA_reporting.md)

it might make sense to do that here, and would probably lead to less compat issues (although  this is probably marginal anyway).

index.bs
```diff
@@ -1394,21 +1404,38 @@ To <dfn>parse filter data</dfn> given a |value|:
 Issue: Determine whether to limit [=string/length=] or
 [=string/code point length=] for |filter| and |d| above.
 
+To <dfn>parse filter config</dfn> given a |value|:
+
+1. If |value| is not a [=map=], return null.
+1. Let |lookbackWindow| be null.
+1. If |value|["`_lookback_window`"] [=map/exists=]:
```
apasel422, line 1411: Do we want to take this opportunity to reserve all keys beginning with `_`, or at least add an issue in the spec to do so? That would help simplify future changes similar to this one.

index.bs
agarant: The key was updated with the "_" prefix.

index.bs
agarant: Good idea, I'll add this in a follow up PR. I added **https://github.com/WICG/attribution-reporting-api/issues/930**.

csharrison (approved):

PR #916, opened 2023-07-26
Add separate verbose debug report for event report window not started

### LINKED ISSUE #911: Update debugData to accurately describe case where trigger occurs before source's event-level report window start time
Currently, if a trigger's trigger time falls outside of the matching source's total event-level report window, obtaining the debug data will return "trigger-event-report-window-passed". The name of this value does not encapsulate all cases where this value is returned, specifically if the trigger time is before the report window's start time (https://wicg.github.io/attribution-reporting-api/#ref-for-trigger-debug-data-type-trigger-event-report-window-passed). I propose we either return a new value along the lines of "trigger-event-report-window-not-started" when the specified case occurs or we modify the existing value to be "trigger-event-not-within-report-window" to encapsulate both before and after cases.

apasel422:
Fixes #911

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1059,6 +1059,7 @@ Possible values are:
 <li>"<dfn><code>trigger-event-low-priority</code></dfn>"
 <li>"<dfn><code>trigger-event-no-matching-configurations</code></dfn>"
 <li>"<dfn><code>trigger-event-noise</code></dfn>"
+<li>"<dfn><code>trigger-event-report-window-not-started</code></dfn>"
 <li>"<dfn><code>trigger-event-report-window-passed</code></dfn>"
 <li>"<dfn><code>trigger-event-storage-limit</code></dfn>"
 <li>"<dfn><code>trigger-no-matching-source</code></dfn>"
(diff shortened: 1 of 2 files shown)


index.bs
```diff
@@ -2914,8 +2925,11 @@ To <dfn noexport>trigger attribution</dfn> given an [=attribution trigger=] |tri
 To <dfn>check whether a moment falls within a window</dfn> given a [=moment=] |moment| and
 a [=report window=] |window|:
 
-1. Return whether |moment| is greater than or equal to [=report window/start=]
-    and |moment| is less than |window|'s [=report window/end=].
+1. If |moment| is less than |window|'s [=report window/start=], return
```
csharrison, line 2928: optional: What do you think about some alternatives like `falls before`, `falls within`, `falls after`? I think `passed` and `active` are a bit ambiguous, especially `passed` which could indicate success.

apasel422: I was trying to make the names match the debug report strings, but I agree these are easier to understand. Done.

csharrison (approved):

yanzhangnyc (approved):
Since the end of the window debug report has already added, I have no concern to add one more debug report before the report window.

PR #917, opened 2023-07-26
Port flexible_event_privacy.py to TypeScript

apasel422:
This will allow us to incorporate this information into the header validator.


flexible-event/index.js
```diff
@@ -0,0 +1,253 @@
...
+  console.log(`Information gain: ${infoGain} bits`)
+  console.log(`Flip percent: ${100 * flipProb}%`)
+
+  const infoGainDefault = maxInformationGain(
+        numFlexibleStates(DEFAULT_CONFIG[sourceType]), epsilon)
+
+  if (infoGain > infoGainDefault) {
+      const [newEps, flipProb] = epsilonToBoundInfoGainAndDp(numStates, infoGainDefault, epsilon)
+      console.log(
+          `WARNING: info gain of ${infoGain} > ${infoGainDefault} for ${sourceType} sources. Would require a ${100 * flipProb}% flip chance (effective epsilon = ${newEps}) to resolve.`)
```
csharrison, line 154: The output is a lot nicer to read if these are capped to a certain # of decimal points using e.g. `toFixed(<decimal places>)`

flexible-event/index.js
```diff
@@ -0,0 +1,253 @@
...
+
+  return {maxEventLevelReports, perTriggerDataConfigs}
+}
+
+const optionDefs = [
+  {
+    name: 'max_event_level_reports',
+    alias: 'm',
+    type: Number,
+    defaultValue: 20,
```
csharrison, line 206: The default should be based on the source type

csharrison (changes requested):

flexible-event/index.js
```diff
@@ -0,0 +1,253 @@
+const commandLineArgs = require('command-line-args')
```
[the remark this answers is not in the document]
apasel422: That would be fine with me, but @maudnals configured our other JS code to not use them.

flexible-event/index.js
apasel422: The original code used 20, no?

csharrison: Ah yeah, sorry this is fine.

flexible-event/index.js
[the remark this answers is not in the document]
csharrison: ack, that's fine.

csharrison (approved):

csharrison (approved):

PR #926, opened 2023-08-01
Deploy header validator using webpack

apasel422:
This bundles all NPM dependencies into a single JS file, which will make it easier to deploy future integration with the flex-event privacy scripts, and to replace JS with TypeScript.


dmdabbs:
Suggestion: post a link to the deployed header validator (https://wicg.github.io/attribution-reporting-api/validate-headers) in the README in that [folder](https://github.com/WICG/attribution-reporting-api/tree/main/header-validator).

csharrison (approved):

PR #928, opened 2023-08-02
Remove JSON schemas

apasel422:
Unfortunately, JSON schema is not powerful enough to fully validate our schemas with respect to the Full Flex proposal, and it is tedious to maintain them alongside the more powerful custom JS-based validator.

The exact caveats for Full Flex are:

- Inability to ensure that trigger data values across trigger specs are disjoint
- Inability to ensure that summary buckets and report windows are increasing
- Inability to ensure that the first report window is greater than the start time

Additionally, we will likely want to retain the custom validator going forward so that its output can be fed into the library that performs privacy-related computations for Flex and shown in the web-based header validator.


csharrison:
Did you evaluate whether its possible to keep the schema for simple checks and the validator depends on the schema to implement purely the sophisticated checks that cannot be done with JSON schema?

apasel422:
> Did you evaluate whether its possible to keep the schema for simple checks and the validator depends on the schema to implement purely the sophisticated checks that cannot be done with JSON schema?

That is possible, yes, but I think it will complicate integration with other tooling, e.g. hooking into the flex privacy library. My plan is to have something like this once the validator has been converted to TypeScript:

```typescript
type ParsedSource = { ... }

function parseSource(json: string, userAgentConfig: Config): ParsedSource|Error;
```

so the header validator will be able to deeply inspect the object in one pass and store the parsed values (including effective values computed using `userAgentConfig` in the output structure), and then other code like the flex privacy library can directly consume those well-typed structures instead of doing their own parsing. This isn't really possible using the JSON schema.

PR #934, opened 2023-08-03
Simplify OS-header validation and fix type issues

apasel422:
This header supports only one optional parameter, so the abstractions were increasing complexity for no benefit.

### DIFF
--- a/header-validator/src/validate-os.ts
+++ b/header-validator/src/validate-os.ts
@@ -1,21 +1,14 @@
 import { Issue, ValidationResult } from './issue'
 const { parseList } = require('structured-headers')
 
-function validateString(item): string {
+function validateURL(item: any): string|undefined {
   if (typeof item !== 'string') {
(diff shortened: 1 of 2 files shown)


header-validator/src/validate-os.ts
```diff
@@ -31,62 +24,36 @@ function validateURL(urlStr: string): string {
...
-      const err = check(params.get(param))
-      if (err) {
-        errors.push({ msg: err, path: [i, param] })
+    for (const [key, value] of params) {
+      if (key === 'debug-reporting') {
+        if (typeof value !== 'boolean') {
+          result.errors.push({ msg: 'must be a boolean', path: [i, key] })
+        }
+      } else {
+        result.warnings.push({ msg: 'unknown parameter', path: [i, key] })
```
agarant, line 53: only the 'debug-reporting' property is expected ?

agarant (approved):

PR #941, opened 2023-08-08
Lookback window: explainer & validator

agarant:
Update the explainer and header validator according to spec changes introduced in #914.

### DIFF
--- a/EVENT.md
+++ b/EVENT.md
@@ -577,12 +577,13 @@ trigger is used.
 Lookback window is supported with an optional reserved keyword
 `_lookback_window` which can be added to trigger's filters.
 
-Unlike other filter keys whose values must be a list of strings. The
-`_lookback_window` key value must be a positive integer which represents a
+Unlike other filter keys whose values must be a list of strings, the
+`_lookback_window` value must be a positive integer that represents a positive
 duration in seconds.
 
 When available on a filter, the duration since the source was registered must be
-smaller or equal to the parsed lookback window duration for the filter to match.
+less than or equal to the parsed lookback window duration for the filter to
+match.
 
 ### Optional: transitional debugging reports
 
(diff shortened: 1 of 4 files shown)


EVENT.md
```diff
@@ -569,6 +572,18 @@ will be created.
 If the filters match for multiple event triggers, the first matching event
 trigger is used.
 
+#### Lookback window
+
+Lookback window is supported with an optional reserved keyword
+`_lookback_window` which can be added to trigger's filters.
+
+Unlike other filter keys whose values must be a list of strings. The
```
apasel422, line 580:
```suggestion
Unlike other filter keys whose values must be a list of strings, the
```

EVENT.md
```diff
@@ -569,6 +572,18 @@ will be created.
...
+`_lookback_window` key value must be a positive integer which represents a
```
apasel422, line 581:
```suggestion
`_lookback_window` value must be a positive integer that represents a positive
```

EVENT.md
```diff
@@ -569,6 +572,18 @@ will be created.
...
+duration in seconds.
+
+When available on a filter, the duration since the source was registered must be
+smaller or equal to the parsed lookback window duration for the filter to match.
```
apasel422, line 585:
```suggestion
less than or equal to the parsed lookback window duration for the filter to match.
```

header-validator/src/validate-json.ts
```diff
@@ -296,12 +296,21 @@ const filterData = () =>
       state.error('is prohibited because it is implicitly set')
       return
     }
+    if (filter === '_lookback_window') {
+      state.error('is prohibited because it would be ignored')
```
apasel422, line 300: Is there a reason to use a different message for this compared to `source_type`?

apasel422 (changes requested):

header-validator/src/validate-json.ts
agarant: The idea was that we are not really comparing to a generated value, but we could probably consider it to be implicitly set. I updated.

apasel422 (approved):

akashnadan (commented):
lgtm

PR #951, opened 2023-08-15
Update chromium-params.md - information gain bounds

akashnadan:
Adding navigation and event information gain bounds

### DIFF
--- a/params/chromium-params.md
+++ b/params/chromium-params.md
@@ -29,8 +29,8 @@ Chromium's implementation assigns the following values:
 | [Max attributions per rate-limit window][] | [100][max attributions per rate-limit window value] |
 | [Allowed aggregatable budget per source][] | [65536][allowed aggregatable budget per source value] |
 | [Randomized aggregatable report delay][] | [10 minutes][randomized aggregatable report delay value] |
-| Max information gain for navigations | 11.46 bits |
-| Max information gain for events | 6.5 bits |
+| [Max information gain for navigations][] | [11.46 bits][max information gain for navigations] |
+| [Max information gain for events][] | [6.5 bits][max information gain for events] |
 
 [Max aggregation keys per attribution]: https://wicg.github.io/attribution-reporting-api/#max-aggregation-keys-per-attribution
 [max aggregation keys per attribution value]: https://source.chromium.org/chromium/chromium/src/+/refs/heads/main:components/attribution_reporting/constants.h;l=19;drc=b646f894a92491033bde5d1e75aba6f44c524f0e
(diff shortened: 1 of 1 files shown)


params/chromium-params.md
```diff
@@ -29,6 +29,8 @@ Chromium's implementation assigns the following values:
 | [Max attributions per rate-limit window][] | [100][max attributions per rate-limit window value] |
 | [Allowed aggregatable budget per source][] | [65536][allowed aggregatable budget per source value] |
 | [Randomized aggregatable report delay][] | [10 minutes][randomized aggregatable report delay value] |
+| [Max information gain for navigations][] | [11.46 bits][max information gain for navigations value] |
+| [Max information gain for events][] | [6.5 bits][max information gain for events value] |
```
apasel422, line 33: It might be clearer to say `for navigation sources` and `for event sources`.

params/chromium-params.md
```diff
@@ -75,3 +77,7 @@ Chromium's implementation assigns the following values:
...
 [Randomized aggregatable report delay]: https://wicg.github.io/attribution-reporting-api/#randomized-aggregatable-report-delay
 [randomized aggregatable report delay value]: https://source.chromium.org/chromium/chromium/src/+/main:content/browser/attribution_reporting/attribution_config.h;l=106;drc=3733a639d724a4353463a872605119d11a1e4d37
+[Max information gain for navigations]: https://wicg.github.io/attribution-reporting-api/#max-event-level-channel-capacity-per-source
+[max information gain for navigations value]: https://source.chromium.org/chromium/chromium/src/+/main:content/browser/attribution_reporting/attribution_config.h;l=113;drc=c8a629d7be311323904c523d33693af8f52d17e9
+[Max information gain for events]: https://wicg.github.io/attribution-reporting-api/#max-event-level-channel-capacity-per-source
+[max information gain for events value]: https://source.chromium.org/chromium/chromium/src/+/main:content/browser/attribution_reporting/attribution_config.h;l=114;drc=c8a629d7be311323904c523d33693af8f52d17e9
```
apasel422, line 83: This link is not correct, as it still shows the old value. This will have to be updated for the corresponding Chromium CL.

apasel422 (changes requested):

params/chromium-params.md
csharrison: Would you prefer @apasel422 to leave it as 1.58 for now and update when https://chromium-review.googlesource.com/c/chromium/src/+/4780891 lands?

apasel422: Either way. It's just weird to put the new value as the text of the link but then have it go to a specific commit line in Code Search with the old value.

csharrison: The CL landed so this should be OK now. I think we need to either remove the `drc` param or update it though.

params/chromium-params.md
akashnadan: edited

params/chromium-params.md
```diff
@@ -75,3 +77,7 @@ Chromium's implementation assigns the following values:
 [allowed aggregatable budget per source value]: https://source.chromium.org/chromium/chromium/src/+/main:content/browser/attribution_reporting/attribution_config.h;l=97;drc=3733a639d724a4353463a872605119d11a1e4d37
 [Randomized aggregatable report delay]: https://wicg.github.io/attribution-reporting-api/#randomized-aggregatable-report-delay
 [randomized aggregatable report delay value]: https://source.chromium.org/chromium/chromium/src/+/main:content/browser/attribution_reporting/attribution_config.h;l=106;drc=3733a639d724a4353463a872605119d11a1e4d37
+[Max information gain for navigation sources]: https://wicg.github.io/attribution-reporting-api/#max-event-level-channel-capacity-per-source
+[max information gain for navigations value]: https://source.chromium.org/chromium/chromium/src/+/main:content/browser/attribution_reporting/attribution_config.h;l=113
+[Max information gain for event sources]: https://wicg.github.io/attribution-reporting-api/#max-event-level-channel-capacity-per-source
+[max information gain for events value]: https://source.chromium.org/chromium/chromium/src/+/main:content/browser/attribution_reporting/attribution_config.h;l=114;bpv=0;bpt=1
```
apasel422, line 83: Please use links at a particular commit so that the value in Code Search is consistent with the value listed in this table.

params/chromium-params.md
akashnadan: edited

params/chromium-params.md
akashnadan: edited

apasel422 (approved):

PR #952, opened 2023-08-15
Replace State class with reusable Context class

apasel422:
In addition to benefiting the structured-header validators, this will make it easier to propagate data such as vendor-specific values to the JSON validators in subsequent changes.


header-validator/src/validate-os.ts
```diff
@@ -1,16 +1,18 @@
-import { Issue, ValidationResult } from './issue'
-const { parseList } = require('structured-headers')
+import { Context, ValidationResult } from './context'
+import { InnerList, Item, parseList } from 'structured-headers'
 
-function validateURL(item: any): string|undefined {
-  if (typeof item !== 'string') {
-    return 'must be a string'
+function validateURL(ctx: Context, member: InnerList|Item): void {
+  if (typeof member[0] !== 'string') {
```
agarant, line 5: Might be useful to spread the values in the member tuple to document what they represent.

agarant, line 5: Might be useful to spread the values in the member tuple to document what they represent.

agarant (approved):
nice!

header-validator/src/validate-os.ts
apasel422: Ack. It's a bit difficult to do so because of the weird typing there, but will pursue in a followup.

PR #958, opened 2023-08-17
Lookback window within negated filter

agarant:
Enhances #914 

At a high level, the currently spec logic is the following:
```
Negated filter true = (no lookback_window || outside lookback_window) || (has empty intersection for all matching keys)
```

This behavior is incoherent with other filtering logic, since as soon as a source is outside of the lookback window, the negated filter would match even if some matching keys have non-empty intersections. It also means that it can still match even if outside of the lookback window so long as it has empty intersection for all matching keys.

This PR updates the logic to the following:
```
Negated filter true = (no lookback_window || outside lookback_window)  &&  (has empty intersection for all matching keys)
```

This is more coherent and being able to match when outside of the lookback window adds capabilities to the API.

***
<a href="https://pr-preview.s3.amazonaws.com/agarant/attribution-reporting-api/pull/958.html" title="Last updated on Aug 25, 2023, 1:05 PM UTC (da2214c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/attribution-reporting-api/958/e3e14b2...agarant:da2214c.html" title="Last updated on Aug 25, 2023, 1:05 PM UTC (da2214c)">Diff</a>

### DIFF
--- a/EVENT.md
+++ b/EVENT.md
@@ -581,9 +581,13 @@ Unlike other filter keys whose values must be a list of strings, the
 `_lookback_window` value must be a positive integer that represents a positive
 duration in seconds.
 
-When available on a filter, the duration since the source was registered must be
-less than or equal to the parsed lookback window duration for the filter to
-match.
+When available on a filter (in `filters`), the duration since the source was
+registered must be less than or equal to the parsed lookback window duration for
+the filter to match. i.e. it must be inside the lookback window.
+
+When available on a negated filter (in `not filters`), the duration since the
+source was registered must be greater than the parsed lookback window duration
+for the filter to match. i.e. it must be outside the lookback window.
 
 ### Optional: transitional debugging reports
 
--- a/EVENT.md
+++ b/EVENT.md
@@ -581,11 +581,11 @@ Unlike other filter keys whose values must be a list of strings, the
 `_lookback_window` value must be a positive integer that represents a positive
 duration in seconds.
 
-When available on a filter (in `filters`), the duration since the source was
+When present on a filter (in `filters`), the duration since the source was
 registered must be less than or equal to the parsed lookback window duration for
 the filter to match, i.e., it must be inside the lookback window.
 
-When available on a negated filter (in `not_filters`), the duration since the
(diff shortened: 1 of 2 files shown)


index.bs
```diff
@@ -2500,7 +2500,8 @@ To <dfn>match an attribution source's filter data against a filter config</dfn>
 1. Let |lookbackWindow| be |filter|'s [=filter config/lookback window=].
 1. If |lookbackWindow| is not null:
     1. If the [=duration from=] |moment| and the |source|'s [=attribution source/source time=] is greater than |lookbackWindow|:
-        1. If |isNegated|, return true. Otherwise, return false.
```
johnivdel, line 2503: Could you clarify this change? 

IIUC we always return false in this case now, is that correct?

agarant: Yes, this is correct. 

When non-negated, we want to check that we are inside the window and the other filters match. We can return early (and false) when we are outside the lookback window.

When negated, we want to check that we are outside and the window and other filters dont match. We can return early (and false) when we are inside the lookback window.

johnivdel (approved):
LGTM

EVENT.md
```diff
@@ -581,9 +581,13 @@ Unlike other filter keys whose values must be a list of strings, the
...
 duration in seconds.
 
-When available on a filter, the duration since the source was registered must be
-less than or equal to the parsed lookback window duration for the filter to
-match.
+When available on a filter (in `filters`), the duration since the source was
+registered must be less than or equal to the parsed lookback window duration for
+the filter to match, i.e., it must be inside the lookback window.
+
+When available on a negated filter (in `not_filters`), the duration since the
```
apasel422, line 588:
```suggestion
When present on a filter (in `filters`), the duration since the source was
registered must be less than or equal to the parsed lookback window duration for
the filter to match, i.e., it must be inside the lookback window.

When present on a negated filter (in `not_filters`), the duration since the
```

apasel422 (approved):

index.bs
agarant: As discussed I added a note in the spec to make this section clearer.

PR #959, opened 2023-08-18
Optionally validate trigger data using Chromium's vendor-specific values

apasel422:


agarant (approved):

csharrison:
I think this is fine but I am not thrilled with yet another place where we need to maintain the Chromium values.

apasel422:
> I think this is fine but I am not thrilled with yet another place where we need to maintain the Chromium values.

I will try to have a single file like `chromium.json` that lists everything in lieu of having both the Markdown file and this one, similar if not fully compatible with [the one used for Chromium's interop tests](https://source.chromium.org/chromium/chromium/src/+/main:content/test/data/attribution_reporting/interop/default_config.json).

csharrison:
Thanks! That would be nice.

PR #970, opened 2023-08-25
Update chromium-params.md - minor value update

akashnadan:


csharrison:
As discussed offline, this is not correct.

PR #977, opened 2023-08-30
Warn about clamped expiry

apasel422:


header-validator/src/validate-json.ts
```diff
@@ -565,6 +568,29 @@ function aggregationKeys(ctx: Context, j: Json): Maybe<Map<string, bigint>> {
...
+  if (n > max) {
+    ctx.warning(`will be clamped to max of ${max}`)
+    return max
+  }
+  return n
+}
+
+function expiry(ctx: Context, j: Json): Maybe<number | bigint> {
+  return legacyDuration(ctx, j).map((n) =>
+    clamp(ctx, n, limits.sourceExpiryRange[0], limits.sourceExpiryRange[1])
```
linnan-github, line 590: Just a note that, for event source, the expiry is rounded first before clamping, but this is fine for now.

linnan-github (approved):

header-validator/src/validate-json.ts
apasel422: Yes, I intend to add the rounding in a followup.

PR #978, opened 2023-08-30
Check that end_times is strictly increasing

apasel422:


giladbarkan-github:
Does this take into account the clamping?

apasel422:
> Does this take into account the clamping?

Good question: It does not. I will wait until #977 is merged and reuse the clamping mechanism from it for that purpose, though it will still be tricky to obtain the spec's `defaultEndDuration` since the validator has no knowledge of the source type, though I intend to add that.

Additionally, I'm confused about [what the Chromium implementation currently does with regard to `defaultEndDuration`](https://source.chromium.org/chromium/chromium/src/+/main:components/attribution_reporting/event_report_windows.cc;l=207-230;drc=b3fc199e1c0ed8a3f6a624a7ff073fc997211498) as well.

CC @ThomasQuesadilla

apasel422:
> Wouldn't only the last and first need to be clamped to both expiry (since event_report_window is disallowed) and min event report window (which is independent of source type?). If anything else _would also_ be clamped, it would no longer be strict.

You're essentially right. I missed the fact that the spec algorithm for the report windows has a source type parameter, but that that parameter is irrelevant for the computation of the last window, which is always the expiry in this case, although it's a bit subtle because the expiry will be rounded if the source type is `event`. But the way this code is organized, there's no easy way to access the expiry here. (I will need to refactor so that the expiry is parsed first, then passed to this function.)

giladbarkan-github:
> > Wouldn't only the last and first need to be clamped to both expiry (since event_report_window is disallowed) and min event report window (which is independent of source type?). If anything else _would also_ be clamped, it would no longer be strict.
> 
> You're essentially right. I missed the fact that the spec algorithm for the report windows has a source type parameter, but that that parameter is irrelevant for the computation of the last window, which is always the expiry in this case, although it's a bit subtle because the expiry will be rounded if the source type is `event`. But the way this code is organized, there's no easy way to access the expiry here. (I will need to refactor so that the expiry is parsed first, then passed to this function.)

Ah, I forgot about the rounding in the case of event, thanks!

ThomasQuesadilla:
> Additionally, I'm confused about [what the Chromium implementation currently does with regard to `defaultEndDuration`](https://source.chromium.org/chromium/chromium/src/+/main:components/attribution_reporting/event_report_windows.cc;l=207-230;drc=b3fc199e1c0ed8a3f6a624a7ff073fc997211498) as well.


Currently, comparisons with `defaultEndDuration` are made [after parsing is finished but before the source is stored](https://source.chromium.org/chromium/chromium/src/+/main:content/browser/attribution_reporting/attribution_storage_sql.cc;l=497-502;drc=80909977e0f2480462cedc4fc82a955c39d12bfd) so that source registration may still fail according to the spec.

PR #979, opened 2023-08-30
Parse dictionaries into well-typed structures

apasel422:
This will make future enhancements like displaying the effective
source registration, given a source type, possible, and will allow for
cross-field validation.

### DIFF
--- a/header-validator/src/validate-json.ts
+++ b/header-validator/src/validate-json.ts
@@ -41,7 +41,10 @@ function struct<T extends object>(
     let ok = true
     for (const prop in fields) {
       let itemOk = false
-      fields[prop](ctx, d).peek((v) => (t[prop] = v))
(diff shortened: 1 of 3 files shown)


header-validator/src/validate-json.ts
```diff
@@ -26,54 +26,94 @@ class Context extends context.Context {
...
+  return (ctx: Context, d: JsonDict): Maybe<T> => {
+    const found: string[] = []
+    let v: Maybe<T> = None
+
+    for (const [key, f] of Object.entries(x)) {
+      const j = d[key]
+      if (j !== undefined) {
+        found.push(key)
+        v = ctx.scope(key, () => f(ctx, j))
+        delete d[key]
```
agarant, line 100: Could be useful to add a comment to indicate why it deletes keys as they are processed (same for field).

agarant (approved):

PR #981, opened 2023-08-31
Add helper for type switches

apasel422:


header-validator/src/validate-json.ts
```diff
@@ -114,32 +114,56 @@ function exclusive<T>(
...
+  }
+  if (Array.isArray(j) && ts.list !== undefined) {
+    return ts.list(ctx, j)
   }
-  ctx.error('must be a string')
+  if (isObject(j) && ts.object !== undefined) {
+    return ts.object(ctx, j)
+  }
+
+  const allowed = Object.keys(ts)
```
agarant, line 146: Not required yet, but do we want to add support for formatting > 2 items in the type switch now ?

header-validator/src/validate-json.ts
```diff
@@ -114,32 +114,56 @@ function exclusive<T>(
...
+    .map((t) => `${t === 'object' ? 'an' : 'a'} ${t}`)
```
agarant, line 147: nit.
```suggestion
    .map((t) => `${t === 'object' ? 'an' : t === 'null' ? '' : 'a'} ${t}`)
```

agarant (approved):

header-validator/src/validate-json.ts
apasel422: It is already supported, e.g. `a list or a string or an object`.

PR #982, opened 2023-09-01
Warn about event-source expiry rounding

apasel422:


header-validator/src/source.test.ts
```diff
@@ -610,6 +610,84 @@ runAll(validateSource, [
       },
     ],
   },
+  {
```
agarant, line 613: Can we add a test for expiry < 86400 ? Unless we have another condition that makes this impossible ?

IIUC this would round down to zero which based on the function name `roundAwayFromZero` might not be desirable.

header-validator/src/validate-json.ts
```diff
@@ -588,10 +588,29 @@ function clamp<N extends bigint | number>(
...
-function expiry(ctx: Context, j: Json): Maybe<number | bigint> {
-  return legacyDuration(ctx, j).map((n) =>
-    clamp(ctx, n, limits.sourceExpiryRange[0], limits.sourceExpiryRange[1])
-  )
+function roundAwayFromZeroToNearestDay(n: number): number {
+  if (n <= 0 || !Number.isInteger(n)) {
+    throw new RangeError()
+  }
+
+  n += secondsPerDay / 2
```
agarant, line 596: nit. we could use a temporary variable instead of mutating the function argument.

header-validator/src/source.test.ts
apasel422: We already have a test for this in the clamping section.

header-validator/src/validate-json.ts
apasel422: Done.

agarant (approved):

PR #985, opened 2023-09-01
Add test cases for start_time

apasel422:


ThomasQuesadilla (approved):
Would it make sense to add a case where [start_time is greater than the default end duration i.e. 30 days, which returns an error](https://wicg.github.io/attribution-reporting-api/#parsing-source-registration:~:text=If%20startDuration%20is%20greater%20than%20defaultEndDuration%2C%20return%20an%20error.)? Otherwise LGTM.

apasel422:
> Would it make sense to add a case where [start_time is greater than the default end duration i.e. 30 days, which returns an error](https://wicg.github.io/attribution-reporting-api/#parsing-source-registration:~:text=If%20startDuration%20is%20greater%20than%20defaultEndDuration%2C%20return%20an%20error.)? Otherwise LGTM.

It would, but we don't have the ability to check that yet. I have a separate PR forthcoming. This PR is just about adding coverage for the checks we currently have.

PR #986, opened 2023-09-01
Fully validate report window(s) with respect to expiry

apasel422:
- aggregatable_report_window is clamped to (1 hour, expiry)
- event_report_window is clamped to (1 hour, expiry)
- event_report_windows.start_time must be <= expiry
- event_report_window.end_times is clamped to (1 hour, expiry)
- event_report_windows.end_times[0] must be > start_time
- event_report_windows.end_times must be strictly increasing

### DIFF
--- a/header-validator/src/validate-json.ts
+++ b/header-validator/src/validate-json.ts
@@ -1,6 +1,6 @@
 import * as psl from 'psl'
 import * as context from './context'
-import { Maybe } from './maybe'
+import { Maybe, Maybeable } from './maybe'
 
 const { None, some } = Maybe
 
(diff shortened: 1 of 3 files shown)


ThomasQuesadilla (approved):

header-validator/src/validate-json.ts
```diff
@@ -29,8 +31,10 @@ class Context extends context.Context {
   }
 }
 
+type FieldFunc<I, T, V> = (ctx: Context, i: I, t: Readonly<Partial<T>>) => V
```
agarant, line 34: [optional idea] Adding this partial `t` everywhere adds a non negligible amount of complexity for the use case it solves.

We're already carrying context. Would it make sense to explicitly set what we need on it (startTime & expiry) and use it directly ?

It's certainly less elegant but might be simpler. We could also enforce the order, i.e., throw an error if we get to a point where we expect the value to be set and it is not.

header-validator/src/validate-json.ts
```diff
@@ -405,14 +456,20 @@ function legacyDuration(ctx: Context, j: Json): Maybe<number | bigint> {
...
+  keepGoing: boolean = true
 ): boolean {
   let ok = true
   for (const [c, j] of js) {
     ctx.scope(c, () => {
       let itemOk = false
       f(ctx, [c, j]).peek(() => (itemOk = true))
-      ok = ok && itemOk
+      if (!itemOk) {
+        if (!keepGoing) {
```
agarant, line 468: Would this only resolve the current callback without actually exiting the top function ?

agarant (approved):

PR #991, opened 2023-09-05
Replace separate Some/None types with Maybe class

apasel422:
This makes it possible to merge map and flatMap, which will simplify additional refactoring in #986.

### DIFF
--- a/header-validator/src/maybe.ts
+++ b/header-validator/src/maybe.ts
@@ -1,29 +1,30 @@
-export type Maybe<T> = typeof None | Some<T>
+export type Maybeable<T> = T | Maybe<T>
 
-export const None = {
-  filter: () => None,
-  map: () => None,
-  flatMap: () => None,
-  peek: () => None,
-}
+export class Maybe<T> {
+  static readonly None: Maybe<any> = new Maybe()
 
-export class Some<T> {
-  constructor(private readonly t: T) {}
+  static some<T>(t: T): Maybe<T> {
+    return new Maybe(t)
+  }
+
+  static flatten<T>(t: Maybeable<T>): Maybe<T> {
+    return t instanceof Maybe ? t : Maybe.some(t)
+  }
+
+  private constructor(private readonly t?: T) {}
 
   filter(f: (t: T) => boolean): Maybe<T> {
-    return f(this.t) ? this : None
+    return this.t === undefined || !f(this.t) ? Maybe.None : this
   }
 
-  map<U>(f: (t: T) => U): Some<U> {
-    return new Some(f(this.t))
+  map<U>(f: (t: T) => Maybeable<U>): Maybe<U> {
+    return this.t === undefined ? Maybe.None : Maybe.flatten(f(this.t))
   }
 
-  flatMap<U>(f: (t: T) => Maybe<U>): Maybe<U> {
-    return f(this.t)
-  }
-
-  peek(f: (t: T) => void): Some<T> {
-    f(this.t)
+  peek(f: (t: T) => void): Maybe<T> {
+    if (this.t !== undefined) {
+      f(this.t)
+    }
     return this
   }
 }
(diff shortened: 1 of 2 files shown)


header-validator/src/maybe.ts
```diff
@@ -1,29 +1,30 @@
...
 
-export class Some<T> {
-  constructor(private readonly t: T) {}
+  static some<T>(t: T): Maybe<T> {
+    return new Maybe(t)
+  }
 
-  filter(f: (t: T) => boolean): Maybe<T> {
-    return f(this.t) ? this : None
+  static flatten<T>(t: Maybeable<T>): Maybe<T> {
```
agarant, line 10: Could it be private ?

header-validator/src/maybe.ts
```diff
@@ -1,29 +1,30 @@
...
-    return new Some(f(this.t))
+  private constructor(private readonly t?: T) {}
+
+  filter(f: (t: T) => boolean): Maybe<T> {
+    return this.t === undefined || !f(this.t) ? Maybe.None : this
   }
 
-  flatMap<U>(f: (t: T) => Maybe<U>): Maybe<U> {
-    return f(this.t)
+  map<U>(f: (t: T) => Maybeable<U>): Maybe<U> {
```
agarant, line 20:
```suggestion
  map<U>(f: (t: T) => U | Maybe<U>): Maybe<U> {
```

I feel like having the type inline here might be simpler if it is its only use.

header-validator/src/maybe.ts
```diff
@@ -1,29 +1,30 @@
...
+    return this.t === undefined ? Maybe.None : Maybe.flatten(f(this.t))
   }
 
-  peek(f: (t: T) => void): Some<T> {
-    f(this.t)
+  peek(f: (t: T) => void): Maybe<T> {
+    if (this.t !== undefined) {
```
agarant, line 25:
```suggestion
    if (this !== None) {
```

To double check, but given that all None variables will point to the static instance, I believe that we could do an object equality check.

header-validator/src/maybe.ts
apasel422: It will be used in the followup PR, but noted.

header-validator/src/maybe.ts
apasel422: It will be used in the followup PR, but noted.

header-validator/src/maybe.ts
apasel422: That is correct, but the compiler would be unable to infer that the value is defined inside the branch.

agarant: Right, this would catch the some(undefined) case. Then perhaps having an hasValue or isNone might be benificial. But different than undefined is still pretty explicit.

agarant (approved):

header-validator/src/maybe.ts
apasel422: If we want to handle `Some(undefined)` properly, then the other checks in this file also need to be updated.

agarant: Can you clarify what you meant by this: "but the compiler would be unable to infer that the value is defined inside the branch"

apasel422: With your suggestion:
```
src/maybe.ts:26:9 - error TS2345: Argument of type 'T | undefined' is not assignable to parameter of type 'T'.
```

apasel422: One way to work around it would be to use separate classes for `None` and `Some` that extend from a common base class.

agarant: Oh, right.

agarant: We could also have a token for undefined similar to absl::nulloptr that we could check. But undefined is semantically close to that.

PR #993, opened 2023-09-06
Rename boolean returning functions

agarant:
Rename boolean returning functions to make them more distinctive from the functions returning parsed values.

### DIFF
--- a/header-validator/src/validate-json.ts
+++ b/header-validator/src/validate-json.ts
@@ -216,7 +216,7 @@ function list(
   })
 }
 
-function matchPattern(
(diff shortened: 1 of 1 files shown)


header-validator/src/validate-json.ts
```diff
@@ -216,7 +216,7 @@ function list(
   })
 }
 
-function pattern(
+function matchPattern(
```
apasel422, line 219:
```suggestion
function matchesPattern(
```

for consistency.

apasel422 (changes requested):

PR #1002, opened 2023-09-11
Check event-level channel capacity limit in header validator

apasel422:
If vendor-specific values are enabled and the source is otherwise valid, an error will be omitted if the effective channel capacity of the source is greater than the limit for the selected source type.

### DIFF
--- a/ts/src/header-validator/validate-json.ts
+++ b/ts/src/header-validator/validate-json.ts
@@ -3,6 +3,7 @@ import { SourceType } from '../source-type'
 import { VendorSpecificValues } from '../vendor-specific-values'
 import * as context from './context'
 import { Maybe, Maybeable } from './maybe'
+import * as privacy from '../flexible-event/privacy'
 
 const { None, some } = Maybe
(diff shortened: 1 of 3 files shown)


ts/src/header-validator/validate-json.ts
```diff
@@ -752,6 +755,47 @@ function eventReportWindow(
...
+    ctx.vsv.triggerDataCardinality === undefined ||
+    s.maxEventLevelReports === null
+  ) {
+    // TODO: consider warning when this cannot be checked
+    return
+  }
+
+  const perTriggerDataConfigs = []
+  for (let i = 0n; i < ctx.vsv.triggerDataCardinality[ctx.sourceType]; i++) {
+    perTriggerDataConfigs.push(
```
apasel422, line 771: @csharrison PTAL at this in particular

csharrison: I read through this function and it LGTM to describe the existing spec.

PR #1009, opened 2023-09-12
Run tests using Node test runner

apasel422:
- Runs all tests instead of stopping at the first failure
- Prints individual success/failure output more clearly

### DIFF
--- a/header-validator/src/util.test.ts
+++ b/header-validator/src/util.test.ts
@@ -1,4 +1,5 @@
 import { strict as assert } from 'assert'
+import test from 'node:test'
 import * as context from './context'
 
 export type TestCase = {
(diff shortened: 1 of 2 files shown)


flexible-event/privacy.test.ts
```diff
@@ -1,11 +1,12 @@
+import { strict as assert } from 'assert'
```
agarant, line 1: Should we use `import assert from 'node:assert';`

apasel422: How does that work with `strict`?

agarant: It looks like strict would be the best approach. Non strict is the legacy version of the api and produce less consistent diffs.

https://nodejs.org/api/assert.html#strict-assertion-mode

agarant, line 1:
```suggestion
import assert from 'node:assert/strict';
```

agarant (approved):

PR #1018, opened 2023-09-13
Separate build/pack Node scripts

apasel422:
This helps prevent redundant build operations, which can be slow.

### DIFF
--- a/ts/package.json
+++ b/ts/package.json
@@ -5,7 +5,7 @@
   "private": true,
   "scripts": {
     "build": "tsc",
-    "deploy": "webpack --mode production",
+    "pack": "webpack --mode production",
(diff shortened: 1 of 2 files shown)


ts/package.json
```diff
@@ -4,9 +4,10 @@
   "description": "",
   "private": true,
   "scripts": {
-    "build": "webpack --mode production",
+    "build": "tsc",
+    "deploy": "webpack --mode production",
```
agarant, line 8:
```suggestion
    "compile": "tsc",
    "build": "webpack --mode production",
```

deploy sounds like deploying to a server.

apasel422: Changed to `pack`.

agarant (approved):

PR #1021, opened 2023-09-14
Restrict and default summary_buckets based on max_event_level_reports

apasel422:
The max_event_level_reports value (whether explicit or defaulted based on source type) restricts the maximum number of reports that can be produced from a single source. Therefore, it is misleading to have more than that number of summary_buckets, since bucket increments will stop once reaching max_event_level_reports.

Rejecting sources with inconsistent values in these fields will help prevent developer errors, and both the rejection and defaulting based on max_event_level_reports bounds the storage space needed per trigger_spec for the summary_buckets field to 20, which is the max settable value for max_event_level_reports.

### DIFF
--- a/flexible_event_config.md
+++ b/flexible_event_config.md
@@ -139,7 +139,8 @@ In addition to the parameters that were added in Phase 1, we will add one additi
     // Represents a bucketization of the integers from [0, MAX_UINT32], encoded as
     // a list of integers where new buckets begin (excluding 0 which is
     // implicitly included).
-    // It must consist of strictly increasing positive integers.
+    // It must consist of strictly increasing positive integers, must be
+    // non-empty, and must have at most max_event_level_reports values.
     //
     // e.g. [5, 10, 100] encodes the following ranges:
     // [[0, 4], [5, 9], [10, 99], [100, MAX_UINT32]]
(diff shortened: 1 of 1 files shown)


yanzhangnyc (commented):
Let take max_event_level_reports = 2, summary_bucket = [1,2,3] for example. There are two options option 1) as this change to reject the source; option 2) keep register the source, and automatically trim [1,2,3] to [1,2]. Currently, we are using 2), but if there are consensus to choose option 1, we can go for it. I am neural, maybe slightly prefer option 2) for false tolerance because Adtech can still get something useful.

apasel422:
> I am neural, maybe slightly prefer option 2) for false tolerance because Adtech can still get something useful.

This seems inconsistent with other changes we have made to parsing to prefer errors to silent truncation / defaults, e.g. #793.

yanzhangnyc:
> > I am neural, maybe slightly prefer option 2) for false tolerance because Adtech can still get something useful.
> 
> This seems inconsistent with other changes we have made to parsing to prefer errors to silent truncation / defaults, e.g. #793.

I see and it is valid point. I don't have strong opinion here and would like to follow the general consensus of everyone.

csharrison (approved):
This is fine w/ me if everyone else is happy!

apasel422:
One downside to this is that if `max_event_level_reports` is omitted, it defaults based on the source type, which means that if the supplied `summary_buckets` is trimmed instead of rejected, it could automatically work with both types, but I suspect that the combination of omitted `max_event_level_reports` and present `summary_buckets` will be rare and that the benefit of this silent truncation is dubious. Source registration requests always indicate which type of source will be registered in the `Attribution-Reporting-Eligible` header, so the reporting origin will always have the opportunity to send a different response that works for each type, if desired.

PR #1024, opened 2023-09-14
Remove duplicate vendor-specific values from flex-event scripts

apasel422:


ts/src/flexible-event/privacy.ts
```diff
@@ -108,26 +101,27 @@ export class Config {
...
-  ),
-  [SourceType.event]: new Config(
-    /*maxEventLevelReports=*/ 1,
-    new Array(2).fill(
-      new PerTriggerDataConfig(/*numWindows=*/ 1, /*numSummaryBuckets=*/ 1)
+export function defaultConfig(
+  sourceType: SourceType,
+  vsv: VendorSpecificValues
+): Config {
+  const maxReports = vsv.defaultEventLevelAttributionsPerSource[sourceType]
```
csharrison, line 115: can we call it `defaultMaxReports`

ts/src/flexible-event/privacy.ts
```diff
@@ -108,26 +101,27 @@ export class Config {
...
+  return new Config(
+    /*maxEventLevelReports=*/ maxReports,
+    new Array(Number(vsv.triggerDataCardinality[sourceType])).fill(
+      new PerTriggerDataConfig(
+        /*numWindows=*/ maxReports,
```
csharrison, line 120: This is kind of a hack that just happens to work. Can we either specify this in the vsv, or else add a comment that this is just a lucky coincidence?

apasel422: I agree that this is a hack, but to make sure I understand: Rather than make this a VSV, we should make it a constant (1 for event-level, 3 for navigation) based on the hardcoding in Step 1 of https://wicg.github.io/attribution-reporting-api/#obtain-default-effective-windows?

apasel422: Done.

PR #1036, opened 2023-09-21
Correct effective range for omitted summary_buckets

apasel422:
MAX_INT always ends the final bucket; it never begins a bucket. In other words, if the default were explicitly represented in the JSON, the final value would be `MAX_INT - 1`.


apasel422:
Actually, maybe I'm misunderstanding what the comment I'm modifying means.

PR #1041, opened 2023-09-25
Use stricter TypeScript configuration

apasel422:


ts/src/flexible-event/privacy.ts
```diff
@@ -54,12 +54,12 @@ export class Config {
...
       (totalCap: number, index: number, w: number, c: number): number => {
-        if (index == 0 && w == 0) {
+        if (index === 0 && w === 0) {
           return 1
         }
 
-        if (w == 0) {
-          const triggerConfig = this.perTriggerDataConfigs[index - 1]
+        if (w === 0) {
+          const triggerConfig = this.perTriggerDataConfigs[index - 1]!
```
agarant, line 62:
```suggestion
          const triggerConfig = this.perTriggerDataConfigs.at(index - 1)!
```

agarant (approved):

ts/src/flexible-event/privacy.ts
apasel422: Just wondering why?

agarant: Just to be consistent with line 80

PR #1042, opened 2023-09-25
Add trigger_data_matching to Flexible Event explainer

### LINKED ISSUE #765: Flexible event backwards incompatibility: modulus operator on trigger data
https://wicg.github.io/attribution-reporting-api/#obtain-an-event-level-report specifies currently how we deal with the `trigger_data` associated with an event-level trigger configuration:

> [trigger data](https://wicg.github.io/attribution-reporting-api/#event-level-report-trigger-data)
The remainder when dividing config’s [trigger data](https://wicg.github.io/attribution-reporting-api/#event-level-trigger-configuration-trigger-data) by triggerDataCardinality.

This is not really compatible with the flexible configuration because that positions `trigger_data` as another filtering criteria. I don't see a good way to support something like this unless we force sources to specify data in the range [0, max], but this breaks the use case illustrated in [this example](https://github.com/WICG/attribution-reporting-api/blob/main/flexible_event_config.md#varying-trigger_specs-from-source-to-source).

cc @avvarad @apasel422

apasel422:
Fixes #765.

### DIFF
--- a/flexible_event_config.md
+++ b/flexible_event_config.md
@@ -155,6 +155,21 @@ In addition to the parameters that were added in Phase 1, we will add one additi
     // Next trigger_spec
   }, ...],
 
+  // Specifies how to match the trigger's trigger_data, which is a 64-bit
+  // unsigned integer, against the source's trigger specs, whose trigger data
+  // is constrained to a maximum of 32 distinct values, each of which is in the
+  // range [0, MAX_UINT32]:
+  //
+  // "exact":   The trigger data must exactly match one of the source's trigger
+  //            specs. Otherwise, no event-level attribution occurs.
+  // "modulus": The trigger data is taken modulus the number of distinct trigger
+  //            data across all of the source's trigger specs, and the resulting
+  //            value is used to index into that set of trigger data sorted in
+  //            ascending order.
+  //
+  // Defaults to "modulus".
+  "trigger_data_matching": <one of "exact" or "modulus">,
+
   // See description in phase 1.
   "max_event_level_reports": <int>,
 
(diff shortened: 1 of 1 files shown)


flexible_event_config.md
```diff
@@ -155,6 +155,21 @@ In addition to the parameters that were added in Phase 1, we will add one additi
...
   }, ...],
 
+  // Specifies how to match the trigger's trigger_data, which is a 64-bit
+  // unsigned integer, against the source's trigger specs, whose trigger data
+  // is constrained to a maximum of 32 distinct values, each of which is in the
+  // range [0, MAX_UINT32]:
+  //
+  // "exact":   The trigger data must exactly match one of the source's trigger
+  //            specs. Otherwise, no event-level attribution occurs.
+  // "modulus": The trigger data is taken modulus the number of distinct trigger
```
csharrison, line 165: It might help to add an example. IIUC, if the trigger data is [1, 5, 1000] and `modulus` is used:

- Trigger data 1 will match 1
- Trigger data 4 will match 1
- Trigger data 5 will match nothing
- Trigger data 1001 will match nothing
- Trgger data 1003 will match 1

is that right?

apasel422: I was thinking it would be something like this:

```c++
uint32_t Modulus(uint64_t input, const base::flat_set<uint32_t>& allowed_trigger_data) {
  CHECK(!allowed_trigger_data.empty());
  if (allowed_trigger_data.contains(input)) {
    return static_cast<uint32_t>(input);
  }
  return allowed_trigger_data[input % allowed_trigger_data.length];
}
```

In other words, `modulus` always causes there to be a match. But if you have other suggestions, let me know. If we think `modulus` should only be used with contiguous ranges starting at 0, we could do that, since we're really only adding it for backwards compatibility with default trigger specs, which are contiguous and which use modulus today.

csharrison: I don't love this, because it has the input acting both as `trigger_data` and an index into the possible trigger data. I feel it could end up confusing.

I wonder if we should explore alternatives where the input as as _either_ an index only, or a `trigger_data` only. The latter is what I had in mind in my above comment, but the former would look something like

```c++
uint32_t Modulus(uint64_t input, const std::vector<uint32_t>& allowed_trigger_data) {
  return allowed_trigger_data[input % allowed_trigger_data.length];
}
```

The problem IMO with this approach is that there is nothing indicating that the `trigger_data` in the trigger registration could possibly "merely" act as an index. It warrants a separation in the "type" of `trigger_data` on the source and trigger side.

apasel422: I'm certainly open to alternatives here. I agree that it is potentially confusing.

PR #1044, opened 2023-09-26
Add trigger_data_matching field to Flexible Event explainer

### LINKED ISSUE #765: Flexible event backwards incompatibility: modulus operator on trigger data

apasel422:
This allows us to retain backwards compatibility with the existing modulus behavior while allowing sources to opt in to exact matching. Other matching modes, such as `clamp_min`, `clamp_max`, and `clamp`, could be added in the future.

Fixes #765

### DIFF
--- a/flexible_event_config.md
+++ b/flexible_event_config.md
@@ -9,6 +9,7 @@ _Note: This document describes possible new functionality in the Attribution Rep
 - [Goals](#goals)
 - [Phase 2: Full Flexible Event-Level](#phase-2-full-flexible-event-level)
   - [API changes](#api-changes)
+  - [Trigger-data modulus matching example](#trigger-data-modulus-matching-example)
 - [Configurations that are equivalent to the current version](#configurations-that-are-equivalent-to-the-current-version)
   - [Equivalent event sources](#equivalent-event-sources)
   - [Equivalent navigation sources](#equivalent-navigation-sources)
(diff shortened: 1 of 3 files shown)


flexible_event_config.md
```diff
@@ -101,6 +102,21 @@ In addition to the parameters that were added in Phase 1, we will add one additi
...
 
+  // Specifies how the 64-bit unsigned trigger_data from the trigger is matched
+  // against the source's trigger_specs trigger_data, which is 32-bit. Defaults
+  // to "modulus".
+  //
+  // If "exact", the trigger_data must exactly match a value contained in the
+  // source's trigger_specs; if there is no such match, no event-level
+  // attribution takes place.
+  //
+  // If "modulus", the source's trigger specs must contain trigger_data forming
```
johnivdel, line 113: This is slightly unclear to me. The example below shoes that the set of all trigger_data values across all trigger_specs must be a contiguous sequence starting at 0. Could we rephrase to make this more clear?

johnivdel (approved):
LGTM

johnivdel (approved):

csharrison:
RS LGTM

PR #1046, opened 2023-09-26
Require vendor-specific values for source/trigger validation

apasel422:
Proper validation relies more and more on the presence of all
vendor-specific values, so requiring this here simplifies the code,
especially for the forthcoming Full Flexible Event-Level changes.

If there are alternative vendor-specific values in the future, we will
expose them in the interactive header validator using a dropdown menu.


csharrison (approved):

csharrison:
Maybe for a follow-up, but it might be good to have a tool-tip / alt text explaining the situation for why the checkbox is not unselectable.

PR #1051, opened 2023-09-29
Fix build due to fenced frame spec update

linnan-github:
Will prepare a PR to move the monkeypatch to fenced frame spec soon.

***
<a href="https://pr-preview.s3.amazonaws.com/linnan-github/conversion-measurement-api/pull/1051.html" title="Last updated on Sep 30, 2023, 8:50 PM UTC (21b2b47)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/attribution-reporting-api/1051/ce437a6...linnan-github:21b2b47.html" title="Last updated on Sep 30, 2023, 8:50 PM UTC (21b2b47)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -532,9 +532,9 @@ Add the step:
 
 # Fenced Frame monkeypatches # {#fenced-frame-monkeypatches}
 
-Add the following items to [=fencedframetype/pending event=]:
+Add the following items to [=fencedframetype/destination enum event=]:
 
-<dl dfn-for="pending event">
+<dl dfn-for="destination enum event">
 : <dfn>attributionReportingEnabled</dfn>
 :: A [=boolean=]
 : <dfn>contextOrigin</dfn>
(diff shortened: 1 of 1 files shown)


linnan-github:
@domfarolino, could you PTAL, thanks!

index.bs
```diff
@@ -532,9 +532,9 @@ Add the step:
 
 # Fenced Frame monkeypatches # {#fenced-frame-monkeypatches}
 
-Add the following items to [=fencedframetype/pending event=]:
+Add the following items to [=fencedframetype/destination enum event=]:
 
-<dl dfn-for="pending event">
+<dl dfn-for="destination enum event">
```
domfarolino, line 537: Hmm, why are these two lines necessary to fix the build? A "pending event" still exists in the FF spec: https://wicg.github.io/fenced-frame/#fencedframetype-pending-event.

index.bs
```diff
@@ -546,98 +546,57 @@ Modify {{Fence/reportEvent()}} as follows:
...
-
-<div algorithm="report-an-event-monkeypatch">
+Before the step
 
-Add a [=Document=] parameter called |document|.
-
-After the step
-
-> If |reporting map|[|destination|] does not [=map/exist=], return.
+> [=list/For each=] |destination| of <var ignore=''>event</var>'s {{FenceEvent/destination}}:
```
domfarolino, line 551: In what algorithm? You should probably link to it in the "Before the step in the [....] algorithm" above, to make this much more clear. Or maybe I'm missing something? Basically I wouldn't reference an unattached step like this...

domfarolino, line 551: You don't need `ignore=''` here. You just need `ignore`.

index.bs
```diff
@@ -546,98 +546,57 @@ Modify {{Fence/reportEvent()}} as follows:
...
 
 <div algorithm="send-a-beacon-monkeypatch">
 
-Add a [=boolean=] parameter called |attributionReportingEnabled| and an [=origin=]
-parameter called |contextOrigin|.
-
 After the step
 
-> Let <var ignore>destination url</var> be <var ignore>destination map</var>[<var ignore>eventType</var>].
+> Let <var ignore=''>destination url</var> be an empty [=string=].
```
domfarolino, line 581: Same as above...

index.bs
linnan-github: Yeah, but I think "destination enum event" is where these two properties are needed.

index.bs
linnan-github: Thanks, it's in the reportEvent() algorithm (please expand the unchanged text above), but added a few more steps to be clearer.

index.bs
domfarolino: OK, so this isn't related to the "fixing a broken build", this is just a general spec change that's included here?

index.bs
domfarolino: Ah, sorry I missed that.

domfarolino (approved):

index.bs
linnan-github: Thanks Dom. The build error is due to accessing the eventType/eventData field of pending event, but I refactored with the latest fenced frame spec.

apasel422 (approved):

PR #1053, opened 2023-10-02
Add version to the aggregation coordinator public key endpoint

linnan-github:
Add version to the .well-known path, this allows future changes to public keys. This changes the endpoint for the client (e.g Chrome) to fetch public keys, and the ad-techs are not impacted.

***
<a href="https://pr-preview.s3.amazonaws.com/linnan-github/conversion-measurement-api/pull/1053.html" title="Last updated on Oct 16, 2023, 4:17 PM UTC (95377ba)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/attribution-reporting-api/1053/74450a3...linnan-github:95377ba.html" title="Last updated on Oct 16, 2023, 4:17 PM UTC (95377ba)">Diff</a>

### DIFF
--- a/AGGREGATE.md
+++ b/AGGREGATE.md
@@ -308,7 +308,7 @@ shared.
 The encryption will use public keys specified by the aggregation service. The
 browser will encrypt payloads just before the report is sent by fetching the
 public key endpoint (the aggregation service coordinator origin at the path
- `/.well-known/aggregation-service/public-keys`) with an un-credentialed request. The processing origin will
+ `/.well-known/aggregation-service/v1/public-keys`) with an un-credentialed request. The processing origin will
 respond with a set of keys which will be stored according to standard HTTP
 caching rules, i.e. using Cache-Control headers to dictate how long to store the
 keys for (e.g. following the [freshness
@@ -328,6 +328,8 @@ encoded public keys is as follows:
   ]
 }
 ```
+
+Note: The version in the `.well-known` path may change.
 
 To limit the impact of a single compromised key, multiple keys (up to a small
 limit) can be provided. The browser should independently pick a key uniformly at
--- a/AGGREGATE.md
+++ b/AGGREGATE.md
@@ -329,7 +329,9 @@ encoded public keys is as follows:
 }
 ```
 
-Note: The version in the `.well-known` path may change.
+Note: The version in the `.well-known` path may be updated in the future
+versions of the spec if the public key serving details change, especially in a
+backwards incompatible way.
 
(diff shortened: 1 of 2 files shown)


apasel422:
What's the plan for doing this backwards-compatibly? Is the browser going to send reports to both the old and new endpoints during a deprecation period? Would it make more sense to continue using the existing endpoint unmodified until there actually is a version 2, at which point reports will be sent to a path that includes `/v2/`? I don't see the benefit of making this change now, since reporting origins will have to be updated to understand the version 2 format anyway, at which point they can also consume it at the new endpoint.

apasel422:
May also want to fix the `public` typo in the PR title.

linnan-github:
> What's the plan for doing this backwards-compatibly? Is the browser going to send reports to both the old and new endpoints during a deprecation period? Would it make more sense to continue using the existing endpoint unmodified until there actually is a version 2, at which point reports will be sent to a path that includes `/v2/`? I don't see the benefit of making this change now, since reporting origins will have to be updated to understand the version 2 format anyway, at which point they can also consume it at the new endpoint.

Thanks @apasel422. This change is for the public key endpoint for the client (Chrome/Android) to fetch public keys, and ad-techs are not impacted by this change.

apasel422:
> > What's the plan for doing this backwards-compatibly? Is the browser going to send reports to both the old and new endpoints during a deprecation period? Would it make more sense to continue using the existing endpoint unmodified until there actually is a version 2, at which point reports will be sent to a path that includes `/v2/`? I don't see the benefit of making this change now, since reporting origins will have to be updated to understand the version 2 format anyway, at which point they can also consume it at the new endpoint.
> 
> Thanks @apasel422. This change is for the public key endpoint for the client (Chrome/Android) to fetch public keys, and ad-techs are not impacted by this change.

Ah, missed that. Thanks!

AGGREGATE.md
```diff
@@ -329,6 +329,8 @@ encoded public keys is as follows:
 }
 ```
 
+Note: The version in the `.well-known` path may change.
```
alexmturner, line 332: Maybe let's add a bit more color here, i.e. when this may change. (Something like "may be updated in future versions of the spec if the details of public key serving details change, especially in a backwards incompatible way".)

alexmturner (commented):
LGTM % maybe updating the note

AGGREGATE.md
linnan-github: Thanks, updated with the details.

alexmturner (approved):

PR #1055, opened 2023-10-02
Export eligibility in spec

linnan-github:
This is needed to move fenced frame monkeypatches to fenced frame spec.

***
<a href="https://pr-preview.s3.amazonaws.com/linnan-github/conversion-measurement-api/pull/1055.html" title="Last updated on Oct 2, 2023, 3:13 PM UTC (296a2c0)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/attribution-reporting-api/1055/a59787c...linnan-github:296a2c0.html" title="Last updated on Oct 2, 2023, 3:13 PM UTC (296a2c0)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -163,7 +163,7 @@ name.
 Whenever an <{img}> or a <{script}> |element| is created or |element|'s
 {{HTMLAttributionSrcElementUtils/attributionSrc}} attribute is set or changed,
 run [=make background attributionsrc requests=] with |element| and
-"<code>[=attributionreporting/eligibility/event-source-or-trigger=]</code>".
+"<code>[=eligibility/event-source-or-trigger=]</code>".
 
 Issue: More precisely specify which mutations are relevant for the
 attributionsrc attribute.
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -1545,9 +1545,9 @@ and a [=suitable origin=] |reportingOrigin|:
 
 <h3 id="making-a-background-attributionsrc-request">Making a background attributionsrc request</h3>
 
-An <dfn>eligibility</dfn> is one of the following:
+An <dfn export for=attributionreporting>eligibility</dfn> is one of the following:
```
apasel422, line 1548: Why is the `attributionreporting` prefix necessary?

apasel422 (changes requested):

index.bs
linnan-github: When it’s accessed outside ARA spec, it would be more clear that this is the ARA eligibility?

apasel422: I don't think we need to clutter our spec to make that possible. The spec that is referencing ours should be able to use prefixes of its own to accomplish that.

https://speced.github.io/bikeshed/#custom-dfns

apasel422: Otherwise, this change is setting the precedent that every single exported definition in our spec should have this prefix, which seems redundant with the fact that our specification is already namespaced by virtue of being located at a particular URL. I don't think other specifications include such a prefix on their definitions, which makes them harder to read within their own Bikeshed files.

linnan-github: I see prefix used in fenced frame spec, but that's fair. Updated to remove the namespace. Thanks.

PR #1064, opened 2023-10-09
Specify trigger-data matching

apasel422:
Followup to #1044

***
<a href="https://pr-preview.s3.amazonaws.com/apasel422/attribution-reporting-api/pull/1064.html" title="Last updated on Oct 12, 2023, 6:05 PM UTC (deb2bf0)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/attribution-reporting-api/1064/7b8dc41...apasel422:deb2bf0.html" title="Last updated on Oct 12, 2023, 6:05 PM UTC (deb2bf0)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -750,7 +750,7 @@ A <dfn>trigger-data matching mode</dfn> is one of the following:
 
 <dl dfn-for="trigger-data matching mode">
 : "<dfn><code>exact</code></dfn>"
-:: [=event-level trigger configuration/Trigger data=] must be >= 0 and the
+:: [=event-level trigger configuration/Trigger data=] must be >= 0 and < the
     [=default trigger data cardinality=]. Otherwise, no event-level attribution
     takes place.
 : "<dfn><code>modulus</code></dfn>"
--- a/index.bs
+++ b/index.bs
@@ -1076,6 +1076,7 @@ Possible values are:
 <li>"<dfn><code>trigger-event-excessive-reports</code></dfn>"
 <li>"<dfn><code>trigger-event-low-priority</code></dfn>"
 <li>"<dfn><code>trigger-event-no-matching-configurations</code></dfn>"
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -2760,6 +2784,8 @@ To <dfn>trigger event-level attribution</dfn> given an [=attribution trigger=] |
     return it.
 1. Let |report| be the result of running [=obtain an event-level report=] with |sourceToAttribute|, |trigger|,
     and |matchedConfig|.
+1. If |report| is null, return the [=triggering result=]
+    ("<code>[=triggering status/dropped=]</code>", null).
```
apasel422, line 2788: I think it would be more consistent to send a verbose debug report here, since it is similar to unmatched filters. Thoughts?

johnivdel: Agreed, this seems simplest to me.

johnivdel (commented):
I don't think this change includes the requirement that matching=modulous requires a set of contiguous integers starting at 0. Do you plan to spec that separately?

apasel422:
> I don't think this change includes the requirement that matching=modulous requires a set of contiguous integers starting at 0. Do you plan to spec that separately?

Because the rest of the Full Flex proposal hasn't been specified yet, there is no way for a source to use non-default trigger data. As written, the spec mandates that the trigger data is in the range `[0, defaultTriggerDataCardinality)`.

index.bs
apasel422: Should there be any new fields in the debug report body? If we use the precedent of mismatched filters, then the answer is no, but it could be helpful to include the trigger's `trigger_data` and the source's set of allowed `trigger_data`. Thoughts?

johnivdel (approved):

akashnadan (approved):

PR #1066, opened 2023-10-11
Specify max bytes per filter string and max bytes per aggregation key identifier as constants

linnan-github:
***
<a href="https://pr-preview.s3.amazonaws.com/linnan-github/conversion-measurement-api/pull/1066.html" title="Last updated on Oct 16, 2023, 1:47 PM UTC (d4c010c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/attribution-reporting-api/1066/6198a8c...linnan-github:d4c010c.html" title="Last updated on Oct 16, 2023, 1:47 PM UTC (d4c010c)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1197,6 +1197,11 @@ Its value is 50.
 controls the maximum [=set/size=] of each [=map/value=] of an
 [=attribution source=]'s [=attribution source/filter data=]. Its value is 50.
 
+<dfn>Max bytes per filter string</dfn> is a positive interger that controls the
+maximum [=string/length=] of an [=attribution source=]'s [=attribution source/filter data=]'s
+[=map/keys=] and its [=map/values=]'s [=set/items=].
+Its value is 25.
+
 <dfn>Attribution rate-limit window</dfn> is a positive [=duration=] that
 controls the rate-limiting window for attribution. Its value is 30 days.
 
@@ -1219,6 +1224,12 @@ Its value is «[ [=source type/navigation=] → 3, [=source type/event=] → 1 ]
 <dfn>Allowed aggregatable budget per source</dfn> is a positive integer that controls the total
 [=aggregatable report/required aggregatable budget=] of all [=aggregatable reports=] created for
 an [=attribution source=]. Its value is 65536.
+
+<dfn>Max bytes per aggregation key identifier</dfn> is a positive integer that controls
+the maximum [=string/length=] of an [=attribution source=]'s [=attribution source/aggregation keys=]'s
+[=map/keys=], an [=attribution trigger=]'s [=attribution trigger/aggregatable values=]'s [=map/keys=],
+and an [=aggregatable trigger data=]'s [=aggregatable trigger data/source keys=]'s [=set/items=].
+Its value is 25.
 
 # Vendor-Specific Values # {#vendor-specific-values}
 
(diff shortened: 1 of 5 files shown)


index.bs
```diff
@@ -1197,6 +1197,11 @@ Its value is 50.
 controls the maximum [=set/size=] of each [=map/value=] of an
 [=attribution source=]'s [=attribution source/filter data=]. Its value is 50.
 
+<dfn>Max bytes per filter string</dfn> is a positive interger that controls the
```
apasel422, line 1200:
```suggestion
<dfn>Max bytes per filter string</dfn> is a positive integer that controls the
```

apasel422 (changes requested):
I'm not completely sure that the spec changes here match the currently implemented behavior in Chromium, based on [the Infra definition of a string](https://infra.spec.whatwg.org/#strings), which "is a sequence of unsigned 16-bit integers", and the TODO at https://wicg.github.io/attribution-reporting-api/#issue-e6d5a493.

I think it's also a bit confusing for our spec to use the term `bytes` when Infra strings do not work that way.

Might be good to get an expert in this area to take a look, as well as adding a non-normative note to the spec explaining why these limits exist in the first place.

linnan-github:
> I'm not completely sure that the spec changes here match the currently implemented behavior in Chromium, based on [the Infra definition of a string](https://infra.spec.whatwg.org/#strings), which "is a sequence of unsigned 16-bit integers", and the TODO at https://wicg.github.io/attribution-reporting-api/#issue-e6d5a493.
> 
> I think it's also a bit confusing for our spec to use the term `bytes` when Infra strings do not work that way.
> 
> Might be good to get an expert in this area to take a look, as well as adding a non-normative note to the spec explaining why these limits exist in the first place.

Thanks. Updated to use length instead of bytes.

@yoavweiss Could you please take a look and suggest how to properly spec the string length limit check? Thanks!

apasel422 (changes requested):

ts/src/header-validator/validate-json.ts
```diff
@@ -561,7 +561,21 @@ function filterDataKeyValue(
     return None
   }
 
-  return set(ctx, j, string, {
+  const filterStringLength = (s: string) => {
```
apasel422, line 564: It may be easier to understand the errors here with an optional prefix so that the key being too long can be more easily distinguished from the value being too long, e.g.

```ts
const filterStringLength = (s: string, errPrefix: string = '') => {
  ...
  ctx.error(`${errPrefix}exceeds...`)

...

if (!filterStringLength(key, 'key ')) {
```

Same for the aggregation keys below.

apasel422 (changes requested):

ts/src/header-validator/validate-json.ts
linnan-github: Is that necessary with the path already?

apasel422: Yes, because the path does not contain an element that distinguishes key from value.

linnan-github: Updated with the error prefix, thanks!

ts/src/header-validator/validate-json.ts
```diff
@@ -1077,7 +1104,11 @@ function source(ctx: Context, j: Json): Maybe<Source> {
 }
 
 function sourceKeys(ctx: Context, j: Json): Maybe<Set<string>> {
-  return set(ctx, j, string)
+  return set(ctx, j, (ctx, j) =>
+    string(ctx, j).filter((s) => {
+      return aggregationKeyIdentifierLength(ctx, s)
+    })
```
apasel422, line 1110:
```suggestion
    string(ctx, j).filter((s) =>
      aggregationKeyIdentifierLength(ctx, s)
    )
```

apasel422 (approved):
Header validator LGTM. Please wait for approval of the spec changes.

linnan-github:
@yoavweiss, just friendly ping, could you please take a look whether this is the right way to specify the limit on the string length? Thanks!!

yoavweiss:
The current definitions based on string/length seem fine to me.

linnan-github:
> The current definitions based on string/length seem fine to me.

Thanks @yoavweiss!

apasel422 (approved):

PR #1073, opened 2023-10-16
attributionsrc eligible response handling should handle in parallel

linnan-github:
***
<a href="https://pr-preview.s3.amazonaws.com/linnan-github/conversion-measurement-api/pull/1073.html" title="Last updated on Oct 16, 2023, 1:20 PM UTC (77ebc5b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/attribution-reporting-api/1073/11882e6...linnan-github:77ebc5b.html" title="Last updated on Oct 16, 2023, 1:20 PM UTC (77ebc5b)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1653,6 +1653,10 @@ To <dfn export>process an attribution eligible response</dfn> given a [=suitable
     state is not immediately detectable. Attribution may also be blocked if the reporting origin is not
     <a href="https://github.com/privacysandbox/attestation">enrolled</a>.
 
+1. [=Queue a task=] on the [=networking task source=] to proceed with the following steps.
+
+    Note: This algorithm can be invoked from while [=in parallel=].
+
 1. [=Assert=]: |eligibility| is
     "<code>[=eligibility/navigation-source=]</code>" or
     "<code>[=eligibility/event-source=]</code>" or


index.bs
```diff
@@ -1653,6 +1653,10 @@ To <dfn export>process an attribution eligible response</dfn> given a [=suitable
     state is not immediately detectable. Attribution may also be blocked if the reporting origin is not
     <a href="https://github.com/privacysandbox/attestation">enrolled</a>.
 
+1. [=Queue a task=] on the [=networking task source=] to proceed with the following steps.
```
agarant, line 1656: It seems that these tasks are more related to dealing with the event queue which I'm not sure is exactly what we need here. I had in mind to look at these steps: https://html.spec.whatwg.org/multipage/infrastructure.html#parallelism i.e., creating a new parallel queue and en queuing on it.

Did you look at both, could you detail the idea ? thanks

linnan-github: I think these tasks should be run in the same queue to ensure the sequence, so using the event queue seems to be better?

agarant: ack. I read further on the event queue, that makes sense, thanks.

agarant (approved):

PR #1074, opened 2023-10-16
Specify parsing of trigger_specs' trigger_data and event_report_windows

apasel422:
Note that we do not *use* the corresponding algorithms yet.

We will handle the `summary_window_operator` and `summary_buckets` fields separately.

https://github.com/WICG/attribution-reporting-api/blob/main/flexible_event_config.md#phase-2-full-flexible-event-level

***
<a href="https://pr-preview.s3.amazonaws.com/apasel422/attribution-reporting-api/pull/1074.html" title="Last updated on Oct 23, 2023, 3:58 PM UTC (0660908)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/attribution-reporting-api/1074/2f80c94...apasel422:0660908.html" title="Last updated on Oct 23, 2023, 3:58 PM UTC (0660908)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -579,9 +579,8 @@ A randomized response output configuration is a [=struct=] with the following it
 <dl dfn-for="randomized response output configuration">
 : <dfn>max attributions per source</dfn>
 :: A positive integer.
-: <dfn>trigger data map</dfn>
-:: A [=map=] whose keys are trigger data (non-negative integers) and whose
-    values are [=report window list=].
+: <dfn>trigger specs</dfn>
+:: A [=trigger spec map=].
 
 </dl>
 
@@ -656,6 +655,19 @@ A <dfn>trigger-data matching mode</dfn> is one of the following:
     [=default trigger data cardinality=].
 
 </dl>
+
+<h3 id="trigger-specs-header">Trigger specs</h3>
+
+A <dfn>trigger spec</dfn> is a [=struct=] with the following items:
+
+<dl dfn-for="trigger spec">
+: <dfn>event-level report windows</dfn>
+:: A [=report window list=].
+
+</dl>
+
+A <dfn>trigger spec map</dfn> is a [=map=] whose keys are unsigned 32-bit
+integers and values are [=trigger specs=].
 
 <h3 dfn-type=dfn>Attribution source</h3>
 
@@ -1150,6 +1162,10 @@ Its value is 25.
 <dfn>Default trigger data cardinality</dfn> is a [=map=] that
 controls the valid range of [=event-level trigger configuration/trigger data=].
 Its value is «[=source type/navigation=] → 8, [=source type/event=] → 2».
+
+<dfn>Max distinct trigger data per source</dfn> is a positive integer that
+controls the maximum [=map/size=] of a [=trigger spec map=] for an
+[=attribution source=]. Its value is 32.
 
 # Vendor-Specific Values # {#vendor-specific-values}
 
@@ -1671,8 +1687,8 @@ To <dfn export>process an attribution eligible response</dfn> given a [=suitable
 
 To <dfn>obtain a set of possible trigger states</dfn> given a [=randomized response output configuration=] |config|:
 1. Let |possibleTriggerStates| be a new [=list/is empty|empty=] [=set=].
-1. [=map/iterate|For each=] |triggerData| → |reportWindows| of |config|'s [=randomized response output configuration/trigger data map=]:
-    1. [=list/For each=] |reportWindow| of |reportWindows|:
+1. [=map/iterate|For each=] |triggerData| → |spec| of |config|'s [=randomized response output configuration/trigger specs=]:
+    1. [=list/For each=] |reportWindow| of |spec|'s [=trigger spec/event-level report windows=]:
         1. Let |state| be a new [=trigger state=] with the items:
             : [=trigger state/trigger data=]
             :: |triggerData|
@@ -1789,7 +1805,7 @@ a [=moment=] |sourceTime|, and a [=duration=] |eventReportWindow|:
     1. Set |lastEnd| to |lastEnd| + |deadline|.
 1. Return |windows|.
 
-To <dfn>parse report windows</dfn> given a [=map=] |map|, a [=moment=] |sourceTime|,
+To <dfn>parse top-level report windows</dfn> given a [=map=] |map|, a [=moment=] |sourceTime|,
 a [=source type=] |sourceType|, and a [=duration=] |expiry|:
 
 1. If |map|["`event_report_window`"] [=map/exists=] and
@@ -1802,16 +1818,21 @@ a [=source type=] |sourceType|, and a [=duration=] |expiry|:
 1. If |map|["`event_report_windows`"] does not [=map/exist=], return the result
     of [=obtaining default effective windows=] given |sourceType|, |sourceTime|,
     and |expiry|.
-1. Let |values| be |map|["`event_report_windows`"].
-1. If |values| is not a [=map=], return an error.
+1. Return the result of [=parsing report windows=] with
+    |map|["`event_report_windows`"], |sourceTime|, and |expiry|.
+
+To <dfn>parse report windows</dfn> given a |value|, a
+[=moment=] |sourceTime|, and a [=duration=] |expiry|:
+
+1. If |value| is not a [=map=], return an error.
 1. Let |startDuration| be 0 seconds.
-1. If |values|["`start_time`"] [=map/exists=]:
-    1. Let |start| be |values|["`start_time`"].
+1. If |value|["`start_time`"] [=map/exists=]:
+    1. Let |start| be |value|["`start_time`"].
     1. If |start| is not a non-negative integer, return an error.
     1. Set |startDuration| to |start| seconds.
     1. If |startDuration| is greater than |expiry|, return an error.
-1. If |values|["`end_times`"] does not [=map/exist=] or is not a [=list=], return an error.
-1. Let |endDurations| be |values|["`end_times`"].
+1. If |value|["`end_times`"] does not [=map/exist=] or is not a [=list=], return an error.
+1. Let |endDurations| be |value|["`end_times`"].
 1. If the [=list/size=] of |endDurations| is greater than [=max settable event-level report windows=], return an error.
 1. If |endDurations| [=list/is empty=], return an error.
 1. Let |windows| be an [=list/is empty|empty=] [=list=].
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -1119,6 +1132,10 @@ Its value is «[ [=source type/navigation=] → 3, [=source type/event=] → 1 ]
 [=aggregatable report/required aggregatable budget=] of all [=aggregatable reports=] created for
 an [=attribution source=]. Its value is 65536.
 
+<dfn>Max distinct trigger data per source</dfn> is a positive integer that
+controls the maxium [=map/size=] of a [=trigger spec map=] for an
```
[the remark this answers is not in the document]
giladbarkan-github: We'll remove event_report_windows (flex lite) in favour of only full flex?

index.bs
```diff
@@ -1131,7 +1148,9 @@ controls how many [=attribution sources=] can be in the
 
 <dfn>Default trigger data cardinality</dfn> is a [=map=] that
 controls the valid range of [=event-level trigger configuration/trigger data=].
-The keys are «[=source type/navigation=], [=source type/event=]». The values are positive integers.
+The keys are «[=source type/navigation=], [=source type/event=]». The values are
+integers greater than 0 and less than or equal to
```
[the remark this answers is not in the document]
giladbarkan-github: The latter case is only for trigger matching type modulus, no?

index.bs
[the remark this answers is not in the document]
apasel422: (Registrations remain the same. In my initial comment in this thread I meant that the [spec field](https://wicg.github.io/attribution-reporting-api/#attribution-source-event-level-report-windows) would be removed.)

index.bs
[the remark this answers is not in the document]
apasel422: > The latter case is only for trigger matching type modulus, no?

This constant is just for the *default* trigger data cardinality, which is what is used when no `trigger_specs` field is present.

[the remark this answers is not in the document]
giladbarkan-github: Oh, I'm not reading in context, sorry. I'll look more carefully next time.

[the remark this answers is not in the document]
apasel422: (This change needs to be rebased on #1064 at which point I will update the comment here.)

index.bs
```diff
@@ -1797,6 +1822,48 @@ a [=source type=] |sourceType|, and a [=duration=] |expiry|:
     1. Set |startDuration| to |endDuration|.
 1. Return |windows|.
 
+To <dfn>parse trigger specs</dfn> given a [=map=] |map|, a [=moment=]
+|sourceTime|, a [=duration=] |expiry|, a [=report window list=]
+|topLevelWindows|, and an unsigned 32-bit integer |defaultTriggerDataCardinality|:
```
linnan-github, line 1827: Not sure whether it's easier to read with |defaultReportWindows|?

index.bs
```diff
@@ -1797,6 +1822,48 @@ a [=source type=] |sourceType|, and a [=duration=] |expiry|:
...
+1. Let |specs| be a new [=map=].
+1. If |map|["`trigger_specs`"] does not [=map/exist=]:
+    1. Let |spec| be a new [=trigger spec=] with the following items:
+        : [=trigger spec/event-level report windows=]
+        :: |topLevelWindows|
+    1. For each integer |triggerData| between 0 (inclusive) and
+        |defaultTriggerDataCardinality| (exclusive):
+        1. [=map/Set=] |specs|[|triggerData|] to |spec|.
+    1. Return |specs|.
+1. If |map|["`trigger_specs`"] is not a [=list=] or [=list/is empty=], or its
```
linnan-github, line 1840: Is this the only case that setting an empty list is treated differently from not setting the field?

apasel422: Yes, but this field is different from other lists in that it enumerates valid configurations, so it is potentially misleading for the empty list to be treated as the default.

This is related to the current behavior of not allowing `null` for `debug_key`, for example.

linnan-github: That makes sense, thanks for clarifying.

index.bs
```diff
@@ -1797,6 +1822,48 @@ a [=source type=] |sourceType|, and a [=duration=] |expiry|:
...
+        :: |topLevelWindows|
+    1. If |item|["`event_report_windows`"] [=map/exists=]:
+        1. Let |reportWindows| be the result of
+            [=parsing report windows ignoring legacy=] with |map|, |sourceTime|,
+            and |expiry|.
+        1. If |reportWindows| is an error, return it.
+        1. Set |spec|'s [=trigger spec/event-level report windows=] to
+            |reportWindows|.
+    1. If |item|["`trigger_data`"] does not [=map/exist=], is not a [=list=], or
+        [=list/is empty=], or its [=list/size=] is greater than
```
linnan-github, line 1856: Add the comment here as well on whether we can allow the no-op configuration for an empty list of "trigger_data".

apasel422: There's no reason to specify an empty list. I would rather make it an error.

That said, we *could* make it so that the `trigger_data` field itself is optional, and if it's omitted it could default to the sequence `[0, defaultTriggerDataCardinality)` which could allow optimization in some cases in which the reporting origin wants to use e.g. a non-default `summary_window_operator` but the default trigger data values.

linnan-github: I don't feel strongly, but as empty trigger_specs is now allowed, this may be the only empty list that will be rejected. Allowing empty trigger_data is no-op, but makes the size checking on the top level difficult, so probably it's fine leaving as is.

linnan-github: BTW, I feel it's a bit similar to source_keys field in aggregatable_trigger_data, which allows the no-op empty list.

apasel422: `event_report_windows.end_times` [also cannot be empty](https://wicg.github.io/attribution-reporting-api/#parse-report-windows).

The yet-to-be-specified `summary_buckets` also cannot be empty.

linnan-github: That's fair,thanks.

index.bs
```diff
@@ -1767,6 +1786,12 @@ a [=source type=] |sourceType|, and a [=duration=] |expiry|:
 1. If |map|["`event_report_windows`"] does not [=map/exist=], return the result
     of [=obtaining default effective windows=] given |sourceType|, |sourceTime|,
     and |expiry|.
+1. Return the result of [=parsing report windows ignoring legacy=] with |map|,
```
giladbarkan-github, line 1789: It seems odd to use the word, "legacy." I think of a specification as always current. Also, in some time, we might have another feature that could have a concept of "legacy" and we may need to differentiate. If a feature is supported then it should have a descriptive name based on the feature.

apasel422: Do you have a suggestion?

index.bs
```diff
@@ -1797,6 +1822,48 @@ a [=source type=] |sourceType|, and a [=duration=] |expiry|:
     1. Set |startDuration| to |endDuration|.
 1. Return |windows|.
 
+To <dfn>parse trigger specs</dfn> given a [=map=] |map|, a [=moment=]
+|sourceTime|, a [=duration=] |expiry|, a [=report window list=]
+|defaultReportWindows|, and an unsigned 32-bit integer |defaultTriggerDataCardinality|:
+
+1. [=Assert=]: |defaultTriggerDataCardinality| is greater than 0 and less than
```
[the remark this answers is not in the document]
apasel422: `|defaultTriggerDataCardinality|` is a parameter to this algorithm, not a constant, so the preconditions on it need to be asserted separately.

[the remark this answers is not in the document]
giladbarkan-github: Ok, makes sense.

index.bs
```diff
@@ -643,6 +643,19 @@ It has the following constraints:
...
+A <dfn>trigger spec</dfn> is a [=struct=] with the following items:
+
+<dl dfn-for="trigger spec">
+: <dfn>event-level report windows</dfn>
+:: A [=report window list=].
+
+</dl>
+
+A <dfn>trigger spec map</dfn> is a [=map=] whose keys are unsigned 32-bit
+integers and values are [=trigger specs=].
```
giladbarkan-github, line 657: The values here seem to lead to the same header link in the webpage. Is this a recursive definition? I didn't understand trigger specs as a recursive structure.

giladbarkan-github: Oh, right. I thought the map was a second entry in the definition of trigger spec. Sorry, I need to improve my reading accuracy.

index.bs
giladbarkan-github: Something like `To parse a report window list`?

index.bs
```diff
@@ -1767,6 +1786,12 @@ a [=source type=] |sourceType|, and a [=duration=] |expiry|:
 1. If |map|["`event_report_windows`"] does not [=map/exist=], return the result
     of [=obtaining default effective windows=] given |sourceType|, |sourceTime|,
     and |expiry|.
+1. Return the result of [=parsing report windows ignoring legacy=] with |map|,
+    |sourceTime|, and |expiry|.
+
+To <dfn>parse report windows ignoring legacy</dfn> given a [=map=] |map|, a
+[=moment=] |sourceTime|, and a [=duration=] |expiry|:
+
```
giladbarkan-github, line 1794: On line 1798, shouldn't ` |map|["`start_time`"] [=map/exists=]` be ` |values|["`start_time`"] [=map/exists=]`?

index.bs
giladbarkan-github: `...event report window struct`?

index.bs
apasel422: Resolved in #1077.

index.bs
```diff
@@ -643,6 +643,19 @@ It has the following constraints:
...
 * There is at least one element in the list.
 
+<h3 id="trigger-specs-header">Trigger specs</h3>
+
+A <dfn>trigger spec</dfn> is a [=struct=] with the following items:
+
+<dl dfn-for="trigger spec">
+: <dfn>event-level report windows</dfn>
+:: A [=report window list=].
+
```
giladbarkan-github, line 653: Should we also mention it could have a `start_time` field?

apasel422: The `report window list` structure already encompasses the start time: https://wicg.github.io/attribution-reporting-api/#report-window-header

index.bs
```diff
@@ -1768,6 +1787,13 @@ a [=source type=] |sourceType|, and a [=duration=] |expiry|:
 1. If |map|["`event_report_windows`"] does not [=map/exist=], return the result
     of [=obtaining default effective windows=] given |sourceType|, |sourceTime|,
     and |expiry|.
+1. Return the result of [=parsing report windows=] with |map|,
+    |sourceTime|, and |expiry|.
+
+To <dfn>parse report windows</dfn> given a [=map=] |map|, a
+[=moment=] |sourceTime|, and a [=duration=] |expiry|:
+
+1. [=Assert=]: |map|["`event_report_windows`"] [=map/exists=].
```
linnan-github, line 1796: Maybe we can just pass the values in so as no assertion is needed?

giladbarkan-github: Wouldn't we need an assertion **somewhere**?

giladbarkan-github (approved):

index.bs
linnan-github: IIUC this algorithm is only invoked after checking that the value exists.

apasel422:
> I assume summary-bucket-related fields are coming later.

Yes, as stated in the PR description.

linnan-github (approved):

giladbarkan-github (approved):

index.bs
```diff
@@ -1832,6 +1853,48 @@ a [=source type=] |sourceType|, and a [=duration=] |expiry|:
...
+        1. Let |reportWindows| be the result of
+            [=parsing report windows=] with |item|["`event_report_windows`"],
+            |sourceTime|, and |expiry|.
+        1. If |reportWindows| is an error, return it.
+        1. Set |spec|'s [=trigger spec/event-level report windows=] to
+            |reportWindows|.
+    1. If |item|["`trigger_data`"] does not [=map/exist=], is not a [=list=], or
+        [=list/is empty=], or its [=list/size=] is greater than
+        [=max distinct trigger data per source=], return an error.
+    1. [=list/iterate|For each=] |triggerData| of |map|["`trigger_data`"]:
```
csharrison, line 1889:
```suggestion
    1. [=list/iterate|For each=] |triggerData| of |item|["`trigger_data`"]:
```

csharrison (changes requested):

PR #1092, opened 2023-11-03
Perform event-report-window check after deduplication instead of before

apasel422:
Event-level report deduplication is based only on whether a given deduplication key has previously been seen for a given source. As such, it makes more sense to perform deduplication as soon as both of these values are known, instead of after the event-report-window check, which is not special with respect to any of the existing checks that are already performed after deduplication.

This change affects backwards compatibility in a minor way, in that the "trigger-event-deduplicated" verbose debug report will now be sent preferentially over the "trigger-event-report-window-not-started" and "trigger-event-report-window-passed" verbose debug reports, but this only affects triggers that are allowed to send verbose debug reports at all and is only observable when both such failure conditions occur together.

This change makes the specification more readily compatible with the Full Flexible Event Level proposal, as with that proposal, the event-report-window check can only be performed after the trigger spec, if any, has been found for the trigger's trigger data. (In the language of the spec, there is no "total window" for the source as a whole, only a total window for each of its trigger specs.)

***
<a href="https://pr-preview.s3.amazonaws.com/apasel422/attribution-reporting-api/pull/1092.html" title="Last updated on Nov 3, 2023, 4:35 PM UTC (e7670ae)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/attribution-reporting-api/1092/2ef5146...apasel422:e7670ae.html" title="Last updated on Nov 3, 2023, 4:35 PM UTC (e7670ae)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2752,21 +2752,6 @@ To <dfn>trigger event-level attribution</dfn> given an [=attribution trigger=] |
         with "<code>[=trigger debug data type/trigger-event-noise=]</code>", |trigger|, |sourceToAttribute| and
         [=obtain debug data on trigger registration/report=] set to null.
     1. Return the [=triggering result=] ("<code>[=triggering status/dropped=]</code>", |debugData|).
-1. Let |windowResult| be the result of [=check whether a moment falls within a window=]
-    with |trigger|'s [=attribution trigger/trigger time=] and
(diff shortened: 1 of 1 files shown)


akashnadan (approved):

vikassahu29 (approved):

johnivdel (approved):
LGTM

PR #1093, opened 2023-11-06
Warn about inconsistent aggregatable keys in trigger registrations

apasel422:
Keys that are present in either aggregatable_trigger_data or aggregatable_values but not both.

### DIFF
--- a/ts/src/header-validator/trigger.test.ts
+++ b/ts/src/header-validator/trigger.test.ts
@@ -28,12 +28,12 @@ const testCases: jsontest.TestCase<Trigger>[] = [
         "not_filters": {"c": ["d"]},
         "source_keys": ["x"]
       }],
-      "aggregatable_values": {"e": 5},
+      "aggregatable_values": {"x": 5},
       "debug_key": "5",
       "debug_reporting": true,
       "event_trigger_data": [{
         "deduplication_key": "123",
-        "filters": {"x": []},
+        "filters": {"e": []},
         "not_filters": {"y": []},
         "priority": "-7",
         "trigger_data": "1"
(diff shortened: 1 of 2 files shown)


ts/src/header-validator/trigger.test.ts
```diff
@@ -28,12 +28,12 @@ const testCases: jsontest.TestCase<Trigger>[] = [
         "not_filters": {"c": ["d"]},
         "source_keys": ["x"]
       }],
-      "aggregatable_values": {"e": 5},
+      "aggregatable_values": {"x": 5},
```
linnan-github, line 31: Just curious, why are we changing this?

apasel422: So that `"x"` doesn't appear in two different locations, which could be confusing.

ts/src/header-validator/validate-json.ts
```diff
@@ -1213,6 +1213,40 @@ function aggregatableSourceRegistrationTime(
...
+  const triggerDataKeys = new Set<string>()
+
+  ctx.scope('aggregatable_trigger_data', () => {
+    for (const [index, datum] of t.aggregatableTriggerData.entries()) {
+      ctx.scope(index, () => {
+        for (const key of datum.sourceKeys) {
+          triggerDataKeys.add(key)
+
+          if (!t.aggregatableValues.has(key)) {
+            ctx.scope('source_keys', () =>
```
linnan-github, line 1226: If multiple keys in the same `source_keys` have warning message, probably can be consolidated to the same one? But it's fine leaving as is for readability.

linnan-github (approved):

ts/src/header-validator/validate-json.ts
apasel422: Ack.

PR #1097, opened 2023-11-07
Specify and validate event_level_epsilon

### LINKED ISSUE #1095: Add an epsilon param to the event-level API
I propose that we add an optional field `event_level_epsilon` to event-level source registration. This field would replace the existing [randomized response epsilon](https://wicg.github.io/attribution-reporting-api/#randomized-response-epsilon) vendor-specific value, and replace that with something like a “maximum randomized response epsilon”, which `event_level_epsilon` could not exceed without the source failing to register. The default `event_level_epsilon` will be this maximum vendor-specific value.

The benefit of this approach is that:
- it allows API callers to experiment with different privacy levels
- it allows privacy-conscious API callers to optionally deploy more private versions of the API
- it allows callers to invoke the API consistently across multiple interoperable implementations, even if each implementation chooses a different maximum epsilon
- it aligns with the design of the aggregation service, which allows [specifying](https://github.com/privacysandbox/aggregation-service/blob/main/docs/api.md) an epsilon up to a maximum

apasel422:
Fixes #1095

***
<a href="https://pr-preview.s3.amazonaws.com/apasel422/attribution-reporting-api/pull/1097.html" title="Last updated on Nov 14, 2023, 4:56 PM UTC (264efb4)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/attribution-reporting-api/1097/8e60251...apasel422:264efb4.html" title="Last updated on Nov 14, 2023, 4:56 PM UTC (264efb4)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1182,9 +1182,9 @@ controls the maximum [=map/size=] of a [=trigger spec map=] for an
 controls how many [=attribution sources=] can be in the
 [=attribution source cache=] per [=attribution source/source origin=].
 
-<dfn>Randomized response epsilon</dfn> is a non-negative double that controls
-the randomized response probability of an [=attribution source=]. If [=automation local testing mode=] is true,
-this is `∞`.
+<dfn>Max settable event-level epsilon</dfn> is a non-negative double that
+controls the default and maximum values that a source registration can specify
+for the epsilon parameter used by [=compute the channel capacity of a source=].
 
 <dfn>Randomized null report rate excluding source registration time</dfn> is a
 double between 0 and 1 (both inclusive) that controls the randomized number of null reports
@@ -1990,7 +1990,11 @@ To <dfn noexport>parse source-registration JSON</dfn> given a [=byte sequence=]
     : [=randomized response output configuration/trigger specs=]
     :: |triggerSpecs|
 
-1. Let |epsilon| be the user agent's [=randomized response epsilon=].
+1. Let |epsilon| be the user agent's [=max settable event-level epsilon=].
+1. Set |epsilon| to |value|["`event_level_epsilon`"] if it [=map/exists=]:
+1. If |epsilon| is not a double that is greater than or equal to 0 and less than
+    or equal to the user agent's [=max settable event-level epsilon=], return null.
+1. If [=automation local testing mode=] is true, set |epsilon| to `∞`.
 1. If the result of [=computing the channel capacity of a source=] with |randomizedResponseConfig| and |epsilon| is greater than
     [=max event-level channel capacity per source=][|sourceType|], return null.
 1. Let |triggerDataMatchingMode| be "<code>[=trigger-data matching mode/modulus=]</code>".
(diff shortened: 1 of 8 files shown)


index.bs
```diff
@@ -1182,9 +1182,9 @@ controls the maximum [=map/size=] of a [=trigger spec map=] for an
 controls how many [=attribution sources=] can be in the
 [=attribution source cache=] per [=attribution source/source origin=].
 
-<dfn>Randomized response epsilon</dfn> is a non-negative double that controls
-the randomized response probability of an [=attribution source=]. If [=automation local testing mode=] is true,
-this is `∞`.
+<dfn>Max settable event-level epsilon</dfn> is a non-negative double that
+controls the default and maximum values that a source registration can specify
+for the epsilon parameter used by [=compute the channel capacity of a source=].
```
csharrison, line 1187: It is weird to highlight the channel capacity algorithm and not https://wicg.github.io/attribution-reporting-api/#obtain-a-randomized-source-response, which is the actual DP mechanism.

index.bs
```diff
@@ -1990,7 +1990,11 @@ To <dfn noexport>parse source-registration JSON</dfn> given a [=byte sequence=]
     : [=randomized response output configuration/trigger specs=]
     :: |triggerSpecs|
 
-1. Let |epsilon| be the user agent's [=randomized response epsilon=].
+1. Let |epsilon| be the user agent's [=max settable event-level epsilon=].
+1. Set |epsilon| to |value|["`event_level_epsilon`"] if it [=map/exists=]:
+1. If |epsilon| is not a double that is greater than or equal to 0 and less than
+    or equal to the user agent's [=max settable event-level epsilon=], return null.
```
csharrison, line 1996:
```suggestion
1. If |epsilon| is less than 0, or greater than the user agent's [=max settable event-level epsilon=], return null.
```

ts/src/header-validator/validate-json.ts
```diff
@@ -17,13 +17,24 @@ const hex128Regex = /^0[xX][0-9A-Fa-f]{1,32}$/
 
 const UINT32_MAX: number = 2 ** 32 - 1
 
+export type ParseOpts = {
+  parseEventLevelEpsilon: boolean
```
csharrison, line 21: Can you help me understand why you added this as an optional thing?

ts/src/header-validator/validate-json.ts
```diff
@@ -17,13 +17,24 @@ const hex128Regex = /^0[xX][0-9A-Fa-f]{1,32}$/
...
   constructor(
     readonly vsv: Readonly<VendorSpecificValues>,
     readonly sourceType: SourceType,
-    readonly parseFullFlex: boolean
+    opts: Partial<ParseOpts>
   ) {
     super()
+    this.opts = {
+      parseEventLevelEpsilon: opts.parseEventLevelEpsilon ?? false,
+      parseFullFlex: opts.parseFullFlex ?? false,
```
csharrison, line 36: Can you use `structuredClone` instead? Like `this.opts = structuredClone(opts)`

csharrison (changes requested):

ts/src/header-validator/validate-json.ts
apasel422: It corresponds to whether the Chromium implementation will disable the feature by default. If we enable instead, I will remove this option.

index.bs
csharrison: Oh sorry I realized I deleted the `double` check.

ts/src/header-validator/validate-json.ts
csharrison: ack

csharrison (approved):

vikassahu29 (approved):

akashnadan (approved):

PR #1098, opened 2023-11-08
Scope event-level report prioritization using original report time

apasel422:
Otherwise, if the report time is adjusted, either due to retries or the browser being offline, the wrong report time will be used, which can cause later reports to erroneously replace earlier ones that haven't been successfully sent yet.

***
<a href="https://pr-preview.s3.amazonaws.com/apasel422/attribution-reporting-api/pull/1098.html" title="Last updated on Nov 8, 2023, 1:37 PM UTC (8ccdb52)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/attribution-reporting-api/1098/5fd28d6...apasel422:8ccdb52.html" title="Last updated on Nov 8, 2023, 1:37 PM UTC (8ccdb52)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2143,6 +2143,8 @@ a [=trigger state=] |triggerState|:
     |triggerState|'s [=trigger state/report window=]'s [=report window/end=].
 1. Return |fakeReport|.
 
+Issue: Set |fakeReport|'s [=event-level report/original report time=] properly.
+
 To <dfn>check if debug reporting is allowed</dfn> given a [=source debug data type=] |dataType|
 and a [=boolean=] |debugCookieSet|:
 1. If |dataType| is:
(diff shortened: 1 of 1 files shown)


linnan-github (approved):

index.bs
```diff
@@ -874,6 +874,10 @@ An attribution report is a [=struct=] with the following items:
 
 </dl>
 
+Issue: Consider removing the [=attribution report/delivered=] field, and instead
```
linnan-github, line 877: This is probably needed for the retry logic.

apasel422: I'm not sure if it's necessary for that, but even if it is, the value of that field should probably be asserted to be false in various other algorithms.

johnivdel (approved):
lgtm

PR #1102, opened 2023-11-09
Perform trigger-data matching before capacity checks

apasel422:
The new order makes more sense for Full Flex, as this step will be needed to obtain the matching trigger spec, which is necessary for the reordered report-window check in #1092.

The `trigger_data_matching` feature hasn't been released yet, so this should present no backwards-compatibility issues.


csharrison (approved):
This is fine, but maybe this check should go right after `trigger-event-no-matching-configurations` since they are checking very similar things? In a sense you could even argue the `trigger-event-no-matching-configurations` fits this error case as well...

apasel422:
> This is fine, but maybe this check should go right after `trigger-event-no-matching-configurations` since they are checking very similar things? In a sense you could even argue the `trigger-event-no-matching-configurations` fits this error case as well...

That would require changing the deduplication vs. matching ordering from #1092, so I'd prefer to keep it as is.

PR #1105, opened 2023-11-09
Reorganize report-delivery algorithms to prevent certain race conditions

### LINKED ISSUE #1101: Report delivery spec algorithm can send reports erroneously
https://wicg.github.io/attribution-reporting-api/#report-delivery says:

> The user agent MUST periodically [iterate](https://infra.spec.whatwg.org/#list-iterate) over its [event-level report cache](https://wicg.github.io/attribution-reporting-api/#event-level-report-cache) and [aggregatable report cache](https://wicg.github.io/attribution-reporting-api/#aggregatable-report-cache) and run [queue a report for delivery](https://wicg.github.io/attribution-reporting-api/#queue-a-report-for-delivery) on each item

But consider a report `R1` that was stored at time `T1` with report time `T4`. At time `T2`, the user agent iterates over its caches and queues `R1` for delivery. Because it is not yet `T4`, the user agent waits. At time `T3`, a higher priority report `R2` whose report time is also `T4` [replaces](https://wicg.github.io/attribution-reporting-api/#maybe-replace-event-level-report) `R1`, causing `R1` to be removed from the report cache. When time `T4` is reached, the user agent finishes waiting and sends `R1` despite it having been removed.

In general, this periodic queuing is prone to inaccurately reflecting the state of the report caches because the validity of the report is not checked immediately prior to sending.

apasel422:
We remove reports from the cache prior to sending them to ensure that reports that are truly pending (and therefore subject to modification or deletion for various reasons) are not mistaken for ones that are in the process of being sent. This also allows us to remove the separate fields for original report time and report time, as while a report is in the report cache, its report time is never modified.

Fixes #1101.

***
<a href="https://pr-preview.s3.amazonaws.com/apasel422/attribution-reporting-api/pull/1105.html" title="Last updated on Nov 15, 2023, 3:00 PM UTC (cb26732)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/attribution-reporting-api/1105/8b14390...apasel422:cb26732.html" title="Last updated on Nov 15, 2023, 3:00 PM UTC (cb26732)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -863,8 +863,6 @@ An attribution report is a [=struct=] with the following items:
 :: A [=moment=].
 : <dfn>original report time</dfn>
 :: A [=moment=].
-: <dfn>delivered</dfn> (default false)
-:: A [=boolean=].
 : <dfn>report ID</dfn>
 :: A [=string=].
 : <dfn>source debug key</dfn>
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -3246,22 +3235,30 @@ Issue: Specify this in terms of <a href="https://html.spec.whatwg.org/multipage/
 
 # Report delivery # {#report-delivery}
 
-The user agent MUST periodically [=set/iterate=] over its [=event-level report cache=] and
-[=aggregatable report cache=] and run [=queue a report for delivery=] on each item.
+The user agent MUST periodically run [=queue reports for delivery=] on the
+[=event-level report cache=] and [=aggregatable report cache=].
+
+To <dfn>queue reports for delivery</dfn> given a [=set=] of
+[=attribution reports=] |cache|, run the following steps [=in parallel=]:
```
csharrison, line 3242: Hm, but we probably don't want to touch the report cache in parallel since that can lead to data races, right? I think step one here should be executed on the same "event sequence" as every other algorithm that touches the report caches.

apasel422: Yes, in general, the entire spec needs to be audited to ensure correctness around exclusive access to all of the attribution caches. I'm not sure what the best way to do that is, but I don't know if that should block this PR, since the existing logic already suffers from the same issue.

It seems that @linnan-github made a [related change](https://github.com/WICG/attribution-reporting-api/pull/1108) recently.

I'm of the opinion that we may want to have a single multi-producer/single-consumer parallel queue containing source/trigger registrations and these periodic report-delivery attempts.

Thoughts?

csharrison: Agree we should audit, but I guess my point was that I think just omitting "in parallel" from this step is arguably more correct than the current PR.

csharrison (approved):

Issue #1103, opened 2023-11-09
High volume of source-storage-limit


michal-kalisz:
Hi,

For approximately 6% of all attempts to register ARA sources, we encounter the source-storage-limit. For PCs, it reaches 10% (for phones, it's about 3%).
For some source websites (often those for which we display the most ads), it goes above 20% (up to a maximum of 33%).

As ARA is only being tested on a small scale, and more ad tech companies are likely to use it as 3rd-party cookies phase out, this problem is becoming more noticeable. Especially, bigger players with lots of ads and partners could hit the limit, preventing new registrations until old ones expire.

Maybe a good solution would be to introduce a storage limit that also considers the reporting origin. Additionally, it seems important to have a mechanism that allows overwriting or deleting previous registrations.

Another idea is to define separate limits for different types of ARA sources (event, navigation). For instance, an event-type event can be registered multiple times, even without any user interaction, unlike navigation events.

Best regards,
Michal

csharrison:
Thanks for the report, let me also cc @agarant @akashnadan . We can look into this.

agarant:
Thanks for the report @michal-kalisz,  an increase from 1024 to 4096 of this limit will be effective from M120. We will be monitoring the impact of the increase. In parallel, we are looking further into additional mitigation measures.

Makpara:
> Hi,
> 
> For approximately 6% of all attempts to register ARA sources, we encounter the source-storage-limit. For PCs, it reaches 10% (for phones, it's about 3%). For some source websites (often those for which we display the most ads), it goes above 20% (up to a maximum of 33%).
> 
> As ARA is only being tested on a small scale, and more ad tech companies are likely to use it as 3rd-party cookies phase out, this problem is becoming more noticeable. Especially, bigger players with lots of ads and partners could hit the limit, preventing new registrations until old ones expire.
> 
> Maybe a good solution would be to introduce a storage limit that also considers the reporting origin. Additionally, it seems important to have a mechanism that allows overwriting or deleting previous registrations.
> 
> Another idea is to define separate limits for different types of ARA sources (event, navigation). For instance, an event-type event can be registered multiple times, even without any user interaction, unlike navigation events.
> 
> Best regards, Michal

AramZS:
I generally think that 

> a mechanism that allows overwriting or deleting previous registrations. 

seems wise. Especially if it can be handled both by count and date (delete x number starting with the least recently registered moving toward the most recently registered).

[...]

PR #1107, opened 2023-11-10
Increase max pending sources per source origin

agarant:
From UMA metrics, we identified that a higher than expected number of sources were being dropped due to the limit of 1024 sources per source origin, specifically ~6% are being dropped due to it.

This was also reported in https://github.com/WICG/attribution-reporting-api/issues/1103

A first step toward mitigating this issue is to increase the limit from 1024 to 4096. Analyses were conducted to confirm that no problematic increase in storage usage is expected from this change.

### DIFF
--- a/params/chromium-params.md
+++ b/params/chromium-params.md
@@ -7,7 +7,7 @@ Chromium's implementation assigns the following values:
 
 | Name | Value |
 | ---- | ----- |
-| [Max pending sources per source origin][] | [1024][max pending sources per source origin value] |
+| [Max pending sources per source origin][] | [4096][max pending sources per source origin value] |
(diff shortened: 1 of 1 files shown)


params/chromium-params.md
```diff
@@ -7,7 +7,7 @@ Chromium's implementation assigns the following values:
 
 | Name | Value |
 | ---- | ----- |
-| [Max pending sources per source origin][] | [1024][max pending sources per source origin value] |
+| [Max pending sources per source origin][] | [4096][max pending sources per source origin value] |
```
apasel422, line 10: Please also update the link below so that it goes to the updated Code Search line.

apasel422 (approved):

PR #1112, opened 2023-11-15
Propose a simple trigger prioritization algo for flex event

csharrison:


flexible_event_config.md
```diff
@@ -169,6 +170,15 @@ When the `event_report_window` for a spec completes, we will map its summary val
   "trigger_summary_bucket": [<bucket start>, <bucket end>]
 }
 ```
+### Trigger prioritization
+
+Given that triggering attribution can affect a source's state without producing a report, we will need a new algorithm for doing trigger prioritization. Here is a sketch of how it could work:
+
+1. Always maintain a sorted list of triggers, sorted in order of priority (descending) and trigger time (ascending)
```
yanzhangnyc, line 177: Always maintain a sorted list of triggers **that are not fully consumed in sent report** , sorted in order of priority (descending), **then** trigger time (ascending).

flexible_event_config.md
```diff
@@ -169,6 +170,15 @@ When the `event_report_window` for a spec completes, we will map its summary val
...
+2. Whenver a spec's window's end time is hit (breaking ties arbitrarily)
```
yanzhangnyc, line 178: When a new trigger "matching" this source, we will perform the following process. Not at end of a reporting window

flexible_event_config.md
```diff
@@ -169,6 +170,15 @@ When the `event_report_window` for a spec completes, we will map its summary val
...
+    1. Iterate through triggers in order, "applying" them to generate a list of "speculative" reports. Stop when privacy limits are hit.
```
yanzhangnyc, line 179: limits are hit OR all of the triggers are consumed.

flexible_event_config.md
giladbarkan-github: Why does this matter?

flexible_event_config.md
apasel422: "Iterate through triggers in order" already encompasses the second part of that disjunction.

flexible_event_config.md
csharrison: I intentionally described this per reporting window to minimize the amount of state / accounting. I think the end result should be the same?

apasel422: To me it's a lot easier to understand this algorithm when it's described as happening at the end of each report window, not when a trigger is handled.

flexible_event_config.md
```diff
@@ -169,6 +170,15 @@ When the `event_report_window` for a spec completes, we will map its summary val
...
+    2. Flush all of the speculative reports that are scheduled to  be emitted in the current window, and update the source's state based on all of the triggers that were successfully applied.
```
apasel422, line 180:
```suggestion
    2. Flush all of the speculative reports that are scheduled to be emitted in the current window, and update the source's state based on all of the triggers that were successfully applied.
```

apasel422 (approved):

flexible_event_config.md
apasel422: What does "fully consumed" mean here?

flexible_event_config.md
apasel422, line 178:
```suggestion
2. Whenever a spec's window's end time is hit (breaking ties arbitrarily):
```

yanzhangnyc: @csharrison @giladbarkan-github , no. If process at end of the reporting window, the trigger data with eariler reporting window will have the highest priority.

apasel422 (approved):

flexible_event_config.md
giladbarkan-github: Trigger value was fully applied in the creation of a report whose report time is equal to or earlier than the trigger time (for "count" operator, the value is 1).

giladbarkan-github: (Current trigger time, so report time less than current window start).

apasel422: In the context of this PR, I don't understand how a trigger could be partially consumed. Not that this PR goes into the details of summary buckets, but in step 2.1 is the trigger's entire value not added to the spec's current value, resulting in N reports depending on the summary buckets?

giladbarkan-github: If the previous report window reached bucket 10 with a trigger whose value is 12, we have a value of 2 left to consume in this window. If the trigger's value was 10, it was fully applied.

apasel422: Why wouldn't the spec's value be 12 at the end of the previous window instead of 10?

giladbarkan-github: All of this may be a bit of implementation details. Yan is pointing out that we don't really need to store triggers whose value has already fully contributed to sent reports.

apasel422: Right, triggers are removed in step 2.3.

flexible_event_config.md
apasel422: @yanzhangnyc The PR says "iterate through triggers in order". I take that to mean all triggers associated with this source.

flexible_event_config.md
yanzhangnyc: > "Iterate through triggers in order" already encompasses the second part of that disjunction.

Yes. You are right. I just suggest in writing to emphasize there are two outcomes for final states: 1, report limited reached, 2 no more triggers

flexible_event_config.md
yanzhangnyc: > @yanzhangnyc The PR says "iterate through triggers in order". I take that to mean all triggers associated with this source.

If so, we need to emphasize "at the end window of any trigger data" instead of just simply mention "trigger window end".

flexible_event_config.md
csharrison: @yanzhangnyc let me know if step 2.3 satisfies you. I updated the wording on the sorting aspect in the latest commit.

flexible_event_config.md
csharrison: I updated the wording. PTAL

flexible_event_config.md
csharrison: I don't think this is necessary

yanzhangnyc: > I don't think this is necessary, lmk if you feel strongly

I am OK. No strong opinion. Just wording.

flexible_event_config.md
yanzhangnyc: > @yanzhangnyc let me know if step 2.3 satisfies you. I updated the wording on the sorting aspect in the latest commit.

I think step 2 and 3 are not "wrong" but some wording may cause misleading. Commented in []
For 2, Flush [Pass?] all of the speculative reports that are scheduled to be emitted in the current window [of given trigger data], and update the source's state [about the matched triggers?] based on all of the triggers that were successfully applied.  

For 3, Erase all of the triggers associated [I think we should emphasize the trigger has been fully consumed, not only associated as discussed before] with the current spec and window from the source's trigger list

csharrison: - Regarding [of the given trigger data] how about "of the given spec", since we are operating on a spec's window which can operate across many trigger data simultaneously?
- Regarding "about the matched triggers", I don't understand this comment or what it is clarifying, isn't the clause afterwards "based on all the triggers that were successfully applied" saying a similar thing?
- Regarding "fully consumed", can you elaborate on the concern? Why isn't deleting all triggers associated with the spec + window sufficient?

flexible_event_config.md
```diff
@@ -169,6 +170,15 @@ When the `event_report_window` for a spec completes, we will map its summary val
...
 }
 ```
+### Trigger prioritization
+
+Given that triggering attribution can affect a source's state without producing a report, we will need a new algorithm for doing trigger prioritization. Here is a sketch of how it could work:
+
+1. For every source, maintain a list of triggers, sorted in order of priority (descending), then trigger time (ascending)
+2. At the end of any report window (across all of a source's specs, breaking ties arbitrarily):
+    1. Iterate through the source's triggers in order, "applying" them to generate a list of "speculative" reports. Stop when privacy limits are hit.
+    2. Flush all of the speculative reports that are scheduled to be emitted in the current window, and update the source's state based on all of the triggers that were successfully applied.
```
giladbarkan-github, line 180: Step 2.2 seems to call for deleting all the reports generated in step 2.1.

flexible_event_config.md
yanzhangnyc: For the original PR, I think we can publish as a general introduction as it is. Here is discussing to clarify some details. 
* In Flex Event API, the report window is affiliated with trigger data, and not an independent concept. When we mentioning do something at end of report window, there will be two understanding about the operation is applied to all triggers/reports or only to the triggers/reports to this trigger data that related to this window. 
* I believe you are indicating the field "attributed triggers" in the Source, right? The term "source state" sounds like a simple state like "deleted", "active", "expired". But it is OK to leave as it is. 
* Suppose a trigger have value of 12. The value of 8 has contributed to a report that have been sent so the trigger is "associated", but there are value of 4 left can be used to next report so it should not been "deleted".

apasel422: @yanzhangnyc I think you're describing an implementation, not an explainer-level algorithm description, particularly with the idea that part of a trigger's value is left over for another report in your third item.

giladbarkan-github: > Why isn't deleting all triggers associated with the spec + window sufficient?

Let's say we consumed 10 of a trigger's value of 12. If we erase this trigger, we have no way of knowing the priority of the remaining value of 2 as it applies to the procedure of generating reports in the next window.

apasel422: > > Why isn't deleting all triggers associated with the spec + window sufficient?
> 
> Let's say we consumed 10 of a trigger's value of 12. If we erase this trigger, we have no way of knowing the priority of the remaining value of 2 as it applies to the procedure of generating reports in the next window.

That is an implementation detail not relevant to the explainer-level description here and is covered by "update the source's state based on all of the triggers that were successfully applied."

giladbarkan-github: Ok, then if nothing else, this shows how hard it is for me to understand the current proposal.

giladbarkan-github: What does "update the source's state based on all of the triggers that were successfully applied" mean?

flexible_event_config.md
csharrison: Yep I think that's good to clarify: How about this 
```suggestion
    2. Flush all of the speculative reports that are scheduled to be emitted in the current window, and update the source's state based on all of the triggers that were successfully applied. Delete all of the speculative reports that are not scheduled to be emitted in this window.
```

giladbarkan-github: Oh, maybe I misunderstood. Does "flush" mean delete or send?

giladbarkan-github: If it means "send" then it's fine.

csharrison: It means send! I don't know why I used such an ambiguous word 🤦

csharrison:
```suggestion
    2. Send all of the speculative reports that are scheduled to be emitted in the current window, and update the source's state based on all of the triggers that were successfully applied. Delete all of the speculative reports that are not scheduled to be emitted in this window.
```

flexible_event_config.md
```diff
@@ -169,6 +170,17 @@ When the `event_report_window` for a spec completes, we will map its summary val
   "trigger_summary_bucket": [<bucket start>, <bucket end>]
 }
 ```
+### Trigger prioritization
+
+Given that triggering attribution can affect a source's state without producing a report, we will need a new algorithm for doing trigger prioritization. Here is a sketch of how it could work:
+
+1. For every source, maintain a list of triggers, sorted in order of priority (descending), then trigger time (ascending)
+2. At the end of any report window (across all of a source's specs, breaking ties arbitrarily):
```
apasel422, line 178: Is there any concern that `breaking ties arbitrarily` will make it harder to ensure interop conformance?

flexible_event_config.md
```diff
@@ -169,6 +170,17 @@ When the `event_report_window` for a spec completes, we will map its summary val
...
+    1. Iterate through the source's triggers in order, "applying" them to generate a list of "speculative" reports. Stop whenever privacy limits are hit.
+    2. Send all of the speculative reports that are scheduled to be emitted in the current window
+    3. Update the source's total value per trigger data, and total # of event-level reports based on all of the triggers that were successfully applied in the current window.
```
giladbarkan-github, line 181: What's the difference between "successfully" applied and otherwise applied?

flexible_event_config.md
```diff
@@ -169,6 +170,17 @@ When the `event_report_window` for a spec completes, we will map its summary val
...
+    4. Delete all of the speculative reports that are not scheduled to be emitted in this window.
```
giladbarkan-github, line 182: How did the reports that are deleted here get created?

flexible_event_config.md
shaiful1040885-code: Call ma various sxe per

flexible_event_config.md
```diff
@@ -169,6 +170,17 @@ When the `event_report_window` for a spec completes, we will map its summary val
...
+    5. Delete all of the triggers associated with the current spec and window from the source's trigger list
```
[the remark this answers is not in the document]
phuocankg778899-gif: 😀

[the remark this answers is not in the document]
phuocankg778899-gif: Reply

PR #1113, opened 2023-11-20
Explainer update for supporting trigger context ID and avoiding report delay for aggregate reporting

### LINKED ISSUE #974: Further reducing aggregatable report delays
Recent feedback ([meeting notes](https://github.com/WICG/attribution-reporting-api/blob/main/meetings/2023-08-21-minutes.md)) indicates that even the new modest report delays for aggregatable reports are not ideal, and we should explore the instant-style approaches outlined in https://github.com/WICG/attribution-reporting-api/blob/main/report_verification.md#could-we-just-tag-reports-with-a-trigger_id-instead-of-using-anonymous-tokens. Here is what a privacy neutral enhancement to the API could look like that enables this use-case:

Trigger registration could come along with an optional top-level parameter that ad techs can set: `trigger_event_id`[^1]. If this parameter is set, a few things happen:
- Aggregatable reports will be generated for this registration unconditionally, even if it was not preceded by a source event. In that case any generated report will act like the [existing null aggregatable reports](https://wicg.github.io/attribution-reporting-api/#generate-null-reports) i.e. they will not affect any aggregate computation.
- Aggregatable reports associated with this registration will be delivered immediately after the browser receives the registration.
- Aggregatable reports associated with this registration will come bearing a new piece of cleartext information in its body which passes the `trigger_event_id`.

By default this will result in a single report per registration. However, if `aggregatable_source_registration_time` is ”include”, it will require generating 30 reports for every trigger registration. Given that 30 reports is non-ideal from a system health perspective, we may want to consider disallowing this combination, and only support instant reports if the `aggregatable_source_registration_time` is “exclude” (feedback welcome though!). The purpose of these “unconditional” reports is to ensure that the output of the trigger registration does not leak any cross-site information (i.e. whether the user had previously seen an ad or not).

cc @thegreatfatzby , @bmayd, and @dmdabbs who voiced support for reducing delays all the way to 0 in our Aug 21 call, especially in relation to CPA-based billing.

Note: this proposal may greatly increase (up to 20x[^2]) the number of reports you receive and need to query the aggregation service with, which can impact aggregation query latency. Here is a derivation to find out how it will impact you:

- Let `true_rate` be the fraction of registrations that lead to a true report (i.e. have a previous matching source)
- With no noise, the number of reports is `true_rate * num_triggers`
- Currently, with the 5% flip probability the number of reports is `((1 - true_rate) * .05 + true_rate) * num_triggers = (.95 * true_rate + .05) * num_triggers`
- With this proposal, the number of reports is `num_triggers`. Thus, adopting this proposal would result in `(1 / (.95 * true_rate + .05))` times more total reports than the status quo (note that this smoothly interpolates between 0x and 20x at the extremes).

Note with `true_rate = .2`, this gives you ~4x times more total reports than the status quo.

[^1]: Keen observers will note the similarities between this and the `context_id` param in https://github.com/patcg-individual-drafts/private-aggregation-api/blob/main/report_verification.md#shared-storage
[^2]: Assuming aggregatable_source_registration_time is “exclude”.

linnan-github:
Fixes #974

### DIFF
--- a/AGGREGATE.md
+++ b/AGGREGATE.md
@@ -248,7 +248,7 @@ The report will be JSON encoded with the following scheme:
   "trigger_debug_key": "[64 bit unsigned integer]",
 
   // Optional trigger context ID.
-  "trigger_context_id": "example string"
+  "trigger_context_id": "example string" //max length 64
 }
 ```
 
--- a/AGGREGATE.md
+++ b/AGGREGATE.md
@@ -453,20 +453,20 @@ In order to limit abuse of the protections above, there will be a maximum limit 
 ### Optional: reduce report delay with trigger context ID
 
 Trigger registration will accept an optional string field `trigger_context_id`, a
-high-entropy ID which represents the data associated with the trigger.
+high-entropy ID that represents the data associated with the trigger.
 
 ```jsonc
 {
-  ..., // exisiting fields
+  ..., // existing fields
   "trigger_context_id": "example string" // max length 64
 }
 ```
 
-This ID would then be embedded unencrypted in the aggregatable report.
+This ID will be embedded unencrypted in the aggregatable report.
 
 To avoid leaking cross-site information through the count of reports with the
 given ID, the browser will unconditionally send an aggregatable report on every
-trigger registration with a trigger context ID. A null report would be sent in the
+trigger registration with a trigger context ID. A null report will be sent in the
 case that the trigger registration did not generate an attribution report. The
 source registration time will always be excluded from the aggregatable report
 with a trigger context ID.
--- a/AGGREGATE.md
+++ b/AGGREGATE.md
@@ -248,7 +248,7 @@ The report will be JSON encoded with the following scheme:
   "trigger_debug_key": "[64 bit unsigned integer]",
 
   // Optional trigger context ID.
-  "trigger_context_id": "example string" // max length 64
+  "trigger_context_id": "example string"
 }
 ```
 
--- a/AGGREGATE.md
+++ b/AGGREGATE.md
@@ -248,7 +248,7 @@ The report will be JSON encoded with the following scheme:
   "trigger_debug_key": "[64 bit unsigned integer]",
 
   // Optional trigger context ID.
-  "trigger_context_id": "example string" //max length 64
+  "trigger_context_id": "example string" // max length 64
 }
 ```
 
@@ -458,7 +458,7 @@ high-entropy ID which represents the data associated with the trigger.
 ```jsonc
 {
   ..., // exisiting fields
-  "trigger_context_id": "example string"
+  "trigger_context_id": "example string" // max length 64
 }
 ```
 
--- a/AGGREGATE.md
+++ b/AGGREGATE.md
@@ -18,7 +18,7 @@
   - [Contribution bounding and budgeting](#contribution-bounding-and-budgeting)
   - [Storage limits](#storage-limits)
   - [Hide the true number of attribution reports](#hide-the-true-number-of-attribution-reports)
-  - [Optional: avoid report delay with trigger context ID](#optional-avoid-report-delay-with-trigger-context-id)
+  - [Optional: reduce report delay with trigger context ID](#optional-reduce-report-delay-with-trigger-context-id)
 - [Data processing through a Secure Aggregation Service](#data-processing-through-a-secure-aggregation-service)
 - [Privacy considerations](#privacy-considerations)
   - [Differential Privacy](#differential-privacy)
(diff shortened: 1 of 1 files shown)


AGGREGATE.md
```diff
@@ -244,7 +245,10 @@ The report will be JSON encoded with the following scheme:
   // Optional debugging information (also present in event-level reports),
   // if the cookie `ar_debug` is present.
   "source_debug_key": "[64 bit unsigned integer]",
-  "trigger_debug_key": "[64 bit unsigned integer]"
+  "trigger_debug_key": "[64 bit unsigned integer]",
+
+  // Optional trigger context ID.
+  "trigger_context_id": "example string"
```
akashnadan, line 251: Will this be limited to a max length of 64?
```suggestion
  "trigger_context_id": "example string" //max length 64
```

AGGREGATE.md
```diff
@@ -446,6 +450,38 @@ those that include this field.
...
+  ..., // exisiting fields
+  "trigger_context_id": "example string"
+}
+```
+
+This ID would then be embedded unencrypted in the aggregatable report.
+
+To avoid leaking cross-site information through the count of reports with the
+given ID, the browser will unconditionally send an aggregatable report on every
+trigger registration with a context ID. A null report would be sent in the
```
akashnadan, line 469:
```suggestion
trigger registration with a trigger context ID. A null report would be sent in the
```

AGGREGATE.md
```diff
@@ -446,6 +450,38 @@ those that include this field.
...
+case that the trigger registration did not generate an attribution report. The
+source registration time will always be excluded from the aggregatable report
+with a context ID.
```
akashnadan, line 472:
```suggestion
with a trigger context ID.
```

AGGREGATE.md
```diff
@@ -446,6 +450,38 @@ those that include this field.
...
+
+As the context ID in the aggregatable report explicitly reveals the
```
akashnadan, line 474:
```suggestion
As the trigger context ID in the aggregatable report explicitly reveals the
```

AGGREGATE.md
```diff
@@ -446,6 +450,38 @@ those that include this field.
 
 In order to limit abuse of the protections above, there will be a maximum limit of 20 aggregatable reports per source.
 
+### Optional: avoid report delay with trigger context ID
```
akashnadan, line 453:
```suggestion
### Optional: reduce report delay with trigger context ID
```

AGGREGATE.md
```diff
@@ -18,6 +18,7 @@
   - [Contribution bounding and budgeting](#contribution-bounding-and-budgeting)
   - [Storage limits](#storage-limits)
   - [Hide the true number of attribution reports](#hide-the-true-number-of-attribution-reports)
+  - [Optional: avoid report delay with trigger context ID](#optional-avoid-report-delay-with-trigger-context-id)
```
akashnadan, line 21:
```suggestion
  - [Optional: reduce report delay with trigger context ID](#optional-reduce-report-delay-with-trigger-context-id)
```

akashnadan (dismissed):
minor edits

akashnadan (approved):

AGGREGATE.md
```diff
@@ -446,6 +450,38 @@ those that include this field.
...
 In order to limit abuse of the protections above, there will be a maximum limit of 20 aggregatable reports per source.
 
+### Optional: reduce report delay with trigger context ID
+
+Trigger registration will accept an optional string field `trigger_context_id`, a
+high-entropy ID which represents the data associated with the trigger.
+
+```jsonc
+{
+  ..., // exisiting fields
```
apasel422, line 460:
```suggestion
  ..., // existing fields
```

AGGREGATE.md
```diff
@@ -446,6 +450,38 @@ those that include this field.
...
+  "trigger_context_id": "example string" // max length 64
+}
+```
+
+This ID would then be embedded unencrypted in the aggregatable report.
```
apasel422, line 465:
```suggestion
This ID will be embedded unencrypted in the aggregatable report.
```

AGGREGATE.md
```diff
@@ -446,6 +450,38 @@ those that include this field.
...
+
+To avoid leaking cross-site information through the count of reports with the
+given ID, the browser will unconditionally send an aggregatable report on every
+trigger registration with a trigger context ID. A null report would be sent in the
```
apasel422, line 469:
```suggestion
trigger registration with a trigger context ID. A null report will be sent in the
```

apasel422 (changes requested):

apasel422 (approved):

PR #1114, opened 2023-11-20
Spec supporting trigger context ID for aggregate reporting

### LINKED ISSUE #974: Further reducing aggregatable report delays

linnan-github:
Fixes #974 

Explainer: https://github.com/WICG/attribution-reporting-api/pull/1113

***
<a href="https://pr-preview.s3.amazonaws.com/linnan-github/conversion-measurement-api/pull/1114.html" title="Last updated on Nov 21, 2023, 1:49 PM UTC (909ebee)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/attribution-reporting-api/1114/ed340d6...linnan-github:909ebee.html" title="Last updated on Nov 21, 2023, 1:49 PM UTC (909ebee)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -3055,7 +3055,7 @@ To <dfn>obtain an event-level report delivery time</dfn> given an [=attribution 
         |window|'s [=report window/end=].
 1. [=Assert=]: not reached.
 
-To <dfn>obtain an aggregatable report delivery time</dfn> given an [attribution trigger=]
+To <dfn>obtain an aggregatable report delivery time</dfn> given an [=attribution trigger=]
 |trigger|, perform the following steps. They return a [=moment=].
 
 1. Let |triggerTime| be |trigger|'s [=attribution trigger/trigger time=].
@@ -3478,7 +3478,7 @@ To <dfn>serialize an [=aggregatable report=] </dfn> |report|, run the following 
     |data|["`trigger_debug_key`"] to |report|'s [=aggregatable report/trigger debug key=],
     [=serialize an integer|serialized=].
 1. If |report|'s [=aggregatable report/trigger context ID=] is not null, [=map/set=]
-    |data|["`context_id`"] to |report|'s [=aggregatable report/trigger context ID=].
+    |data|["`trigger_context_id`"] to |report|'s [=aggregatable report/trigger context ID=].
 1. Return the [=byte sequence=] resulting from executing [=serialize an infra value to JSON bytes=] on |data|.
 
 To <dfn>serialize an [=attribution report=]</dfn> |report|, run the following steps:
--- a/index.bs
+++ b/index.bs
@@ -2438,7 +2438,7 @@ and a [=moment=] |triggerTime|:
 1. Let |triggerContextID| be null.
 1. If |value|["`trigger_context_id`"] [=map/exists=]:
     1. If |value|["`trigger_context_id`"] is not a [=string=], return null.
-    1. If |value|["`trigger_context_id`"] is empty or its [=string/length=] is greater than the [=max length per trigger context ID=],
+    1. If |value|["`trigger_context_id`"]'s [=string/length=] is 0 or is greater than the [=max length per trigger context ID=],
         return null.
     1. If |aggregatableSourceRegTimeConfig| is not "<code>[=aggregatable source registration time configuration/exclude=]</code>", return null.
     1. Set |triggerContextID| to |value|["`trigger_context_id`"].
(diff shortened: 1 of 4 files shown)


index.bs
```diff
@@ -1172,6 +1176,10 @@ Its value is «[=source type/navigation=] → 8, [=source type/event=] → 2».
 controls the maximum [=map/size=] of a [=trigger spec map=] for an
 [=attribution source=]. Its value is 32.
 
+<dfn>Max length per trigger context ID</dfn> is a positive integer that controls
+the maximum [=string/length=] of an [=attribution trigger=]'s [=attribution trigger/trigger context ID=],
+and an [=aggregatable report=]'s [=aggregatable report/trigger context ID=]. Its value is 64.
```
apasel422, line 1181: Similar to https://github.com/WICG/attribution-reporting-api/pull/1113/files#r1400607818 I'd rather not explicitly say that this constant limits the length of the *report* field, but rather just the input field.

index.bs
```diff
@@ -2425,6 +2435,13 @@ and a [=moment=] |triggerTime|:
     1. If |value|["`aggregatable_source_registration_time`"] is not an [=aggregatable source registration time configuration=],
         return null.
     1. Set |aggregatableSourceRegTimeConfig| to |value|["`aggregatable_source_registration_time`"].
+1. Let |triggerContextID| be null.
+1. If |value|["`trigger_context_id`"] [=map/exists=]:
+    1. If |value|["`trigger_context_id`"] is not a [=string=], return null.
+    1. If |value|["`trigger_context_id`"] is empty or its [=string/length=] is greater than the [=max length per trigger context ID=],
```
apasel422, line 2441:
```suggestion
    1. If |value|["`trigger_context_id`"]'s [=string/length=] is 0 or is greater than the [=max length per trigger context ID=],
```

index.bs
```diff
@@ -3172,13 +3196,17 @@ To <dfn>generate null reports</dfn> given an [=attribution trigger=] |trigger| a
 
 1. Let |nullReports| be a new [=list/is empty|empty=] [=list=].
 1. If |trigger|'s [=attribution trigger/aggregatable source registration time configuration=] is "<code>[=aggregatable source registration time configuration/exclude=]</code>":
+    1. Let |randomizedNullReportRate| be [=randomized null report rate excluding source registration time=].
+    1. If |trigger|'s [=attribution trigger|trigger context ID=] is not null, set
```
apasel422, line 3200:
```suggestion
    1. If |trigger|'s [=attribution trigger/trigger context ID=] is not null, set
```

apasel422 (changes requested):

index.bs

apasel422 (approved):

PR #1123, opened 2023-12-05
Update Attribution Reporting EVENT.md to note that browsers may choose to enable debugging for specific use-cases

nancymic2:
Add explanation about debugging for specific use cases. b/314977849

Adds note that browsers may choose to enable debugging for specific use-cases (for example, reporting origins can enable debugging without the cookie check for [Mode B groups during Chrome-facilitated testing](https://developer.chrome.com/docs/privacy-sandbox/chrome-testing/#mode-b)).

### DIFF
--- a/EVENT.md
+++ b/EVENT.md
@@ -668,7 +668,10 @@ set by the reporting origin:
 ```http
 Set-Cookie: ar_debug=1; SameSite=None; Secure; HttpOnly
 ```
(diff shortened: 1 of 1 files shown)


linnan-github:
Please update the description to be more meaningful.

linnan-github (changes requested):
Looks good, please update the PR title, thanks.

linnan-github (approved):

PR #1131, opened 2024-01-02
Remove inaccurate mention of modulus on source_event_id field

apasel422:
Per the spec, source_event_id is parsed as a 64-bit unsigned integer and never adjusted.

### DIFF
--- a/EVENT.md
+++ b/EVENT.md
@@ -460,7 +460,7 @@ following keys:
 
 ### Data Encoding
 
-The source event id and trigger data should be specified in a way that is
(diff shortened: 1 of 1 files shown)


csharrison:
The text doesn't really mention modulus so the PR title is a little confusing. Otherwise LGTM

apasel422:
> The text doesn't really mention modulus so the PR title is a little confusing. Otherwise LGTM

It does mention modulus a few lines later.

csharrison (approved):
My bad should have expanded the PR context!

PR #1132, opened 2024-01-02
Round up to nearest tenths place for information gain limit for event-level navigations

akashnadan:
Clarification of information gain limit for event-level navigations. Information gain has been rounded to the nearest tenths place (11.46 to 11.5) to match the rounding of event-level views


apasel422 (changes requested):
Regarding the PR title, the "flexible" part is irrelevant. This will be the max event-level info gain for all navigation sources.

apasel422 (approved):

PR #1135, opened 2024-01-04
Clean up fake report generation

linnan-github:
Many fields in the trigger is irrelevant of event-level report generation. Removed fakeTrigger and passed the corresponding fields into the algorithm.

***
<a href="https://pr-preview.s3.amazonaws.com/linnan-github/conversion-measurement-api/pull/1135.html" title="Last updated on Jan 4, 2024, 7:07 PM UTC (936a053)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/attribution-reporting-api/1135/be4d34b...linnan-github:936a053.html" title="Last updated on Jan 4, 2024, 7:07 PM UTC (936a053)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2188,37 +2188,9 @@ a [=trigger state=] |triggerState|:
     :: «[ "`source_type`" → « |source|'s [=attribution source/source type=] » ]»
 1. Let |triggerTime| be the greatest [=moment=] that is strictly less than
     |triggerState|'s [=trigger state/report window=]'s [=report window/end=].
-1. Let |fakeTrigger| be a new [=attribution trigger=] with the items:
-    : [=attribution trigger/attribution destination=]
-    :: |source|'s [=attribution source/source site=]
-    : [=attribution trigger/trigger time=]
-    :: |triggerTime|
-    : [=attribution trigger/reporting origin=]
-    :: |source|'s [=attribution source/reporting origin=]
-    : [=attribution trigger/filters=]
-    :: «[]»
-    : [=attribution trigger/debug key=]
-    :: null
-    : [=attribution trigger/event-level trigger configurations=]
-    :: « |fakeConfig| »
-    : [=attribution trigger/aggregatable trigger data=]
-    :: «»
-    : [=attribution trigger/aggregatable values=]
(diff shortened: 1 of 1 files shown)


apasel422:
I don't think this is correct. Per the Chromium link, it's only the *context origin* that is set that way, not the actual attribution destinations.

linnan-github:
> I don't think this is correct. Per the Chromium link, it's only the _context origin_ that is set that way, not the actual attribution destinations.

Yeah the fake trigger's destination is not actually not used for the event-level report, but I feel it's easier to understand.

BTW, this needs to be fixed anyways as the destination is a list.

apasel422:
> Yeah the fake trigger's destination is not actually not used for the event-level report, but I feel it's easier to understand.

I disagree: The context origin only really exists as a concept in the Chromium implementation, as it's needed strictly for data deletion and browser policy.

> BTW, this needs to be fixed anyways as the destination is a list.

You're right, but maybe the better approach is to stop generating a fake *trigger* at all in the spec.

index.bs
```diff
@@ -2190,7 +2190,7 @@ a [=trigger state=] |triggerState|:
     |triggerState|'s [=trigger state/report window=]'s [=report window/end=].
 1. Let |fakeTrigger| be a new [=attribution trigger=] with the items:
     : [=attribution trigger/attribution destinations=]
```
apasel422, line 2192:
```suggestion
    : [=attribution trigger/attribution destination=]
```

linnan-github:
> > Yeah the fake trigger's destination is not actually not used for the event-level report, but I feel it's easier to understand.
> 
> I disagree: The context origin only really exists as a concept in the Chromium implementation, as it's needed strictly for data deletion and browser policy.
> 
> > BTW, this needs to be fixed anyways as the destination is a list.
> 
> You're right, but maybe the better approach is to stop generating a fake _trigger_ at all in the spec.

Yeah that's fair. Maybe we can update "obtain an event-level report" algorithm to take trigger time and trigger debug key instead of a trigger.

apasel422 (approved):

PR #1149, opened 2024-01-24
Allow trigger side filtering on aggregation keys

### LINKED ISSUE #649: Allow filtering aggregation keys
Currently, the API only supports filtering on key_pieces, rather than keys themselves. This means that you cannot control what aggregation keys get recorded for a trigger based on cross-site data, but instead only source side data.

ThomasQuesadilla:
This allows for a list of dictionaries to be taken in for the `aggregatable_values` field in trigger registrations, where each dictionary contains a dictionary of aggregatable values as well as a pair of filters, allowing for filtering aggregation keys on the trigger side. This change is backwards compatible and maintains original behavior as well.

Resolves #649.

***
<a href="https://pr-preview.s3.amazonaws.com/WICG/attribution-reporting-api/pull/1149.html" title="Last updated on Feb 12, 2024, 4:34 PM UTC (f1ff596)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/attribution-reporting-api/1149/5ec6061...f1ff596.html" title="Last updated on Feb 12, 2024, 4:34 PM UTC (f1ff596)">Diff</a>

### DIFF
--- a/AGGREGATE.md
+++ b/AGGREGATE.md
@@ -188,12 +188,13 @@ to it just like for event-level reports.
 
 Note: The `aggregatable_values` field may also be specified as a list of dict,
 where each dict contains a dict of key-value pair as outlined above as well as
-an optional `filters` (and its corresponding `not_filters`) field, allowing
-trigger registrations to customize how values are contributed to keys depending
-on the context of the source.
+optional `filters` and `not_filters` fields, allowing trigger registrations to
+customize how values are contributed to keys depending on source event data. If
+multiple list entries have filters that match the source's filters, only the
+first entry and its corresponding values will be used.
 ```jsonc
 {
-  ...
+  ...,
   "aggregatable_values": [
     {
       "values": {
--- a/AGGREGATE.md
+++ b/AGGREGATE.md
@@ -186,12 +186,15 @@ Note: The `filters` field will still apply to aggregatable reports, and each
 dict in `aggregatable_trigger_data` can still optionally have filters applied
 to it just like for event-level reports.
 
-Note: The `aggregatable_values` field may also be specified as a list of dict,
-where each dict contains a dict of key-value pair as outlined above as well as
-optional `filters` and `not_filters` fields, allowing trigger registrations to
-customize how values are contributed to keys depending on source event data. If
-multiple list entries have filters that match the source's filters, only the
-first entry and its corresponding values will be used.
+Note: The `aggregatable_values` field may also be specified as a list of
+dictionaries, where each dictionary contains a dictionary `values` of key-value
+pairs as outlined above as well as optional `filters` and `not_filters` fields,
+allowing trigger registrations to customize how values are contributed to keys
+depending on source filter data. If multiple list entries have filters that
+match the source's filters, only the first entry and its corresponding values
+will be used.
+
+For example:
 ```jsonc
 {
   ...,
--- a/AGGREGATE.md
+++ b/AGGREGATE.md
@@ -172,9 +172,6 @@ The scheme above will generate the following abstract histogram contributions:
   value: 1664
 }]
 ```
-Note: The `filters` field will still apply to aggregatable reports, and each
-dict in `aggregatable_trigger_data` can still optionally have filters applied
-to it just like for event-level reports.
 
 Note: the above scheme was used to maximize the [contribution
 budget](#contribution-bounding-and-budgeting) and optimize utility in the face
@@ -183,6 +180,30 @@ of constant noise. To rescale, simply inverse the scaling factors used above:
 L1 = 1 << 16
 true_agg_campaign_counts = raw_agg_campaign_counts / (L1 / 2)
 true_agg_geo_value = 1024 * raw_agg_geo_value / (L1 / 2)
+```
+
+Note: The `filters` field will still apply to aggregatable reports, and each
+dict in `aggregatable_trigger_data` can still optionally have filters applied
+to it just like for event-level reports.
+
+Note: The `aggregatable_values` field may also be specified as a list of dict,
+where each dict contains a dict of key-value pair as outlined above as well as
+an optional `filters` (and its corresponding `not_filters`) field, allowing
+trigger registrations to customize how values are contributed to keys depending
+on the context of the source.
+```jsonc
+{
+  ...
+  "aggregatable_values": [
+    {
+      "values": {
+        "campaignCounts": 32768,
+        "geoValue": 1664
+      },
+      "filters": {"source_type": ["navigation"]}
+    }
+  ]
+}
 ```
 
 Trigger registration will accept an optional field
--- a/index.bs
+++ b/index.bs
@@ -2715,9 +2715,9 @@ To <dfn>create [=aggregatable contributions=]</dfn> given an [=attribution sourc
     1. If the result of running [=match an attribution source's filter data against filters and negated filters=] with
         |source|, |aggregatableValuesConfiguration|'s [=aggregatable values configuration/filters=],
         |aggregatableValuesConfiguration|'s [=aggregatable values configuration/negated filters=], and
-        |trigger|'s [=attribution trigger/trigger time=] is true: 
+        |trigger|'s [=attribution trigger/trigger time=] is true:
+        1. Let |aggregatableValues| be |aggregatableValuesConfiguration|'s [=aggregatable values configuration/values=].
         1. [=map/iterate|For each=] |id| → |key| of |aggregationKeys|:
-            1. Let |aggregatableValues| be |aggregatableValuesConfiguration|'s [=aggregatable values configuration/values=].
             1. If |aggregatableValues|[|id|] does not [=map/exist=], [=iteration/continue=].
             1. Let |contribution| be a new [=aggregatable contribution=] with the items:
                 : [=aggregatable contribution/key=]
@@ -2725,7 +2725,7 @@ To <dfn>create [=aggregatable contributions=]</dfn> given an [=attribution sourc
                 : [=aggregatable contribution/value=]
                 :: |aggregatableValues|[|id|]
             1. [=list/Append=] |contribution| to |contributions|.
-        1. If |contributions| is not [=list/is empty|empty=], [=iteration/break=].
+        1. [=iteration/Break=].
 1. Return |contributions|.
 
 <h3 id="can-source-create-aggregatable-contributions">Can source create aggregatable contributions</h3>
--- a/index.bs
+++ b/index.bs
@@ -1209,7 +1209,7 @@ controls the maximum [=map/size=] of an [=attribution source=]'s
 
 <dfn>Max length per aggregation key identifier</dfn> is a positive integer that controls
 the maximum [=string/length=] of an [=attribution source=]'s [=attribution source/aggregation keys=]'s
-[=map/keys=], an [=attribution trigger=]'s [=attribution trigger/aggregatable values configurations=]'s item's 
+[=map/keys=], an [=attribution trigger=]'s [=attribution trigger/aggregatable values configurations=]'s [=list/item=]'s 
 [=aggregatable values configuration/values=]'s [=map/keys=], and an [=aggregatable trigger data=]'s 
 [=aggregatable trigger data/source keys=]'s [=set/items=]. Its value is 25.
 
@@ -2427,7 +2427,7 @@ To <dfn>parse aggregatable values</dfn> given an [=ordered map=] |map|:
 1. If |map|["`aggregatable_values`"] does not [=map/exist=], return a new [=list/is empty|empty=] [=list=].
 1. Let |values| be |map|["`aggregatable_values`"].
 1. If |values| is not an [=ordered map=] or a [=list=], return null.
-1. Let |aggregatableValuesConfigurations| be a [=list=] of [=aggregatable values configuration=], initially empty.
+1. Let |aggregatableValuesConfigurations| be a [=list=] of [=aggregatable values configurations=], initially empty.
 1. If |values| is an [=ordered map=]:
     1. Let |aggregatableKeyValues| be the result of running [=parse aggregatable key-values=] with |values|.
     1. If |aggregatableKeyValues| is null, return null.
@@ -2695,8 +2695,8 @@ Given an [=attribution trigger=] |trigger|, an [=attribution source=]
 
 <h3 algorithm id="creating-aggregatable-contributions">Creating aggregatable contributions</h3>
 
-To <dfn>match [=aggregatable contributions=] to [=attribution source/aggregation keys=]</dfn> given
- an [=ordered map=] |aggregationKeys| and an [=ordered map=] |aggregatableValues|,
+To <dfn>create [=aggregatable contributions=] from [=attribution source/aggregation keys=] and aggregatable values</dfn>
+ given an [=ordered map=] |aggregationKeys| and an [=ordered map=] |aggregatableValues|,
  run the following steps:
 
 1. Let |contributions| be an empty [=list=].
@@ -2730,7 +2730,7 @@ To <dfn>create [=aggregatable contributions=]</dfn> given an [=attribution sourc
         |source|, |aggregatableValuesConfiguration|'s [=aggregatable values configuration/filters=],
         |aggregatableValuesConfiguration|'s [=aggregatable values configuration/negated filters=], and
         |trigger|'s [=attribution trigger/trigger time=] is true:
-        1. Return the result of running [=match aggregatable contributions to aggregation keys=] with
+        1. Return the result of running [=create aggregatable contributions from aggregation keys and aggregatable values=] with
             |aggregationKeys| and |aggregatableValuesConfiguration|'s [=aggregatable values configuration/values=].
 1. Return a new [=list/is empty|empty=] [=list=].
 
--- a/index.bs
+++ b/index.bs
@@ -2696,7 +2696,7 @@ Given an [=attribution trigger=] |trigger|, an [=attribution source=]
 <h3 algorithm id="creating-aggregatable-contributions">Creating aggregatable contributions</h3>
 
 To <dfn>create [=aggregatable contributions=] from [=attribution source/aggregation keys=] and aggregatable [=aggregatable values configuration/values=]</dfn>
- given an [=ordered map=] |aggregationKeys| and a [=map=] |aggregatableValues|,
+ given a [=map=] |aggregationKeys| and a [=map=] |aggregatableValues|,
  run the following steps:
 
 1. Let |contributions| be an empty [=list=].
--- a/index.bs
+++ b/index.bs
@@ -797,6 +797,18 @@ An aggregatable trigger data is a [=struct=] with the following items:
 
 </dl>
 
+<h3 dfn-type=dfn>Aggregatable value</h3>
+
+An aggregatable value is a [=struct=] with the following items:
+
+<dl dfn-for="aggregatable value">
+: <dfn>values</dfn>
+:: An [=ordered map=] whose [=map/key|keys=] are [=strings=] and whose [=map/value|values=] are non-negative 32-bit integers.
+: <dfn>filters</dfn>
+:: A [=list=] of [=filter configs=].
+: <dfn>negated filters</dfn>
+:: A [=list=] of [=filter configs=].
+
 <h3 dfn-type=dfn>Aggregatable dedup key</h3>
 
 An aggregatable dedup key is a [=struct=] with the following items:
@@ -869,8 +881,7 @@ An attribution trigger is a [=struct=] with the following items:
 : <dfn>aggregatable trigger data</dfn>
 :: A [=list=] of [=aggregatable trigger data=].
 : <dfn>aggregatable values</dfn>
-:: An [=ordered map=] whose [=map/key|keys=] are [=strings=] and whose
-    [=map/value|values=] are non-negative 32-bit integers.
+:: A [=list=] of [=aggregatable value=].
 : <dfn>aggregatable dedup keys</dfn>
 :: A [=list=] of [=aggregatable dedup key=].
 : <dfn>verifications</dfn>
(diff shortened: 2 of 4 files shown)


index.bs
```diff
@@ -2665,14 +2710,21 @@ To <dfn>create [=aggregatable contributions=]</dfn> given an [=attribution sourc
...
+            |aggregatableValue|'s [=aggregatable value/negated filters=], and
+            |trigger|'s [=attribution trigger/trigger time=]
+            is false, [=iteration/continue=].
+        1. Let |contribution| be a new [=aggregatable contribution=] with the items:
+            : [=aggregatable contribution/key=]
+            :: |key|
+            : [=aggregatable contribution/value=]
+            :: |aggregatableValues|[|id|]
+        1. [=list/Append=] |contribution| to |contributions|.
+    1. If |contributions| is not [=list/is empty|empty=], [=iteration/Break=].
```
apasel422, line 2727: It might be easier to follow the loop logic here by performing the filtering step before iterating over `aggregationKeys`. Then I think the `break` could be omitted.

IIUC, we're trying to find the first `aggregatableValue` whose filters match, and then applying it. The loop structure as written makes it seems like multiple `aggregatableValues` structures are being combined.

ts/src/header-validator/validate-json.ts
```diff
@@ -1194,6 +1194,10 @@ function aggregatableTriggerData(
   )
 }
 
+export type AggregatableValue = FilterPair & {
```
apasel422, line 1197: Let's rename this type to whatever we end up going with in the spec per my other comment.

ts/src/header-validator/validate-json.ts
```diff
@@ -1194,6 +1194,10 @@ function aggregatableTriggerData(
...
+  aggregatableValue: Map<string, number>
```
apasel422, line 1198: Maybe this should be called `values` for parity with the spec?

ts/src/header-validator/validate-json.ts
```diff
@@ -1208,8 +1212,34 @@ function aggregatableKeyValue(
     )
 }
 
-function aggregatableValues(ctx: Context, j: Json): Maybe<Map<string, number>> {
-  return keyValues(ctx, j, aggregatableKeyValue)
+function aggregatableValues(ctx: Context, j: Json): Maybe<AggregatableValue[]> {
+  return typeSwitch(ctx, j, {
+    object: (ctx, j) =>
+      struct(
```
apasel422, line 1218: I don't think it's correct to use `struct` here: If `j` is an object, then `j` directly contains the key-value pairs. I'd keep `function aggregatableValues` but rename it to `function aggregatableKeyValues` and then make this function like:

```typescript
function aggregatablevalues(ctx: Context, j: Json): Maybe<AggregatableValue[]> {
  return typeSwitch(ctx, j, {
    object: (ctx, j) => aggregatableKeyValues(ctx, j).map((values) => ({ values, filters: [], notFilters: []})),
    list: ... // what you currently have, but using `aggregatableKeyValues` instead of `keyValues`
```

ts/src/header-validator/validate-json.ts
```diff
@@ -1208,8 +1212,34 @@ function aggregatableKeyValue(
...
+        struct(
+          ctx,
+          j,
+          {
+            aggregatableValue: field('values', (ctx, j) =>
+              keyValues(ctx, j, aggregatableKeyValue)
+            ),
+            ...filterFields,
+          },
+          /*warnUnknown=*/ false
```
apasel422, line 1239: We actually do want to warn about unknown fields here, as this is a struct with 3 known fields (`values, `filters`, `not_filters`).

ts/src/header-validator/validate-json.ts
```diff
@@ -1276,39 +1306,40 @@ function aggregatableSourceRegistrationTime(
...
-          ctx.warning(
-            'absence from aggregatable_trigger_data source_keys equivalent to presence with key_piece 0x0'
-          )
-        )
-      }
-    }
-  })
-}
+// TODO(apasel422): Update with new AggregatableValue structure.
+// function warnInconsistentAggregatableKeys(ctx: Context, t: Trigger): void {
```
apasel422, line 1310: Let's keep the function declaration uncommented but leave its body commented using `/* ... */` and continue calling it on line 1426.

apasel422 (changes requested):

index.bs
```diff
@@ -797,6 +797,20 @@ An aggregatable trigger data is a [=struct=] with the following items:
 
 </dl>
 
+<h3 dfn-type=dfn>Aggregatable value</h3>
+
+An aggregatable value is a [=struct=] with the following items:
```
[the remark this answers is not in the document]
linnan-github: As it contains more than a map, maybe something like `aggregatable values configuration`?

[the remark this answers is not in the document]
ThomasQuesadilla: Went ahead with `aggregatable values configuration`, I changed the name of the respective field in `attribution trigger` as we;; but thinking about it I'm not sure if that should be done, what do yall think?

index.bs
ThomasQuesadilla: I made the requested changes but I'm unsure that A) if I should leave the nesting how it currently is (I didn't want a `continue` in two different places) and B) that the break should be omitted, IIUC we wouldn't want to continue looping through every entry in the `aggregatableValuesConfigurations` list if we find an entry with matching filters but my understanding can be wrong.

apasel422: Ack. Maybe @linnan-github can suggest a restructuring here.

index.bs
```diff
@@ -2663,16 +2709,23 @@ To <dfn>create [=aggregatable contributions=]</dfn> given an [=attribution sourc
...
+        1. [=map/iterate|For each=] |id| → |key| of |aggregationKeys|:
+            1. Let |aggregatableValues| be |aggregatableValuesConfiguration|'s [=aggregatable values configuration/values=].
+            1. If |aggregatableValues|[|id|] does not [=map/exist=], [=iteration/continue=].
+            1. Let |contribution| be a new [=aggregatable contribution=] with the items:
+                : [=aggregatable contribution/key=]
+                :: |key|
+                : [=aggregatable contribution/value=]
+                :: |aggregatableValues|[|id|]
+            1. [=list/Append=] |contribution| to |contributions|.
+        1. If |contributions| is not [=list/is empty|empty=], [=iteration/Break=].
```
apasel422, line 2728:
```suggestion
        1. If |contributions| is not [=list/is empty|empty=], [=iteration/break=].
```

ts/src/header-validator/trigger.test.ts
```diff
@@ -364,41 +391,80 @@ const testCases: jsontest.TestCase<Trigger>[] = [
...
+  //     {
+  //       path: ['aggregatable_trigger_data', 1, 'source_keys'],
+  //       msg: 'key "a" will never result in a contribution due to absence from aggregatable_values',
+  //     },
+  //     {
+  //       path: ['aggregatable_values', 'c'],
+  //       msg: 'absence from aggregatable_trigger_data source_keys equivalent to presence with key_piece 0x0',
+  //     },
+  //   ],
+  // },
```
apasel422, line 466: Would it be possible to comment this out using `/* ... */` instead?

apasel422 (changes requested):
Generally LGTM.

index.bs
```diff
@@ -1196,7 +1209,7 @@ controls the maximum [=map/size=] of an [=attribution source=]'s
 
 <dfn>Max length per aggregation key identifier</dfn> is a positive integer that controls
 the maximum [=string/length=] of an [=attribution source=]'s [=attribution source/aggregation keys=]'s
-[=map/keys=], an [=attribution trigger=]'s [=attribution trigger/aggregatable values=]'s [=map/keys=],
+[=map/keys=], an [=attribution trigger=]'s [=attribution trigger/aggregatable values configurations=]'s [=map/keys=],
```
linnan-github, line 1212: I think this is not right. It should be the keys of configurations's items's values field.

index.bs
```diff
@@ -2399,18 +2412,51 @@ To <dfn>parse aggregatable trigger data</dfn> given an [=ordered map=] |map|:
...
+    1. Let |aggregatableValuesConfiguration| be a new [=aggregatable values configuration=] with the items:
+        : [=aggregatable values configuration/values=]
+        :: |aggregatableKeyValues|
+        : [=aggregatable values configuration/filters=]
+        :: «[]»
+        : [=aggregatable values configuration/negated filters=]
+        :: «[]»
+    1. [=list/Append=] |aggregatableValuesConfiguration| to |aggregatableValuesConfigurations|.
+    1. Return |aggregatableValuesConfigurations|.
+1. [=list/iterate|For each=] |value| of |values|:
```
linnan-github, line 2443: Do we allow an empty list?

apasel422: In my opinion we should.

index.bs
```diff
@@ -2663,16 +2709,23 @@ To <dfn>create [=aggregatable contributions=]</dfn> given an [=attribution sourc
...
+        1. [=map/iterate|For each=] |id| → |key| of |aggregationKeys|:
+            1. Let |aggregatableValues| be |aggregatableValuesConfiguration|'s [=aggregatable values configuration/values=].
+            1. If |aggregatableValues|[|id|] does not [=map/exist=], [=iteration/continue=].
+            1. Let |contribution| be a new [=aggregatable contribution=] with the items:
+                : [=aggregatable contribution/key=]
+                :: |key|
+                : [=aggregatable contribution/value=]
+                :: |aggregatableValues|[|id|]
+            1. [=list/Append=] |contribution| to |contributions|.
+        1. If |contributions| is not [=list/is empty|empty=], [=iteration/break=].
```
linnan-github, line 2728: To align with the algorithm for aggregatable debug keys, I think we should break whenever the filters match, regardless of whether the contributions is empty or not.

index.bs
linnan-github: Yeah given that we allow empty `event_trigger_data` and `aggregatable_trigger_data`, we should allow that.

index.bs
```diff
@@ -2663,16 +2709,23 @@ To <dfn>create [=aggregatable contributions=]</dfn> given an [=attribution sourc
...
-        : [=aggregatable contribution/value=]
-        :: |aggregatableValues|[|id|]
-    1. [=list/Append=] |contribution| to |contributions|.
+1. [=list/iterate|For each=] |aggregatableValuesConfiguration| of |aggregatableValuesConfigurations|:
+    1. If the result of running [=match an attribution source's filter data against filters and negated filters=] with
+        |source|, |aggregatableValuesConfiguration|'s [=aggregatable values configuration/filters=],
+        |aggregatableValuesConfiguration|'s [=aggregatable values configuration/negated filters=], and
+        |trigger|'s [=attribution trigger/trigger time=] is true: 
+        1. [=map/iterate|For each=] |id| → |key| of |aggregationKeys|:
+            1. Let |aggregatableValues| be |aggregatableValuesConfiguration|'s [=aggregatable values configuration/values=].
```
linnan-github, line 2720: This step can be moved outside the loop.

index.bs
linnan-github: It's probably slightly clearer to add a sub algorithm to find the aggregatable values that matches, so that we can avoid two loops here.

index.bs
ThomasQuesadilla: Nice catch!

index.bs
ThomasQuesadilla: Added a sub algorithm and refactored the main algorithm a bit to clean it up.

ts/src/header-validator/validate-json.ts
```diff
@@ -1208,10 +1212,32 @@ function aggregatableKeyValue(
...
+): Maybe<AggregatableValuesConfiguration[]> {
+  return typeSwitch(ctx, j, {
+    object: (ctx, j) =>
+      aggregatableKeyValues(ctx, j).map((values) => [
+        { values, positive: [], negative: [] },
+      ]),
+    list: (ctx, j) =>
+      array(ctx, j, (ctx, j) =>
+        struct(ctx, j, {
+          values: field('values', (ctx, j) => aggregatableKeyValues(ctx, j)),
```
apasel422, line 1234:
```suggestion
          values: field('values', aggregatableKeyValues),
```

index.bs
```diff
@@ -2399,18 +2412,51 @@ To <dfn>parse aggregatable trigger data</dfn> given an [=ordered map=] |map|:
...
+        :: «[]»
+    1. [=list/Append=] |aggregatableValuesConfiguration| to |aggregatableValuesConfigurations|.
+    1. Return |aggregatableValuesConfigurations|.
+1. [=list/iterate|For each=] |value| of |values|:
+    1. If |value| is not an [=ordered map=], return null.
+    1. If |value|["`values`"] does not [=map/exist=], return null.
+    1. Let |aggregatableKeyValues| be the result of running [=parse aggregatable key-values=] with |value|["`values`"].
+    1. If |aggregatableKeyValues| is null, return null.
+    1. Let |filterPair| be the result of running [=parse a filter pair=] with
+        |aggregatableKeyValues|.
```
apasel422, line 2449: Shouldn't this be `|value|`? `|aggregatableKeyValues|` is the result of parsing the `"values"` field.

index.bs
```diff
@@ -2649,6 +2695,21 @@ Given an [=attribution trigger=] |trigger|, an [=attribution source=]
 
 <h3 algorithm id="creating-aggregatable-contributions">Creating aggregatable contributions</h3>
 
+To <dfn>match [=aggregatable contributions=] to [=attribution source/aggregation keys=]</dfn> given
+ an [=ordered map=] |aggregationKeys| and an [=ordered map=] |aggregatableValues|,
+ run the following steps:
+
+1. Let |contributions| be an empty list.
```
apasel422, line 2702:
```suggestion
1. Let |contributions| be an empty [=list=].
```

index.bs
```diff
@@ -2663,17 +2724,15 @@ To <dfn>create [=aggregatable contributions=]</dfn> given an [=attribution sourc
...
-1. Return |contributions|.
+1. Let |aggregatableValuesConfigurations| be |trigger|'s [=attribution trigger/aggregatable values configurations=].
+1. [=list/iterate|For each=] |aggregatableValuesConfiguration| of |aggregatableValuesConfigurations|:
+    1. If the result of running [=match an attribution source's filter data against filters and negated filters=] with
+        |source|, |aggregatableValuesConfiguration|'s [=aggregatable values configuration/filters=],
+        |aggregatableValuesConfiguration|'s [=aggregatable values configuration/negated filters=], and
+        |trigger|'s [=attribution trigger/trigger time=] is true:
+        1. Return the result of running [=match aggregatable contributions to aggregation keys=] with
+            |aggregationKeys| and |aggregatableValuesConfiguration|'s [=aggregatable values configuration/values=].
+1. Return «[]».
```
apasel422, line 2735: If the intent is to return an empty list here, the `[]` between the angle quotes is syntactically invalid: https://infra.spec.whatwg.org/#lists. It's probably clearer though to just say `Return a new [=list/is empty=|empty=] [=list=].`

index.bs
```diff
@@ -3019,7 +3078,7 @@ To <dfn>check if an [=attribution trigger=] contains aggregatable data</dfn> giv
 run the following steps:
 
 1. If |trigger|'s [=attribution trigger/aggregatable trigger data=] is not [=list/is empty|empty=], return true.
-1. If |trigger|'s [=attribution trigger/aggregatable values=] is not [=map/is empty|empty=], return true.
+1. If |trigger|'s [=attribution trigger/aggregatable values configurations=] is not [=map/is empty|empty=], return true.
```
apasel422, line 3081:
```suggestion
1. If |trigger|'s [=attribution trigger/aggregatable values configurations=] is not [=list/is empty|empty=], return true.
```

apasel422 (changes requested):

index.bs
ThomasQuesadilla: Ah good to know, adjusted the other sites where this was declared.

index.bs
ThomasQuesadilla: Completely missed that, good catch!

index.bs
```diff
@@ -1196,7 +1209,7 @@ controls the maximum [=map/size=] of an [=attribution source=]'s
 
 <dfn>Max length per aggregation key identifier</dfn> is a positive integer that controls
 the maximum [=string/length=] of an [=attribution source=]'s [=attribution source/aggregation keys=]'s
-[=map/keys=], an [=attribution trigger=]'s [=attribution trigger/aggregatable values=]'s [=map/keys=],
+[=map/keys=], an [=attribution trigger=]'s [=attribution trigger/aggregatable values configurations=]'s [=aggregatable values configuration/values=]'s [=map/keys=],
```
linnan-github, line 1212: Should it be aggregatable values configurations's item's...?

ThomasQuesadilla: Ah right good point.

index.bs
```diff
@@ -2663,17 +2724,15 @@ To <dfn>create [=aggregatable contributions=]</dfn> given an [=attribution sourc
...
-    1. Let |contribution| be a new [=aggregatable contribution=] with the items:
-        : [=aggregatable contribution/key=]
-        :: |key|
-        : [=aggregatable contribution/value=]
-        :: |aggregatableValues|[|id|]
-    1. [=list/Append=] |contribution| to |contributions|.
-1. Return |contributions|.
+1. Let |aggregatableValuesConfigurations| be |trigger|'s [=attribution trigger/aggregatable values configurations=].
+1. [=list/iterate|For each=] |aggregatableValuesConfiguration| of |aggregatableValuesConfigurations|:
+    1. If the result of running [=match an attribution source's filter data against filters and negated filters=] with
```
apasel422, line 2729: This will need to be rebased due to #1153.

ThomasQuesadilla: Thanks for letting me know!

ThomasQuesadilla:
The change has been approved by the measurement review team.

AGGREGATE.md
```diff
@@ -185,6 +182,30 @@ true_agg_campaign_counts = raw_agg_campaign_counts / (L1 / 2)
 true_agg_geo_value = 1024 * raw_agg_geo_value / (L1 / 2)
 ```
 
+Note: The `filters` field will still apply to aggregatable reports, and each
+dict in `aggregatable_trigger_data` can still optionally have filters applied
+to it just like for event-level reports.
+
+Note: The `aggregatable_values` field may also be specified as a list of dict,
+where each dict contains a dict of key-value pair as outlined above as well as
+an optional `filters` (and its corresponding `not_filters`) field, allowing
```
linnan-github, line 191:
```suggestion
an optional `filters` and an optional `not_filters` fields, allowing
```

AGGREGATE.md
```diff
@@ -185,6 +182,30 @@ true_agg_campaign_counts = raw_agg_campaign_counts / (L1 / 2)
...
+trigger registrations to customize how values are contributed to keys depending
+on the context of the source.
```
linnan-github, line 193:
```suggestion
on source event data.
```

AGGREGATE.md
```diff
@@ -185,6 +182,30 @@ true_agg_campaign_counts = raw_agg_campaign_counts / (L1 / 2)
...
+```jsonc
+{
+  ...
```
linnan-github, line 196:
```suggestion
  ...,
```

AGGREGATE.md
linnan-github, line 193: Please also mention that if multiple values match, then the first matching one will be used.

index.bs
```diff
@@ -1196,9 +1209,9 @@ controls the maximum [=map/size=] of an [=attribution source=]'s
 
 <dfn>Max length per aggregation key identifier</dfn> is a positive integer that controls
 the maximum [=string/length=] of an [=attribution source=]'s [=attribution source/aggregation keys=]'s
-[=map/keys=], an [=attribution trigger=]'s [=attribution trigger/aggregatable values=]'s [=map/keys=],
-and an [=aggregatable trigger data=]'s [=aggregatable trigger data/source keys=]'s [=set/items=].
-Its value is 25.
+[=map/keys=], an [=attribution trigger=]'s [=attribution trigger/aggregatable values configurations=]'s item's
```
linnan-github, line 1212:
```suggestion
[=map/keys=], an [=attribution trigger=]'s [=attribution trigger/aggregatable values configurations=]'s [=list/item=]'s 
```

index.bs
```diff
@@ -2399,18 +2412,51 @@ To <dfn>parse aggregatable trigger data</dfn> given an [=ordered map=] |map|:
...
-1. If |values| is not an [=ordered map=], return null.
-1. [=map/iterate|For each=] |key| → |value| of |values|:
-     1. If |key|'s [=string/length=] is greater than the [=max length per aggregation key identifier=],
-         return null.
-     1. If |value| is not an integer, return null.
-     1. If |value| is less than or equal to 0, return null.
-     1. If |value| is greater than [=allowed aggregatable budget per source=], return null.
-1. Return |values|.
+1. If |values| is not an [=ordered map=] or a [=list=], return null.
+1. Let |aggregatableValuesConfigurations| be a [=list=] of [=aggregatable values configuration=], initially empty.
```
linnan-github, line 2430:
```suggestion
1. Let |aggregatableValuesConfigurations| be a [=list=] of [=aggregatable values configurations=], initially empty.
```

index.bs
```diff
@@ -2649,6 +2695,21 @@ Given an [=attribution trigger=] |trigger|, an [=attribution source=]
 
 <h3 algorithm id="creating-aggregatable-contributions">Creating aggregatable contributions</h3>
 
+To <dfn>match [=aggregatable contributions=] to [=attribution source/aggregation keys=]</dfn> given
```
linnan-github, line 2698: Wdyt about  renaming this algorithm to `create aggregatable contributions from aggregation keys and aggregatable values`?

AGGREGATE.md
```diff
@@ -185,6 +182,31 @@ true_agg_campaign_counts = raw_agg_campaign_counts / (L1 / 2)
 true_agg_geo_value = 1024 * raw_agg_geo_value / (L1 / 2)
 ```
 
+Note: The `filters` field will still apply to aggregatable reports, and each
+dict in `aggregatable_trigger_data` can still optionally have filters applied
+to it just like for event-level reports.
+
+Note: The `aggregatable_values` field may also be specified as a list of dict,
+where each dict contains a dict of key-value pair as outlined above as well as
```
apasel422, line 190:
```suggestion
Note: The `aggregatable_values` field may also be specified as a list of dictionaries,
where each dictionary contains a dictionary `values` of key-value pairs as outlined above as well as
```

AGGREGATE.md
```diff
@@ -185,6 +182,31 @@ true_agg_campaign_counts = raw_agg_campaign_counts / (L1 / 2)
...
+optional `filters` and `not_filters` fields, allowing trigger registrations to
+customize how values are contributed to keys depending on source event data. If
```
apasel422, line 192:
```suggestion
customize how values are contributed to keys depending on source filter data. If
```

AGGREGATE.md
```diff
@@ -185,6 +182,31 @@ true_agg_campaign_counts = raw_agg_campaign_counts / (L1 / 2)
...
+multiple list entries have filters that match the source's filters, only the
+first entry and its corresponding values will be used.
```
apasel422, line 194:
```suggestion
first entry and its corresponding values will be used.

For example:
```

index.bs
```diff
@@ -797,6 +797,20 @@ An aggregatable trigger data is a [=struct=] with the following items:
 
 </dl>
 
+<h3 dfn-type=dfn>Aggregatable values configuration</h3>
+
+An aggregatable values configuration is a [=struct=] with the following items:
+
+<dl dfn-for="aggregatable values configuration">
+: <dfn>values</dfn>
+:: An [=ordered map=] whose [=map/key|keys=] are [=strings=] and whose [=map/value|values=] are non-negative 32-bit integers.
```
apasel422, line 806:
```suggestion
:: A [=map=] whose [=map/key|keys=] are [=strings=] and whose [=map/value|values=] are non-negative 32-bit integers.
```
Since the order is irrelevant.

index.bs
```diff
@@ -2399,18 +2412,51 @@ To <dfn>parse aggregatable trigger data</dfn> given an [=ordered map=] |map|:
     1. [=list/Append=] |aggregatableTrigger| to |aggregatableTriggerData|.
 1. Return |aggregatableTriggerData|.
 
+To <dfn>parse aggregatable key-values</dfn> given an [=ordered map=] |map|:
```
apasel422, line 2415:
```suggestion
To <dfn>parse aggregatable key-values</dfn> given a [=map=] |map|:
```
Since the order is irrelevant.

index.bs
```diff
@@ -2399,18 +2412,51 @@ To <dfn>parse aggregatable trigger data</dfn> given an [=ordered map=] |map|:
...
-1. [=map/iterate|For each=] |key| → |value| of |values|:
-     1. If |key|'s [=string/length=] is greater than the [=max length per aggregation key identifier=],
-         return null.
-     1. If |value| is not an integer, return null.
-     1. If |value| is less than or equal to 0, return null.
-     1. If |value| is greater than [=allowed aggregatable budget per source=], return null.
-1. Return |values|.
+1. If |values| is not an [=ordered map=] or a [=list=], return null.
+1. Let |aggregatableValuesConfigurations| be a [=list=] of [=aggregatable values configurations=], initially empty.
+1. If |values| is an [=ordered map=]:
```
apasel422, line 2431:
```suggestion
1. If |values| is a [=map=]:
```

index.bs
```diff
@@ -2399,18 +2412,51 @@ To <dfn>parse aggregatable trigger data</dfn> given an [=ordered map=] |map|:
...
+        : [=aggregatable values configuration/values=]
+        :: |aggregatableKeyValues|
+        : [=aggregatable values configuration/filters=]
+        :: «»
+        : [=aggregatable values configuration/negated filters=]
+        :: «»
+    1. [=list/Append=] |aggregatableValuesConfiguration| to |aggregatableValuesConfigurations|.
+    1. Return |aggregatableValuesConfigurations|.
+1. [=list/iterate|For each=] |value| of |values|:
+    1. If |value| is not an [=ordered map=], return null.
```
apasel422, line 2444:
```suggestion
    1. If |value| is not a [=map=], return null.
```

index.bs
```diff
@@ -2649,6 +2695,21 @@ Given an [=attribution trigger=] |trigger|, an [=attribution source=]
 
 <h3 algorithm id="creating-aggregatable-contributions">Creating aggregatable contributions</h3>
 
+To <dfn>create [=aggregatable contributions=] from [=attribution source/aggregation keys=] and aggregatable values</dfn>
+ given an [=ordered map=] |aggregationKeys| and an [=ordered map=] |aggregatableValues|,
```
apasel422, line 2699:
```suggestion
 given a [=map=] |aggregationKeys| and a [=map=] |aggregatableValues|,
```

apasel422 (changes requested):

index.bs
ThomasQuesadilla: Sounds good to me.

index.bs
```diff
@@ -2649,6 +2695,21 @@ Given an [=attribution trigger=] |trigger|, an [=attribution source=]
 
 <h3 algorithm id="creating-aggregatable-contributions">Creating aggregatable contributions</h3>
 
+To <dfn>create [=aggregatable contributions=] from [=attribution source/aggregation keys=] and aggregatable values</dfn>
+ given an [=ordered map=] |aggregationKeys| and a [=map=] |aggregatableValues|,
```
apasel422, line 2699:
```suggestion
 given a [=map=] |aggregationKeys| and a [=map=] |aggregatableValues|,
```

apasel422 (approved):

linnan-github (approved):

PR #1162, opened 2024-02-09
Spec change for supporting preferred platform for cross app attribution

linnan-github:


index.bs
```diff
@@ -1470,6 +1471,24 @@ To <dfn>parse a filter pair</dfn> given a [=map=] |map|:
...
+<h3 id="getting-preferred-platform">Getting preferred platform</h3>
+
+To <dfn>get preferred platform from a header list</dfn> given a
+[=header list=] |headers|:
+
+1. If |headers| does not [=header list/contain=] "`Attribution-Reporting-Info`", return null.
+1. Let |map| be the result of
+    [=header list/get a structured field value|getting=] "`Attribution-Reporting-Info`" from |headers|
+    with a type of "`dictionary`".
+1. If |map| is not an [=ordered map=], return an error.
```
apasel422, line 1483:
```suggestion
1. If |map| is not an [=map=], return an error.
```
Since the order is irrelevant.

index.bs
```diff
@@ -1470,6 +1471,24 @@ To <dfn>parse a filter pair</dfn> given a [=map=] |map|:
...
+1. If |map|["`preferred_platform`"] does not [=map/exist=], return null.
+1. Let |value| be |map|["`preferred_platform`"][0].
```
apasel422, line 1485:
```suggestion
1. If |map|["`preferred-platform`"] does not [=map/exist=], return null.
1. Let |value| be |map|["`preferred-platform`"][0].
```

index.bs
```diff
@@ -1470,6 +1471,24 @@ To <dfn>parse a filter pair</dfn> given a [=map=] |map|:
...
+1. If |value| is not a [=string=], return an error.
+1. If |value| is not a [=registrar=], return an error.
+1. return |value|.
```
apasel422, line 1488:
```suggestion
1. Return |value|.
```

index.bs
```diff
@@ -1470,6 +1471,24 @@ To <dfn>parse a filter pair</dfn> given a [=map=] |map|:
...
+[=header list=] |headers|:
+
+1. If |headers| does not [=header list/contain=] "`Attribution-Reporting-Info`", return null.
+1. Let |map| be the result of
+    [=header list/get a structured field value|getting=] "`Attribution-Reporting-Info`" from |headers|
+    with a type of "`dictionary`".
+1. If |map| is not an [=ordered map=], return an error.
+1. If |map|["`preferred_platform`"] does not [=map/exist=], return null.
+1. Let |value| be |map|["`preferred_platform`"][0].
+1. If |value| is not a [=string=], return an error.
```
apasel422, line 1486: This step is redundant with the `registrar` check below, especially since we have an issue to ensure this is a token.

index.bs
```diff
@@ -1470,6 +1471,24 @@ To <dfn>parse a filter pair</dfn> given a [=map=] |map|:
...
+1. If |value| is not a [=registrar=], return an error.
+1. return |value|.
+
+Issue: |value| must be a result from [=structured header/parse a token|parsing a token=].
```
apasel422, line 1490: For clarity, maybe something like `Issue: Require |value| to be a token.` and link to the token definition in the structured-header spec.

index.bs
```diff
@@ -1658,6 +1677,37 @@ and a [=response=] |response|:
 1. Run [=process an attribution eligible response=] with |contextOrigin|,
     |eligibility|, and |response|.
 
+To <dfn>get the registration platform</dfn> given |webHeader|,
+|osRegistrations|, and |preferredPlatform|:
```
apasel422, line 1681: Please add types for these parameters.

index.bs
```diff
@@ -1658,6 +1677,37 @@ and a [=response=] |response|:
...
+
+1. If |webHeader| and |osRegistrations| are both null, return null.
+1. If |preferredPlatform| is null:
+    1. If |webHeader| and |osRegistrations| are both not null, return null.
+    1. If |webHeader| is not null and the user agent supports web registrations,
+        return "<code>[=registrar/web=]</code>".
+    1. if |osRegistrations| is not null and the user agent supports OS registrations,
```
apasel422, line 1688:
```suggestion
    1. If |osRegistrations| is not null and the user agent supports OS registrations,
```

index.bs
```diff
@@ -1692,62 +1742,96 @@ To <dfn export>process an attribution eligible response</dfn> given a [=suitable
...
-                |contextOrigin|, |reportingOrigin|, |sourceType|, and [=current wall time=].
-            1. If |source| is not null, [=process an attribution source|process=]
-                |source|.
-        1. If |osSourceRegistrations| is not null and the user agent supports OS
-            registrations:
-              1. Process |osSourceRegistrations| according to an [=implementation-defined=] algorithm.
-              1. Run [=obtain and deliver debug reports on OS registrations=] with
-                  "<code>[=OS debug data type/os-source-delegated=]</code>", |osSourceRegistrations|,
-                  and |contextOrigin|.
+        1. Let |platform| be the result of [=get the registration platform =] with
```
apasel422, line 1754:
```suggestion
        1. Let |platform| be the result of [=get the registration platform=] with
```

index.bs
```diff
@@ -1658,6 +1677,37 @@ and a [=response=] |response|:
...
+      ::
+         1. If |osRegistrations| is null, return null.
+         1. If the user agent supports OS registrations, return "<code>[=registrar/os=]</code>".
+         1. If |webHeader| is not null and the user agent supports web registrations,
+             return "<code>[=registrar/web=]</code>".
+         1. Return null.
+
+    </dl>
+
 To <dfn export>process an attribution eligible response</dfn> given a [=suitable origin=]
```
apasel422, line 1711: I think this algorithm has gotten a bit hard to follow given its length. Can we split out some of the subalgorithms?

ts/src/header-validator/validate-info.test.ts
```diff
@@ -0,0 +1,60 @@
+import * as testutil from './util.test'
+import { validateInfo } from './validate-info'
+
+const tests = [
+  {
+    input: 'preferred-platform=os',
+  },
```
apasel422, line 7: Please also add a case for `web`.

ts/src/header-validator/validate-info.ts
```diff
@@ -0,0 +1,41 @@
+import { Context, ValidationResult } from './context'
+import { parseDictionary } from 'structured-headers'
+import { Token } from 'structured-headers'
```
apasel422, line 3:
```suggestion
import { Token, parseDictionary } from 'structured-headers'
```

apasel422 (changes requested):

PR #1172, opened 2024-02-22
Add KelkooGroup as member of tester list

pmainguet:


csharrison:
The guidelines to add to this list include that you have a github profile affiliation with the business you are adding, which I'm not seeing. Is it possible to add one?

csharrison:
Thanks!

PR #1186, opened 2024-03-08
OS registration header validator - replace invalid urls errors with warnings

agarant:
[The spec](https://wicg.github.io/attribution-reporting-api/#get-os-registrations-from-a-header-value) and [the implementation](https://source.chromium.org/chromium/chromium/src/+/main:components/attribution_reporting/os_registration.cc;l=47?q=os_registration.cc&ss=chromium%2Fchromium%2Fsrc) are ignoring invalid urls but still consider the header to be valid. 

This PR updates the header validator to issue warnings instead of errors when encountering invalid urls.

### DIFF
--- a/ts/src/header-validator/validate-os.ts
+++ b/ts/src/header-validator/validate-os.ts
@@ -3,7 +3,7 @@ import { InnerList, Item, parseList } from 'structured-headers'
 
 function validateURL(ctx: Context, member: InnerList | Item): void {
   if (typeof member[0] !== 'string') {
-    ctx.error('must be a string')
+    ctx.warning('ignored, must be a string')
     return
   }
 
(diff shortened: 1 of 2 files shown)


ts/src/header-validator/validate-os.ts
```diff
@@ -22,15 +22,15 @@ function validateURL(ctx: Context, member: InnerList | Item): void {
       (url.hostname === 'localhost' || url.hostname === '127.0.0.1')
     )
   ) {
-    ctx.error('must contain a potentially trustworthy URL')
+    ctx.warning('ignored, must contain a potentially trustworthy URL')
```
agarant, line 25: This isn't part of the spec, I'm planning on removing this check/warning in a follow up PR.

apasel422 (approved):

ts/src/header-validator/validate-os.ts
agarant: Follow-up PR: https://github.com/WICG/attribution-reporting-api/pull/1198

PR #1193, opened 2024-03-13
Remove irrelevant "ordered" prefix from map and set types

apasel422:
The correctness of our algorithms does not depend on these types being ordered, so it is clearer to omit the prefix.

***
<a href="https://pr-preview.s3.amazonaws.com/apasel422/attribution-reporting-api/pull/1193.html" title="Last updated on May 3, 2024, 1:12 PM UTC (eed4291)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/attribution-reporting-api/1193/1a9478a...apasel422:eed4291.html" title="Last updated on May 3, 2024, 1:12 PM UTC (eed4291)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1435,6 +1435,9 @@ To <dfn>serialize [=event-level report/attribution destinations=]</dfn> |destina
 1. If |destinationStrings|'s [=set/size=] is equal to 1, return |destinationStrings|[0].
 1. Return |destinationStrings|.
 
+Issue: |destinationStrings| should be sorted to avoid revealing the ordering
+of destinations within the original source registration.
+
 To <dfn>check if a scheme is suitable</dfn> given a [=string=] |scheme|:
 
 1. If |scheme| is "`http`" or "`https`", return true.
--- a/index.bs
+++ b/index.bs
@@ -611,9 +611,9 @@ A randomized source response is null or a [=set=] of [=trigger states=].
 
 <h3 id="attribution-filtering">Attribution filtering</h3>
 
-A <dfn>filter value</dfn> is an [=ordered set=] of [=strings=].
+A <dfn>filter value</dfn> is a [=set=] of [=strings=].
 
-A <dfn>filter map</dfn> is an [=ordered map=] whose [=map/key|keys=] are [=strings=] and whose
+A <dfn>filter map</dfn> is a [=map=] whose [=map/key|keys=] are [=strings=] and whose
 [=map/value|values=] are [=filter values=].
 
 A <dfn>filter config</dfn> is a [=struct=] with the following items:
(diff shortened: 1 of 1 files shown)


apasel422:
Ping @linnan-github.

linnan-github (approved):
Sorry missed this one.

index.bs
```diff
@@ -3630,7 +3630,7 @@ An [=aggregatable report=] |report|'s <dfn for="aggregatable report">shared info
 of running the following steps:
 
 1. Let |reportingOrigin| be |report|'s [=aggregatable report/reporting origin=].
-1. Let |sharedInfo| be an [=ordered map=] of the following key/value pairs:
+1. Let |sharedInfo| be a [=map=] of the following key/value pairs:
```
giladbarkan-github, line 3633: Would this affect the comparison of some encrypted value?

apasel422: It wouldn't *affect* anything, since a map *is* an ordered map, but as for whether the ordered nature of the map here is important, it depends on whether https://infra.spec.whatwg.org/#serialize-an-infra-value-to-a-json-string preserves ordering.

index.bs
```diff
@@ -760,7 +760,7 @@ An attribution source is a [=struct=] with the following items:
 : <dfn>event ID</dfn>
 :: A non-negative 64-bit integer.
 : <dfn>attribution destinations</dfn>
-:: An [=ordered set=] of [=sites=].
+:: A [=set=] of [=sites=].
```
giladbarkan-github, line 763: Does this also affect their appearance in the report? I think keeping it ordered in the report makes sens.

index.bs
giladbarkan-github: I meant like for aggregation service, but I think they ignore the field order.

index.bs
apasel422: Again, this change doesn't *affect* anything, but whether it's relevant is a valid question. I don't think we should ascribe any semantics to the ordering of destinations within a report unless or until such semantics are needed.

giladbarkan-github (approved):

index.bs
apasel422: Re my findings in https://github.com/WICG/attribution-reporting-api/pull/1217#discussion_r1589148599 the use of `ordered set` here actually implies that the report's `attribution_destination` field must have the same ordering as the attributed source's `destination` field, which we do not want, as it reveals (minor) extra information from the source. IMO the field should be sorted by site to avoid leaking anything additional here.

Will update to this effect.

index.bs
apasel422: @linnan-github Is the order of the fields relevant here?

index.bs
apasel422: Done.

apasel422:
I will sort attribution destinations in a follow-up PR.

index.bs
linnan-github: The order of these fields are irrelevant and is not impacting privacy.

PR #1204, opened 2024-03-19
Gate all source verbose debug reports behind cookie access

linnan-github:
The privacy-neutral argument would depend on the order of the checks. To reduce complexity, gate all source verbose debug reports behind cookie access.

Note that this change is backwards incompatible as source-destination-limit and source-destination-rate-limit debug reports now require ar_debug cookie. There is no major concern as we expect that ad-techs that enable source verbose debug reports would already set ar_debug cookie.

***
<a href="https://pr-preview.s3.amazonaws.com/linnan-github/conversion-measurement-api/pull/1204.html" title="Last updated on Mar 19, 2024, 11:48 PM UTC (f9e10a6)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/attribution-reporting-api/1204/8501298...linnan-github:f9e10a6.html" title="Last updated on Mar 19, 2024, 11:48 PM UTC (f9e10a6)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2410,28 +2410,11 @@ a [=trigger state=] |triggerState|:
     |triggerState|'s [=trigger state/report window=]'s [=report window/end=].
 1. Return |fakeReport|.
 
-To <dfn>check if debug reporting is allowed</dfn> given a [=source debug data type=] |dataType|
-and a [=boolean=] |debugCookieSet|:
-1. If |dataType| is:
-    <dl class="switch">
-    : "<code>[=source debug data type/source-destination-limit=]</code>"
-    : "<code>[=source debug data type/source-destination-rate-limit=]</code>"
-    :: Return <strong>allowed</strong>.
-    : "<code>[=source debug data type/source-noised=]</code>"
-    : "<code>[=source debug data type/source-storage-limit=]</code>"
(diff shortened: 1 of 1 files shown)


linnan-github:
@arpanah Could you PTAL, thanks!

apasel422 (commented):
Is there a particular issue that is motivating this, or is this just  for spec simplification?

Either way, is there any concern that this is backwards incompatible?

linnan-github:
> Is there a particular issue that is motivating this, or is this just for spec simplification?
> 
> Either way, is there any concern that this is backwards incompatible?

This comes up in the post 3PCD debug reporting discussion for privacy analysis. 

This would be backwards incompatible regarding source-destination-limit and source-destination-rate-limit as they now require cookie access. But I don’t see major concerns with this as we would expect that ad-techs that enable source verbose debug reports would already set ar_debug cookie. Cc @akashnadan

apasel422:
> This would be backwards incompatible [...]

That makes sense, but the PR should all out the compatibility issue.

linnan-github:
> > This would be backwards incompatible [...]
> 
> That makes sense, but the PR should all out the compatibility issue.

Updated in the description.

apasel422 (approved):

PR #1208, opened 2024-03-20
Fix adding source rate-limit records

linnan-github:
The source rate-limit records should not be partially added if any of the destination failed the rate limit check.


index.bs
```diff
@@ -2504,7 +2505,8 @@ To <dfn>process an attribution source</dfn> given an [=attribution source=] |sou
         |rateLimitRecord| is <strong>blocked</strong>:
         1. Run [=obtain and deliver a debug report on source registration=] with "[=source debug data type/source-success=]</code>" and |source|.
         1. Return.
-    1. [=set/Append=] |rateLimitRecord| to the [=attribution rate-limit cache=].
+    1. [=set/Append=] |rateLimitRecord| to |newRateLimitRecords|.
+1. [=set/Append=] |newRateLimitRecords| to the [=attribution rate-limit cache=].
```
apasel422, line 2509: The `set/append` method only works on a single item, not on a set of items.

apasel422 (changes requested):

apasel422 (approved):

PR #1211, opened 2024-03-22
Split attribution rate-limit into separate event & aggregate rate-limits

vikassahu29:
Currently, we have a single rate-limit which counts the overall attribution (either event or aggregate or both) of the trigger. This change separates the attribution counts for event & aggregate for accurate counting of reported conversions. This will facilitate easier report deletions counting as well as accurate counting for [Full Flex proposal](https://github.com/WICG/attribution-reporting-api/blob/main/flexible_event_config.md) where more than 1 event report may be generated per trigger.

#### Deletion Issue:
1. Let's say there is 1 source (event) which is registered at t = 0: S
1. Trigger T1 registered at T = 1 and generates an event report E1 (priority=1) and no aggregatable reports
1. Trigger T2 registered at T = 2 and generates an event report E2 (priority=2) and no aggregatable reports. E2 will replace E1 here due to higher priority.

##### Current Implementation:
We will count this as 2 attribution which is incorrect because E1 is never delivered.

##### With Proposed Changes:
We will count this as 1 event-attribution for E2 and 0 for E1 (rate-limit is deleted when we replace the report)

***
<a href="https://pr-preview.s3.amazonaws.com/vikassahu29/attribution-reporting-api/pull/1211.html" title="Last updated on Apr 4, 2024, 5:23 PM UTC (2b78ed3)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/attribution-reporting-api/1211/c21d9ff...vikassahu29:2b78ed3.html" title="Last updated on Apr 4, 2024, 5:23 PM UTC (2b78ed3)">Diff</a>

### DIFF
--- a/flexible_event_config.md
+++ b/flexible_event_config.md
@@ -296,7 +296,7 @@ The values are summed (to 8) and reported in the following reports after 7 days:
   "trigger_summary_bucket": [5, 9]
 }
 ```
-
+Instead of these triggers being counted as 3 attributions from rate-limit perspective, they will be only counted as 1 attribution because 1 
 In the subsequent 7 days, the following triggers are registered:
 
 ```jsonc
@@ -426,6 +426,10 @@ is dropped.
 ```
 
 We encourage developers to suggest different use cases they may have for this API extension, and we will update this explainer with sample configurations for those use cases.
+
+## Attribution Rate Limit
+
+Attribution rate-limit behavior will be based on the numbers of reports generated rather than number of triggers that lead to report generation. For example: The 1st report in the [Reporting trigger value buckets](#reporting-trigger-value-buckets) example will be counted as 1 contribution towards the [max attribution per rate-limit window](https://wicg.github.io/attribution-reporting-api/#max-attributions-per-rate-limit-window) limit despite it being a result of contributions from 3 triggers.
 
 ## Privacy considerations
 
--- a/flexible_event_config.md
+++ b/flexible_event_config.md
@@ -296,7 +296,6 @@ The values are summed (to 8) and reported in the following reports after 7 days:
   "trigger_summary_bucket": [5, 9]
 }
 ```
-Instead of these triggers being counted as 3 attributions from rate-limit perspective, they will be only counted as 1 attribution because 1 
 In the subsequent 7 days, the following triggers are registered:
 
 ```jsonc
--- a/flexible_event_config.md
+++ b/flexible_event_config.md
@@ -296,6 +296,7 @@ The values are summed (to 8) and reported in the following reports after 7 days:
   "trigger_summary_bucket": [5, 9]
 }
 ```
+
 In the subsequent 7 days, the following triggers are registered:
 
 ```jsonc
@@ -428,7 +429,7 @@ We encourage developers to suggest different use cases they may have for this AP
 
 ## Attribution Rate Limit
 
-Attribution rate-limit behavior will be based on the numbers of reports generated rather than number of triggers that lead to report generation. For example: The 1st report in the [Reporting trigger value buckets](#reporting-trigger-value-buckets) example will be counted as 1 contribution towards the [max attribution per rate-limit window](https://wicg.github.io/attribution-reporting-api/#max-attributions-per-rate-limit-window) limit despite it being a result of contributions from 3 triggers.
+Attribution rate-limit behavior will be based on the number of reports generated rather than the number of triggers that lead to report generation. For example: The first report in the ["Reporting trigger value buckets"](#reporting-trigger-value-buckets) example will be counted as 1 contribution towards the [max attribution per rate-limit window](https://wicg.github.io/attribution-reporting-api/#max-attributions-per-rate-limit-window) limit despite it being the result of contributions from 3 triggers.
 
 ## Privacy considerations
 
--- a/flexible_event_config.md
+++ b/flexible_event_config.md
@@ -429,7 +429,13 @@ We encourage developers to suggest different use cases they may have for this AP
 
 ## Attribution Rate Limit
 
-Attribution rate-limit behavior will be based on the number of reports generated rather than the number of triggers that lead to report generation. For example: The first report in the ["Reporting trigger value buckets"](#reporting-trigger-value-buckets) example will be counted as 1 contribution towards the [max attribution per rate-limit window](https://wicg.github.io/attribution-reporting-api/#max-attributions-per-rate-limit-window) limit despite it being the result of contributions from 3 triggers.
+Attribution rate-limit behavior will be based on the number of reports generated
+rather than the number of triggers that lead to report generation. For example:
+The first report in the
+["Reporting trigger value buckets"](#reporting-trigger-value-buckets) example
+will be counted as 1 contribution towards the
+[max attribution per rate-limit window](https://wicg.github.io/attribution-reporting-api/#max-attributions-per-rate-limit-window)
+limit despite it being the result of contributions from 3 triggers.
 
 ## Privacy considerations
 
--- a/index.bs
+++ b/index.bs
@@ -3382,9 +3382,9 @@ To <dfn noexport>trigger attribution</dfn> given an [=attribution trigger=] |tri
 1. [=list/iterate|For each=] |item| of |matchingSources|:
     1. [=set/Remove=] |item| from the [=attribution source cache=].
 1. Let |eventLevelResult| be the result of running [=trigger event-level attribution=]
-    with |trigger|, |sourceToAttribute|.
+    with |trigger| and |sourceToAttribute|.
 1. Let |aggregatableResult| be the result of running [=trigger aggregatable attribution=]
-    with |trigger|, |sourceToAttribute|, and |rateLimitRecord|.
+    with |trigger| and |sourceToAttribute|.
 1. Let |eventLevelDebugData| be |eventLevelResult|'s [=triggering result/debug data=].
 1. Let |aggregatableDebugData| be |aggregatableResult|'s [=triggering result/debug data=].
 1. Let |debugDataList| be an [=list/is empty|empty=] [=list=].
--- a/index.bs
+++ b/index.bs
@@ -2884,10 +2884,9 @@ Given an [=attribution trigger=] |trigger|, [=attribution source=] |sourceToAttr
 Given an [=attribution rate-limit record=] |newRecord|:
 
 1. Let |max| be [=max source reporting origins per rate-limit window=].
-1. If |newRecord|'s [=attribution rate-limit record/scope=] is "<code>[=rate-limit scope/event-attribution=]</code>", set |max| to
-     [=max event attribution reporting origins per rate-limit window=].
-1. If |newRecord|'s [=attribution rate-limit record/scope=] is "<code>[=rate-limit scope/aggregatable-attribution=]</code>", set |max| to
-     [=max aggregatable attribution reporting origins per rate-limit window=].
+1. If |newRecord|'s [=attribution rate-limit record/scope=] is "<code>[=rate-limit scope/event-attribution=]</code>" 
+     or "<code>[=rate-limit scope/aggregatable-attribution=]</code>", set |max| to
+     [=max attribution reporting origins per rate-limit window=].
 1. Let |matchingRateLimitRecords| be all [=attribution rate-limit records=] |record| in the [=attribution rate-limit cache=] where all of the following are true:
      * |record|'s [=attribution rate-limit record/scope=] and |newRecord|'s [=attribution rate-limit record/scope=] are equal
      * |record|'s [=attribution rate-limit record/source site=] and |newRecord|'s [=attribution rate-limit record/source site=] are equal
@@ -3098,7 +3097,7 @@ To <dfn>maybe replace event-level report</dfn> given an [=attribution source=]
 1. Decrement |sourceToAttribute|'s [=attribution source/number of event-level reports=] value by 1.
 1. Let |rateLimitRecord| be the element from [=attribution rate-limit cache=] whose 
     [=attribution rate-limit record/entity id=] is equal to |report|'s [=event-level report/report id=]
-    and [=attribution rate-limit record/=] is equal to [=rate-limit scope/event-attribution=].
+    and [=attribution rate-limit record/scope=] is equal to [=rate-limit scope/event-attribution=].
 1. [=set/Remove=] |rateLimitRecord| from the [=attribution rate-limit cache=].
 1. Return "<code>[=event-level-report-replacement result/add-new-report=]</code>".
 
--- a/index.bs
+++ b/index.bs
@@ -1615,11 +1615,11 @@ This algorithm will return either a non-negative 128-bit integer or an error.
 Given an [=attribution rate-limit record=] |newRecord|:
 
 1. Let |max| be [=max source reporting origins per rate-limit window=].
-1. Let |scopeSet| be «"<code>[=rate-limit scope/source=]</code>"».
+1. Let |scopeSet| be « "<code>[=rate-limit scope/source=]</code>" ».
 1. If |newRecord|'s [=attribution rate-limit record/scope=] is "<code>[=rate-limit scope/event-attribution=]</code>" 
      or "<code>[=rate-limit scope/aggregatable-attribution=]</code>":
      1. Set |max| to [=max attribution reporting origins per rate-limit window=].
-     1. Set |scopeSet| to «"<code>[=rate-limit scope/event-attribution=]</code>", "<code>[=rate-limit scope/aggregatable-attribution=]</code>"».
+     1. Set |scopeSet| to « "<code>[=rate-limit scope/event-attribution=]</code>", "<code>[=rate-limit scope/aggregatable-attribution=]</code>" ».
 1. Let |matchingRateLimitRecords| be all [=attribution rate-limit records=] |record| in the [=attribution rate-limit cache=] where all of the following are true:
      * |scopeSet| [=set/contains=] |record|'s [=attribution rate-limit record/scope=]
      * |record|'s [=attribution rate-limit record/source site=] and |newRecord|'s [=attribution rate-limit record/source site=] are equal
--- a/index.bs
+++ b/index.bs
@@ -1037,8 +1037,8 @@ An <dfn>attribution rate-limit record</dfn> is a [=struct=] with the following i
 :: A [=moment=].
 : <dfn>expiry time</dfn>
 :: Null or a [=moment=].
-: <dfn>entity id</dfn>
-:: A [=event-level report=]'s [=event-level report/report id=] or [=aggregatable report=]'s [=aggregatable report/report id=] or [=attribution source/source identifier=].
+: <dfn>entity ID</dfn>
+:: A [=event-level report=]'s [=event-level report/report ID=] or [=aggregatable report=]'s [=aggregatable report/report ID=] or [=attribution source/source identifier=].
 
 </dl>
 
@@ -1070,7 +1070,7 @@ Possible values are:
 <li>"<dfn><code>trigger-aggregate-storage-limit</code></dfn>"
 <li>"<dfn><code>trigger-aggregate-report-window-passed</code></dfn>"
 <li>"<dfn><code>trigger-event-attributions-per-source-destination-limit</code></dfn>"
-<li>"<dfn><code>trigger-aggregatable-attributions-per-source-destination-limit</code></dfn>"
+<li>"<dfn><code>trigger-aggregate-attributions-per-source-destination-limit</code></dfn>"
 <li>"<dfn><code>trigger-event-deduplicated</code></dfn>"
 <li>"<dfn><code>trigger-event-excessive-reports</code></dfn>"
 <li>"<dfn><code>trigger-event-low-priority</code></dfn>"
@@ -1618,8 +1618,8 @@ This algorithm will return either a non-negative 128-bit integer or an error.
 
 Given an [=attribution rate-limit record=] |record| and a [=moment=] |now|:
 1. If the [=duration from=] |record|'s [=attribution rate-limit record|time=] and |now| is <= [=attribution rate-limit window=] , return false.
-1. If |record|'s [=attribution rate-limit record/scope=] is "<code>[=rate-limit scope/event-attribution=]</code>", return true.
-1. If |record|'s [=attribution rate-limit record/scope=] is "<code>[=rate-limit scope/aggregatable-attribution=]</code>", return true.
+1. If |record|'s [=attribution rate-limit record/scope=] is "<code>[=rate-limit scope/event-attribution=]</code>" 
+    or "<code>[=rate-limit scope/aggregatable-attribution=]</code>", return true.
 1. If |record|'s [=attribution rate-limit record/expiry time=] is after |now|, return false.
 1. Return true.
 
@@ -2542,7 +2542,7 @@ To <dfn>process an attribution source</dfn> given an [=attribution source=] |sou
             :: |source|'s [=attribution source/source time=]
             : [=attribution rate-limit record/expiry time=]
             :: null
-            : [=attribution rate-limit record/entity id=]
+            : [=attribution rate-limit record/entity ID=]
             :: null
         1. [=set/Append=] |rateLimitRecord| to the [=attribution rate-limit cache=].
     1. Set |debugDataType| to "<code>[=source debug data type/source-noised=]</code>".
@@ -2866,7 +2866,7 @@ To <dfn>match an attribution source against filters and negated filters</dfn> gi
 
 <h3 dfn id="should-block-attribution-for-attribution-limit">Should attribution be blocked by attribution rate limit</h3>
 
-Given an [=attribution trigger=] |trigger|, [=attribution source=] |sourceToAttribute| and [=attribution rate-limit record/scope=] |rateLimitScope| :
+Given an [=attribution trigger=] |trigger|, an [=attribution source=] |sourceToAttribute|, and an [=attribution rate-limit record/scope=] |rateLimitScope|:
 
 1. Let |matchingRateLimitRecords| be all [=attribution rate-limit records=] |record| of [=attribution rate-limit cache=] where all of the following are true:
      * |record|'s [=attribution rate-limit record/scope=] is |rateLimitScope|
@@ -2874,8 +2874,8 @@ Given an [=attribution trigger=] |trigger|, [=attribution source=] |sourceToAttr
      * |record|'s [=attribution rate-limit record/attribution destination=] and |trigger|'s [=attribution trigger/attribution destination=] are equal
      * |record|'s [=attribution rate-limit record/reporting origin=] and |trigger|'s [=attribution trigger/reporting origin=] are [=same site=]
      * |record|'s [=attribution rate-limit record/time=] is greater than [=attribution rate-limit window=] before |trigger|'s [=attribution trigger/trigger time=]
-1. Let |limit| be [=max aggregatable attributions per rate-limit window=]
-1. If |rateLimitScope| is [=rate-limit scope/event-attribution=], set |limit| = [=max event attributions per rate-limit window=]
+1. Let |limit| be [=max aggregatable attributions per rate-limit window=].
+1. If |rateLimitScope| is "<code>[=rate-limit scope/event-attribution=]</code>", set |limit| to [=max event attributions per rate-limit window=].
 1. If |matchingRateLimitRecords|'s [=list/size=] is greater than or equal to |limit|, return <strong>blocked</strong>.
 1. Return <strong>allowed</strong>.
 
@@ -2916,12 +2916,12 @@ Given an [=attribution trigger=] |trigger|, an [=attribution source=]
 1. If the result of running [=should attribution be blocked by attribution rate limit=] with |trigger|, |sourceToAttribute|
     and |scope| is <strong>blocked</strong>:
     1. Let |debugDataType| be "<code>[=trigger debug data type/trigger-event-attributions-per-source-destination-limit=]</code>"
-    2. If |scope| is "<code>[=rate-limit scope/aggregatable-attribution=]</code>", set |debugDataType| to 
-        "<code>[=trigger debug data type/trigger-aggregatable-attributions-per-source-destination-limit=]</code>"
-    3. Let |debugData| be the result of running [=obtain debug data on trigger registration=]
+    1. If |scope| is "<code>[=rate-limit scope/aggregatable-attribution=]</code>", set |debugDataType| to 
+        "<code>[=trigger debug data type/trigger-aggregate-attributions-per-source-destination-limit=]</code>"
+    1. Let |debugData| be the result of running [=obtain debug data on trigger registration=]
         with |debugDataType|, |trigger|, |sourceToAttribute| and
         [=obtain debug data on trigger registration/report=] set to null.
-    4. Return the [=triggering result=] ("<code>[=triggering status/dropped=]</code>", |debugData|).
+    1. Return the [=triggering result=] ("<code>[=triggering status/dropped=]</code>", |debugData|).
 1. If the result of running [=should processing be blocked by reporting-origin limit=] with
     |newRecord| is <strong>blocked</strong>:
     1. Let |debugData| be the result of running [=obtain debug data on trigger registration=]
@@ -2997,7 +2997,7 @@ an optional [=attribution source=] <dfn for="obtain debug data body on trigger r
     : "<code>[=trigger debug data type/trigger-event-attributions-per-source-destination-limit=]</code>"
     :: [=map/Set=] |body|["`limit`"] to the user agent's [=max event attributions per rate-limit window=],
          [=serialize an integer|serialized=].
-    : "<code>[=trigger debug data type/trigger-aggregatable-attributions-per-source-destination-limit=]</code>"
+    : "<code>[=trigger debug data type/trigger-aggregate-attributions-per-source-destination-limit=]</code>"
     :: [=map/Set=] |body|["`limit`"] to the user agent's [=max aggregatable attributions per rate-limit window=],
          [=serialize an integer|serialized=].
     : "<code>[=trigger debug data type/trigger-reporting-origin-limit=]</code>"
@@ -3096,7 +3096,7 @@ To <dfn>maybe replace event-level report</dfn> given an [=attribution source=]
 1. [=set/Remove=] |lowestPriorityReport| from the [=event-level report cache=].
 1. Decrement |sourceToAttribute|'s [=attribution source/number of event-level reports=] value by 1.
 1. Let |rateLimitRecord| be the element from [=attribution rate-limit cache=] whose 
-    [=attribution rate-limit record/entity id=] is equal to |report|'s [=event-level report/report id=]
+    [=attribution rate-limit record/entity ID=] is equal to |report|'s [=event-level report/report ID=]
     and [=attribution rate-limit record/scope=] is equal to [=rate-limit scope/event-attribution=].
 1. [=set/Remove=] |rateLimitRecord| from the [=attribution rate-limit cache=].
 1. Return "<code>[=event-level-report-replacement result/add-new-report=]</code>".
@@ -3105,7 +3105,7 @@ Issue: This algorithm is not compatible with the behavior proposed for
 [=experimental Flexible Event support=] with differing
 [=trigger spec/event-level report windows=] for a given source.
 
-To <dfn>trigger event-level attribution</dfn> given an [=attribution trigger=] |trigger|, and an
+To <dfn>trigger event-level attribution</dfn> given an [=attribution trigger=] |trigger| and an
 [=attribution source=] |sourceToAttribute|, run the following steps:
 
 1. If |trigger|'s [=attribution trigger/event-level trigger configurations=]
@@ -3184,8 +3184,8 @@ To <dfn>trigger event-level attribution</dfn> given an [=attribution trigger=] |
     :: |sourceToAttribute|'s [=attribution source/source time=]
     : [=attribution rate-limit record/expiry time=]
     :: null
-    : [=attribution rate-limit record/entity id=]
-    :: |report|'s [=event-level report/report id=]
+    : [=attribution rate-limit record/entity ID=]
+    :: |report|'s [=event-level report/report ID=]
 1. If the result of running [=should attribution be blocked by rate limits=]
     with |trigger|, |sourceToAttribute|, and |rateLimitRecord|'s [=attribution rate-limit record/scope=] is not null,
     return it.
@@ -3227,16 +3227,16 @@ To <dfn>trigger event-level attribution</dfn> given an [=attribution trigger=] |
 1. Increment |sourceToAttribute|'s [=attribution source/number of event-level reports=] value by 1.
 1. If |matchedConfig|'s [=event-level trigger configuration/dedup key=] is not null,
     [=list/append=] it to |sourceToAttribute|'s [=attribution source/dedup keys=].
+1. [=set/Append=] |rateLimitRecord| to the [=attribution rate-limit cache=].
 1. If |triggeringStatus| is "<code>[=triggering status/attributed=]</code>" and
     |report|'s [=event-level report/source debug key=] is not null and |report|'s
     [=event-level report/trigger debug key=] is not null, [=queue a task=] to
     [=attempt to deliver a debug report=] with |report|.
-1. [=set/Append=] |rateLimitRecord| to the [=attribution rate-limit cache=].
 1. Return the [=triggering result=] (|triggeringStatus|, |debugData|).
 
 <h3 algorithm id="triggering-aggregatable-attribution">Triggering aggregatable attribution</h3>
 
-To <dfn>trigger aggregatable attribution</dfn> given an [=attribution trigger=] |trigger|, and an
+To <dfn>trigger aggregatable attribution</dfn> given an [=attribution trigger=] |trigger| and an
 [=attribution source=] |sourceToAttribute|, run the following steps:
 
 1. If the result of running [=check if an attribution trigger contains aggregatable data=] is false,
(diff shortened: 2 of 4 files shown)


index.bs
```diff
@@ -1066,7 +1069,8 @@ Possible values are:
 <li>"<dfn><code>trigger-aggregate-insufficient-budget</code></dfn>"
 <li>"<dfn><code>trigger-aggregate-storage-limit</code></dfn>"
 <li>"<dfn><code>trigger-aggregate-report-window-passed</code></dfn>"
-<li>"<dfn><code>trigger-attributions-per-source-destination-limit</code></dfn>"
+<li>"<dfn><code>trigger-event-attributions-per-source-destination-limit</code></dfn>"
+<li>"<dfn><code>trigger-aggregatable-attributions-per-source-destination-limit</code></dfn>"
```
linnan-github, line 1073:
```suggestion
<li>"<dfn><code>trigger-aggregate-attributions-per-source-destination-limit</code></dfn>"
```

index.bs
```diff
@@ -1345,8 +1349,15 @@ positive integer that controls the maximum number of distinct
 [=attribution rate-limit record/attribution destination=]) that can create
 [=event-level reports=] per [=attribution rate-limit window=].
 
-<dfn>Max attributions per rate-limit window</dfn> is a positive integer that
-controls the maximum number of attributions for a
+<dfn>Max event attributions per rate-limit window</dfn> is a positive integer that
```
linnan-github, line 1352: Can we use a tuple, similar to "max destinations per rate-limit window"?

index.bs
```diff
@@ -1607,7 +1618,8 @@ This algorithm will return either a non-negative 128-bit integer or an error.
 
 Given an [=attribution rate-limit record=] |record| and a [=moment=] |now|:
 1. If the [=duration from=] |record|'s [=attribution rate-limit record|time=] and |now| is <= [=attribution rate-limit window=] , return false.
-1. If |record|'s [=attribution rate-limit record/scope=] is "<code>[=rate-limit scope/attribution=]</code>", return true.
+1. If |record|'s [=attribution rate-limit record/scope=] is "<code>[=rate-limit scope/event-attribution=]</code>", return true.
```
linnan-github, line 1621: Maybe use a switch?

index.bs
```diff
@@ -1036,6 +1037,8 @@ An <dfn>attribution rate-limit record</dfn> is a [=struct=] with the following i
 :: A [=moment=].
 : <dfn>expiry time</dfn>
 :: Null or a [=moment=].
+: <dfn>entity id</dfn>
```
apasel422, line 1040:
```suggestion
: <dfn>entity ID</dfn>
```

index.bs
```diff
@@ -1036,6 +1037,8 @@ An <dfn>attribution rate-limit record</dfn> is a [=struct=] with the following i
...
+:: A [=event-level report=]'s [=event-level report/report id=] or [=aggregatable report=]'s [=aggregatable report/report id=] or [=attribution source/source identifier=].
```
apasel422, line 1041:
```suggestion
:: A [=event-level report=]'s [=event-level report/report ID=] or [=aggregatable report=]'s [=aggregatable report/report ID=] or [=attribution source/source identifier=].
```

index.bs
```diff
@@ -2852,23 +2866,26 @@ To <dfn>match an attribution source against filters and negated filters</dfn> gi
...
 
 1. Let |matchingRateLimitRecords| be all [=attribution rate-limit records=] |record| of [=attribution rate-limit cache=] where all of the following are true:
-     * |record|'s [=attribution rate-limit record/scope=] is "<code>[=rate-limit scope/attribution=]</code>"
+     * |record|'s [=attribution rate-limit record/scope=] is |rateLimitScope|
      * |record|'s [=attribution rate-limit record/source site=] and |sourceToAttribute|'s [=attribution source/source site=] are equal
      * |record|'s [=attribution rate-limit record/attribution destination=] and |trigger|'s [=attribution trigger/attribution destination=] are equal
      * |record|'s [=attribution rate-limit record/reporting origin=] and |trigger|'s [=attribution trigger/reporting origin=] are [=same site=]
      * |record|'s [=attribution rate-limit record/time=] is greater than [=attribution rate-limit window=] before |trigger|'s [=attribution trigger/trigger time=]
-1. If |matchingRateLimitRecords|'s [=list/size=] is greater than or equal to [=max attributions per rate-limit window=], return <strong>blocked</strong>.
+1. Let |limit| be [=max aggregatable attributions per rate-limit window=]
```
apasel422, line 2877:
```suggestion
1. Let |limit| be [=max aggregatable attributions per rate-limit window=].
```

index.bs
```diff
@@ -2852,23 +2866,26 @@ To <dfn>match an attribution source against filters and negated filters</dfn> gi
...
+1. If |rateLimitScope| is [=rate-limit scope/event-attribution=], set |limit| = [=max event attributions per rate-limit window=]
```
apasel422, line 2878:
```suggestion
1. If |rateLimitScope| is "<code>[=rate-limit scope/event-attribution=]</code>", set |limit| to [=max event attributions per rate-limit window=].
```

index.bs
```diff
@@ -2894,14 +2911,17 @@ Given an [=attribution rate-limit record=] |newRecord|:
...
+1. If the result of running [=should attribution be blocked by attribution rate limit=] with |trigger|, |sourceToAttribute|
+    and |scope| is <strong>blocked</strong>:
+    1. Let |debugDataType| be "<code>[=trigger debug data type/trigger-event-attributions-per-source-destination-limit=]</code>"
+    2. If |scope| is "<code>[=rate-limit scope/aggregatable-attribution=]</code>", set |debugDataType| to 
+        "<code>[=trigger debug data type/trigger-aggregatable-attributions-per-source-destination-limit=]</code>"
+    3. Let |debugData| be the result of running [=obtain debug data on trigger registration=]
+        with |debugDataType|, |trigger|, |sourceToAttribute| and
         [=obtain debug data on trigger registration/report=] set to null.
-    1. Return the [=triggering result=] ("<code>[=triggering status/dropped=]</code>", |debugData|).
+    4. Return the [=triggering result=] ("<code>[=triggering status/dropped=]</code>", |debugData|).
```
apasel422, line 2924: We use `1.` for numbering each step to avoid having to renumber when steps are added or deleted.

index.bs
```diff
@@ -3072,15 +3095,18 @@ To <dfn>maybe replace event-level report</dfn> given an [=attribution source=]
     return "<code>[=event-level-report-replacement result/drop-new-report-low-priority=]</code>".
 1. [=set/Remove=] |lowestPriorityReport| from the [=event-level report cache=].
 1. Decrement |sourceToAttribute|'s [=attribution source/number of event-level reports=] value by 1.
+1. Let |rateLimitRecord| be the element from [=attribution rate-limit cache=] whose 
+    [=attribution rate-limit record/entity id=] is equal to |report|'s [=event-level report/report id=]
```
apasel422, line 3099:
```suggestion
    [=attribution rate-limit record/entity ID=] is equal to |report|'s [=event-level report/report ID=]
```

index.bs
```diff
@@ -3072,15 +3095,18 @@ To <dfn>maybe replace event-level report</dfn> given an [=attribution source=]
     return "<code>[=event-level-report-replacement result/drop-new-report-low-priority=]</code>".
 1. [=set/Remove=] |lowestPriorityReport| from the [=event-level report cache=].
 1. Decrement |sourceToAttribute|'s [=attribution source/number of event-level reports=] value by 1.
+1. Let |rateLimitRecord| be the element from [=attribution rate-limit cache=] whose
```
apasel422, line 3098: Does this imply that there can only be one such element?

index.bs
```diff
@@ -3142,12 +3168,27 @@ To <dfn>trigger event-level attribution</dfn> given an [=attribution trigger=] |
...
+    : [=attribution rate-limit record/attribution destination=]
+    :: |trigger|'s [=attribution trigger/attribution destination=]
+    : [=attribution rate-limit record/reporting origin=]
+    :: |sourceToAttribute|'s [=attribution source/reporting origin=]
+    : [=attribution rate-limit record/time=]
+    :: |sourceToAttribute|'s [=attribution source/source time=]
+    : [=attribution rate-limit record/expiry time=]
+    :: null
+    : [=attribution rate-limit record/entity id=]
+    :: |report|'s [=event-level report/report id=]
```
apasel422, line 3188:
```suggestion
    : [=attribution rate-limit record/entity ID=]
    :: |report|'s [=event-level report/report ID=]
```

index.bs
```diff
@@ -3072,15 +3095,18 @@ To <dfn>maybe replace event-level report</dfn> given an [=attribution source=]
...
 1. Return "<code>[=event-level-report-replacement result/add-new-report=]</code>".
 
 Issue: This algorithm is not compatible with the behavior proposed for
 [=experimental Flexible Event support=] with differing
 [=trigger spec/event-level report windows=] for a given source.
 
-To <dfn>trigger event-level attribution</dfn> given an [=attribution trigger=] |trigger|, an
-[=attribution source=] |sourceToAttribute|, and an
-[=attribution rate-limit record=] |rateLimitRecord|, run the following steps:
+To <dfn>trigger event-level attribution</dfn> given an [=attribution trigger=] |trigger|, and an
```
apasel422, line 3108:
```suggestion
To <dfn>trigger event-level attribution</dfn> given an [=attribution trigger=] |trigger| and an
```

index.bs
```diff
@@ -3190,13 +3231,13 @@ To <dfn>trigger event-level attribution</dfn> given an [=attribution trigger=] |
...
     [=attempt to deliver a debug report=] with |report|.
+1. [=set/Append=] |rateLimitRecord| to the [=attribution rate-limit cache=].
 1. Return the [=triggering result=] (|triggeringStatus|, |debugData|).
 
 <h3 algorithm id="triggering-aggregatable-attribution">Triggering aggregatable attribution</h3>
 
-To <dfn>trigger aggregatable attribution</dfn> given an [=attribution trigger=] |trigger|, an
-[=attribution source=] |sourceToAttribute|, and an
-[=attribution rate-limit record=] |rateLimitRecord|, run the following steps:
+To <dfn>trigger aggregatable attribution</dfn> given an [=attribution trigger=] |trigger|, and an
```
apasel422, line 3239:
```suggestion
To <dfn>trigger aggregatable attribution</dfn> given an [=attribution trigger=] |trigger| and an
```

index.bs
```diff
@@ -3237,8 +3278,23 @@ To <dfn>trigger aggregatable attribution</dfn> given an [=attribution trigger=]
...
+    : [=attribution rate-limit record/attribution destination=]
+    :: |trigger|'s [=attribution trigger/attribution destination=]
+    : [=attribution rate-limit record/reporting origin=]
+    :: |sourceToAttribute|'s [=attribution source/reporting origin=]
+    : [=attribution rate-limit record/time=]
+    :: |sourceToAttribute|'s [=attribution source/source time=]
+    : [=attribution rate-limit record/expiry time=]
+    :: null
+    : [=attribution rate-limit record/entity id=]
+    :: |report|'s [=aggregatable report/report id=]
```
apasel422, line 3295:
```suggestion
    : [=attribution rate-limit record/entity ID=]
    :: |report|'s [=aggregatable report/report ID=]
```

index.bs
```diff
@@ -2852,23 +2866,26 @@ To <dfn>match an attribution source against filters and negated filters</dfn> gi
 
 <h3 dfn id="should-block-attribution-for-attribution-limit">Should attribution be blocked by attribution rate limit</h3>
 
-Given an [=attribution trigger=] |trigger| and [=attribution source=] |sourceToAttribute|:
+Given an [=attribution trigger=] |trigger|, [=attribution source=] |sourceToAttribute| and [=attribution rate-limit record/scope=] |rateLimitScope| :
```
apasel422, line 2869:
```suggestion
Given an [=attribution trigger=] |trigger|, [=attribution source=] |sourceToAttribute|, and [=attribution rate-limit record/scope=] |rateLimitScope|:
```

apasel422 (changes requested):
Is the "if we start generating more than 1 report per type per trigger" in the PR description referring to Full Flex? If so, maybe we can say that explicitly.

index.bs
```diff
@@ -3190,13 +3231,13 @@ To <dfn>trigger event-level attribution</dfn> given an [=attribution trigger=] |
     |report|'s [=event-level report/source debug key=] is not null and |report|'s
     [=event-level report/trigger debug key=] is not null, [=queue a task=] to
     [=attempt to deliver a debug report=] with |report|.
+1. [=set/Append=] |rateLimitRecord| to the [=attribution rate-limit cache=].
```
linnan-github, line 3234: Let's move the step before queueing the report, similarly for aggregtable reports.

apasel422:
@vikassahu29 Could you provide some more context for this in the PR description? Is there an issue you can link to that describes what's wrong with the existing limits, if anything?

index.bs
vikassahu29: Yes, correct.

index.bs
vikassahu29: I replace it with one if statement because it was anyway doing the same thing for both branches.

index.bs
vikassahu29: I think it is more readable in current format compared to the tuple impl (similar to max destinations per rate-limit window).
What do you think?

index.bs
```diff
@@ -1036,6 +1037,8 @@ An <dfn>attribution rate-limit record</dfn> is a [=struct=] with the following i
 :: A [=moment=].
 : <dfn>expiry time</dfn>
 :: Null or a [=moment=].
+: <dfn>entity ID</dfn>
+:: A [=event-level report=]'s [=event-level report/report ID=] or [=aggregatable report=]'s [=aggregatable report/report ID=] or [=attribution source/source identifier=].
```
linnan-github, line 1041:
```suggestion
:: An [=event-level report=]'s [=event-level report/report ID=] or an [=aggregatable report=]'s [=aggregatable report/report ID=] or an [=attribution source=]'s [=attribution source/source identifier=].
```

index.bs
linnan-github: Alternatively, maybe we can still use one constant for both? Unless there's a reason to use different limits?

linnan-github: If we use separate variables, please update https://github.com/WICG/attribution-reporting-api/blob/main/params/chromium-params.md as well.

index.bs
```diff
@@ -2894,12 +2911,15 @@ Given an [=attribution rate-limit record=] |newRecord|:
 <h3 dfn id="should-block-attribution-for-rate-limits">Should attribution be blocked by rate limits</h3>
 
 Given an [=attribution trigger=] |trigger|, an [=attribution source=]
-|sourceToAttribute|, and an [=attribution rate-limit record=] |newRecord|:
+|sourceToAttribute|, and an [=attribution rate-limit record/scope=] |scope|:
```
linnan-github, line 2914: |newRecord| is still used in this algorithm.

index.bs
linnan-github: Is it guaranteed that such an element always exists?

index.bs
```diff
@@ -1066,7 +1071,8 @@ Possible values are:
 <li>"<dfn><code>trigger-aggregate-insufficient-budget</code></dfn>"
 <li>"<dfn><code>trigger-aggregate-storage-limit</code></dfn>"
 <li>"<dfn><code>trigger-aggregate-report-window-passed</code></dfn>"
-<li>"<dfn><code>trigger-attributions-per-source-destination-limit</code></dfn>"
+<li>"<dfn><code>trigger-event-attributions-per-source-destination-limit</code></dfn>"
+<li>"<dfn><code>trigger-aggregate-attributions-per-source-destination-limit</code></dfn>"
```
apasel422, line 1075: Please also update https://github.com/WICG/attribution-reporting-api/blob/main/verbose_debugging_reports.md with the deleted/added verbose debug report types.

index.bs
```diff
@@ -2896,10 +2915,13 @@ Given an [=attribution rate-limit record=] |newRecord|:
 Given an [=attribution trigger=] |trigger|, an [=attribution source=]
 |sourceToAttribute|, and an [=attribution rate-limit record=] |newRecord|:
 
-1. If the result of running [=should attribution be blocked by attribution rate limit=] with |trigger| and
-    |sourceToAttribute| is <strong>blocked</strong>:
+1. If the result of running [=should attribution be blocked by attribution rate limit=] with |trigger|, |sourceToAttribute|
```
apasel422, line 2918:
```suggestion
1. If the result of running [=should attribution be blocked by attribution rate limit=] with |trigger|, |sourceToAttribute|,
```

index.bs
```diff
@@ -2896,10 +2915,13 @@ Given an [=attribution rate-limit record=] |newRecord|:
...
+    and |newRecord|'s [=attribution rate-limit record/scope=] is <strong>blocked</strong>:
+    1. Let |debugDataType| be "<code>[=trigger debug data type/trigger-event-attributions-per-source-destination-limit=]</code>"
+    1. If |newRecord|'s [=attribution rate-limit record/scope=] is "<code>[=rate-limit scope/aggregatable-attribution=]</code>",
+        set |debugDataType| to "<code>[=trigger debug data type/trigger-aggregate-attributions-per-source-destination-limit=]</code>"
     1. Let |debugData| be the result of running [=obtain debug data on trigger registration=]
-        with "<code>[=trigger debug data type/trigger-attributions-per-source-destination-limit=]</code>", |trigger|, |sourceToAttribute| and
+        with |debugDataType|, |trigger|, |sourceToAttribute| and
```
apasel422, line 2924:
```suggestion
        with |debugDataType|, |trigger|, |sourceToAttribute|, and
```

index.bs
```diff
@@ -3072,15 +3097,18 @@ To <dfn>maybe replace event-level report</dfn> given an [=attribution source=]
     return "<code>[=event-level-report-replacement result/drop-new-report-low-priority=]</code>".
 1. [=set/Remove=] |lowestPriorityReport| from the [=event-level report cache=].
 1. Decrement |sourceToAttribute|'s [=attribution source/number of event-level reports=] value by 1.
+1. Let |rateLimitRecord| be the element from [=attribution rate-limit cache=] whose 
+    [=attribution rate-limit record/entity ID=] is equal to |report|'s [=event-level report/report ID=]
+    and [=attribution rate-limit record/scope=] is equal to [=rate-limit scope/event-attribution=].
```
apasel422, line 3102:
```suggestion
    and [=attribution rate-limit record/scope=] is equal to "<code>[=rate-limit scope/event-attribution=]</code>".
```

apasel422 (changes requested):
The PR description mentions "accurate counting for [Full Flex proposal](https://github.com/WICG/attribution-reporting-api/blob/main/flexible_event_config.md)", but that proposal does not discuss rate-limiting at all, so it's unclear to me whether this PR is compatible with Full Flex. Would it be possible for you to update that proposal in this PR as well?

verbose_debugging_reports.md
```diff
@@ -43,8 +43,11 @@ A trigger is rejected due to no matching sources in storage that match <reportin
 #### `trigger-no-matching-filter-data`
 A trigger is rejected due to no [matching filter data](https://github.com/WICG/attribution-reporting-api/blob/main/EVENT.md#optional-attribution-filters).
 
-#### `trigger-attributions-per-source-destination-limit`
-A trigger is rejected due to the [max attributions rate limit](https://github.com/WICG/attribution-reporting-api/blob/main/EVENT.md#reporting-cooldown--rate-limits).
+#### `trigger-event-attributions-per-source-destination-limit`
+An event-level report is rejected due to the [max attributions rate limit](https://github.com/WICG/attribution-reporting-api/blob/main/EVENT.md#reporting-cooldown--rate-limits).
+
```
apasel422, line 48: The table at the bottom of the file also needs to be updated.

verbose_debugging_reports.md
```diff
@@ -43,8 +43,11 @@ A trigger is rejected due to no matching sources in storage that match <reportin
 #### `trigger-no-matching-filter-data`
 A trigger is rejected due to no [matching filter data](https://github.com/WICG/attribution-reporting-api/blob/main/EVENT.md#optional-attribution-filters).
 
-#### `trigger-attributions-per-source-destination-limit`
-A trigger is rejected due to the [max attributions rate limit](https://github.com/WICG/attribution-reporting-api/blob/main/EVENT.md#reporting-cooldown--rate-limits).
+#### `trigger-event-attributions-per-source-destination-limit`
```
linnan-github, line 46: Please update the corresponding errors in the table below.

apasel422 (changes requested):

index.bs
```diff
@@ -1036,6 +1037,10 @@ An <dfn>attribution rate-limit record</dfn> is a [=struct=] with the following i
 :: A [=moment=].
 : <dfn>expiry time</dfn>
 :: Null or a [=moment=].
+: <dfn>entity ID</dfn>
+:: An [=event-level report=]'s [=event-level report/report ID=] or an
+    [=aggregatable report=]'s [=aggregatable report/report ID=] or an
+    [=attribution source=]'s [=attribution source/source identifier=].
```
linnan-github, line 1043: Where is this set for the source?

flexible_event_config.md
```diff
@@ -427,6 +427,10 @@ is dropped.
 
 We encourage developers to suggest different use cases they may have for this API extension, and we will update this explainer with sample configurations for those use cases.
 
+## Attribution Rate Limit
+
+Attribution rate-limit behavior will be based on the numbers of reports generated rather than number of triggers that lead to report generation. For example: The 1st report in the [Reporting trigger value buckets](#reporting-trigger-value-buckets) example will be counted as 1 contribution towards the [max attribution per rate-limit window](https://wicg.github.io/attribution-reporting-api/#max-attributions-per-rate-limit-window) limit despite it being a result of contributions from 3 triggers.
```
apasel422, line 432: If Full Flex is accepted, would it make more sense to rename this limit to reflect that it is report-based, rather than attribution-based? The current naming is confusing.

flexible_event_config.md
```diff
@@ -296,7 +296,7 @@ The values are summed (to 8) and reported in the following reports after 7 days:
   "trigger_summary_bucket": [5, 9]
 }
 ```
-
+Instead of these triggers being counted as 3 attributions from rate-limit perspective, they will be only counted as 1 attribution because 1
```
apasel422, line 299: This sentence is incomplete.

apasel422 (changes requested):

flexible_event_config.md
vikassahu29: Removed the line.

index.bs
vikassahu29: I have added the set line for this, but it is not used in querying anywhere yet.

flexible_event_config.md
vikassahu29: Yes, I think so. Maybe something on the lines of max reported-attributions, max reported-conversions or just max reports.

giladbarkan-github: I'd support renaming it already.

linnan-github: One caveat with "max reports" is the fake event-level reports, for which we only count 1 regardless of how many fake reports it creates.

flexible_event_config.md
```diff
@@ -296,7 +296,6 @@ The values are summed (to 8) and reported in the following reports after 7 days:
   "trigger_summary_bucket": [5, 9]
 }
 ```
-
```
apasel422, line 299: Please leave this blank line.

flexible_event_config.md
```diff
@@ -427,6 +426,10 @@ is dropped.
 
 We encourage developers to suggest different use cases they may have for this API extension, and we will update this explainer with sample configurations for those use cases.
 
+## Attribution Rate Limit
+
+Attribution rate-limit behavior will be based on the numbers of reports generated rather than number of triggers that lead to report generation. For example: The 1st report in the [Reporting trigger value buckets](#reporting-trigger-value-buckets) example will be counted as 1 contribution towards the [max attribution per rate-limit window](https://wicg.github.io/attribution-reporting-api/#max-attributions-per-rate-limit-window) limit despite it being a result of contributions from 3 triggers.
```
apasel422, line 431:
```suggestion
Attribution rate-limit behavior will be based on the number of reports generated rather than the number of triggers that lead to report generation. For example: The first report in the ["Reporting trigger value buckets"](#reporting-trigger-value-buckets) example will be counted as 1 contribution towards the [max attribution per rate-limit window](https://wicg.github.io/attribution-reporting-api/#max-attributions-per-rate-limit-window) limit despite it being the result of contributions from 3 triggers.
```

verbose_debugging_reports.md
```diff
@@ -43,8 +43,11 @@ A trigger is rejected due to no matching sources in storage that match <reportin
...
+An event-level report is rejected due to the [max attributions rate limit](https://github.com/WICG/attribution-reporting-api/blob/main/EVENT.md#reporting-cooldown--rate-limits).
```
apasel422, line 47:
```suggestion
An event-level attribution is rejected due to the [max attributions rate limit](https://github.com/WICG/attribution-reporting-api/blob/main/EVENT.md#reporting-cooldown--rate-limits).
```

verbose_debugging_reports.md
```diff
@@ -43,8 +43,11 @@ A trigger is rejected due to no matching sources in storage that match <reportin
...
+
+#### `trigger-aggregate-attributions-per-source-destination-limit`
+An aggregatable report is rejected due to the [max attributions rate limit](https://github.com/WICG/attribution-reporting-api/blob/main/EVENT.md#reporting-cooldown--rate-limits).
```
apasel422, line 50:
```suggestion
An aggregatable attribution is rejected due to the [max attributions rate limit](https://github.com/WICG/attribution-reporting-api/blob/main/EVENT.md#reporting-cooldown--rate-limits).
```

apasel422 (changes requested):

linnan-github (commented):
Could you please also update the website explainer rate-limits section to reflect this change?

index.bs
```diff
@@ -2881,7 +2893,7 @@ Given an [=attribution rate-limit record=] |newRecord|:
 
     NOTE: [=rate-limit scope/source=] scopes have an auxiliary [=max source reporting origins per source reporting site=] rate limit that also must be enforced.
 
-1. If |newRecord|'s [=attribution rate-limit record/scope=] is "<code>[=rate-limit scope/attribution=]</code>", return <strong>allowed</strong>
+1. If |newRecord|'s [=attribution rate-limit record/scope=] is "<code>[=rate-limit scope/event-attribution=]</code>" or "<code>[=rate-limit scope/aggregatable-attribution=]</code>", return <strong>allowed</strong>
```
apasel422, line 2896:
```suggestion
1. If |newRecord|'s [=attribution rate-limit record/scope=] is "<code>[=rate-limit scope/event-attribution=]</code>" or "<code>[=rate-limit scope/aggregatable-attribution=]</code>", return <strong>allowed</strong>.
```

apasel422 (approved):

index.bs
```diff
@@ -1036,6 +1037,10 @@ An <dfn>attribution rate-limit record</dfn> is a [=struct=] with the following i
 :: A [=moment=].
 : <dfn>expiry time</dfn>
 :: Null or a [=moment=].
+: <dfn>entity ID</dfn>
+:: An [=event-level report=]'s [=event-level report/report ID=] or an
+    [=aggregatable report=]'s [=aggregatable report/report ID=] or an
```
apasel422, line 1042: What is the purpose in storing the aggregatable report ID or source ID? IIUC the ID here is only needed so that rows corresponding event-level attributions can be replaced.

index.bs
```diff
@@ -1036,6 +1037,10 @@ An <dfn>attribution rate-limit record</dfn> is a [=struct=] with the following i
 :: A [=moment=].
 : <dfn>expiry time</dfn>
 :: Null or a [=moment=].
+: <dfn>entity ID</dfn>
```
apasel422, line 1040: How do we expect this field to work in the context of Full Flex? There may not be a persistent identifier corresponding to a report at the time of attribution, since reports under that proposal are effectively computed at the end of each report window by a cumulative operation across all triggers received during the window.

I assume this whole notion of rate-limit-row replacement logic is going to have to be itself replaced as part of Full Flex?

linnan-github (approved):

index.bs
vikassahu29: I don't thing this whole thing will need to replaced.

Whenever report re-computation happens, we can delete the obsolete rate-limit records and add new ones.

apasel422: I'm not sure the spec is going to be written using report re-computation, as that's more of an implementation detail, but we can deal with that when Full Flex is specified.

index.bs
vikassahu29: Source ID will be used for some upcoming features where we want to modify some rate-limit data.

Aggregate Report id is not used right but might have utility if we have deletion cases for aggregate report.

flexible_event_config.md
```diff
@@ -427,6 +427,10 @@ is dropped.
 
 We encourage developers to suggest different use cases they may have for this API extension, and we will update this explainer with sample configurations for those use cases.
 
+## Attribution Rate Limit
+
+Attribution rate-limit behavior will be based on the number of reports generated rather than the number of triggers that lead to report generation. For example: The first report in the ["Reporting trigger value buckets"](#reporting-trigger-value-buckets) example will be counted as 1 contribution towards the [max attribution per rate-limit window](https://wicg.github.io/attribution-reporting-api/#max-attributions-per-rate-limit-window) limit despite it being the result of contributions from 3 triggers.
```
giladbarkan-github, line 432: In this PR state, https://pr-preview.s3.amazonaws.com/vikassahu29/attribution-reporting-api/pull/1211.html#max-attributions-per-rate-limit-window doesn't seem to have language about applying to different scopes. Should we add some?

index.bs
giladbarkan-github: I'd leave unused fields out unless there's a compelling reason to include them.

index.bs
```diff
@@ -2487,6 +2494,8 @@ To <dfn>process an attribution source</dfn> given an [=attribution source=] |sou
         :: |source|'s [=attribution source/source time=]
         : [=attribution rate-limit record/expiry time=]
         :: |source|'s [=attribution source/expiry time=]
+        : [=attribution rate-limit record/entity ID=]
+        :: |source|'s [=attribution source/source identifier=]
```
giladbarkan-github, line 2498: If the spec isn't using this, I'd keep it null.

index.bs
```diff
@@ -2841,10 +2852,10 @@ To <dfn>match an attribution source against filters and negated filters</dfn> gi
 
 <h3 dfn id="should-block-attribution-for-attribution-limit">Should attribution be blocked by attribution rate limit</h3>
 
-Given an [=attribution trigger=] |trigger| and [=attribution source=] |sourceToAttribute|:
+Given an [=attribution trigger=] |trigger|, an [=attribution source=] |sourceToAttribute|, and an [=attribution rate-limit record/scope=] |rateLimitScope|:
```
giladbarkan-github, line 2855: "a scope" , not "an scope"

index.bs
apasel422: +1 to @giladbarkan-github. The spec can be modified in the future if there's a reason to store other IDs here. We don't need to add support for speculative extension, unless you're planning to post a specific proposal in the near future, in which case, maybe add a non-normative `Note` here explaining that this will be used for a particular purpose.

index.bs
apasel422: We should assert that the element exists if it's relevant. If there's a possibility of the element not existing, we should say what to do here.

apasel422 (changes requested):

index.bs
```diff
@@ -2857,7 +2868,8 @@ Given an [=attribution trigger=] |trigger| and [=attribution source=] |sourceToA
 Given an [=attribution rate-limit record=] |newRecord|:
 
 1. Let |max| be [=max source reporting origins per rate-limit window=].
-1. If |newRecord|'s [=attribution rate-limit record/scope=] is "<code>[=rate-limit scope/attribution=]</code>", set |max| to
+1. If |newRecord|'s [=attribution rate-limit record/scope=] is "<code>[=rate-limit scope/event-attribution=]</code>"
```
giladbarkan-github, line 2871: This is a limit on distinct reporting origins. What is the reasoning for splitting this into event- and aggregate-levels?

index.bs
```diff
@@ -3061,15 +3077,18 @@ To <dfn>maybe replace event-level report</dfn> given an [=attribution source=]
     return "<code>[=event-level-report-replacement result/drop-new-report-low-priority=]</code>".
 1. [=set/Remove=] |lowestPriorityReport| from the [=event-level report cache=].
 1. Decrement |sourceToAttribute|'s [=attribution source/number of event-level reports=] value by 1.
+1. Let |rateLimitRecord| be the element from [=attribution rate-limit cache=] whose 
+    [=attribution rate-limit record/entity ID=] is equal to |report|'s [=event-level report/report ID=]
```
giladbarkan-github, line 3081: Shouldn't the entity ID be equal to `lowestPriorityReport`'s record ID?

giladbarkan-github: And wouldn't the scope check be redundant?

index.bs
```diff
@@ -3061,15 +3077,18 @@ To <dfn>maybe replace event-level report</dfn> given an [=attribution source=]
...
+    and [=attribution rate-limit record/scope=] is equal to "<code>[=rate-limit scope/event-attribution=]</code>".
```
giladbarkan-github, line 3082: And wouldn't the scope check be redundant?

index.bs
giladbarkan-github: Spec has a section where attribution rate-limit records are regularly removed. Could this be making an assumption that if we get here, it wasn't yet removed (which we haven't tracked the dependencies of)?

index.bs
```diff
@@ -3175,6 +3208,7 @@ To <dfn>trigger event-level attribution</dfn> given an [=attribution trigger=] |
 1. Increment |sourceToAttribute|'s [=attribution source/number of event-level reports=] value by 1.
 1. If |matchedConfig|'s [=event-level trigger configuration/dedup key=] is not null,
     [=list/append=] it to |sourceToAttribute|'s [=attribution source/dedup keys=].
+1. [=set/Append=] |rateLimitRecord| to the [=attribution rate-limit cache=].
```
giladbarkan-github, line 3211: I don't think we are supposed to add an attribution rate-limit record if `sourceToAttribute`'s randomized response is not `null`.

index.bs
```diff
@@ -3226,6 +3259,21 @@ To <dfn>trigger aggregatable attribution</dfn> given an [=attribution trigger=]
...
+    : [=attribution rate-limit record/attribution destination=]
+    :: |trigger|'s [=attribution trigger/attribution destination=]
+    : [=attribution rate-limit record/reporting origin=]
+    :: |sourceToAttribute|'s [=attribution source/reporting origin=]
+    : [=attribution rate-limit record/time=]
+    :: |sourceToAttribute|'s [=attribution source/source time=]
+    : [=attribution rate-limit record/expiry time=]
+    :: null
+    : [=attribution rate-limit record/entity ID=]
+    :: |report|'s [=aggregatable report/report ID=]
```
giladbarkan-github, line 3276: If the spec isn't using this, I'd keep it null.

index.bs
linnan-github: The IDs for sources and aggregatable reports will be used in https://github.com/WICG/attribution-reporting-api/pull/1217, but feel free to remove this from this PR and I can update there.

index.bs
```diff
@@ -2857,7 +2868,8 @@ Given an [=attribution trigger=] |trigger| and [=attribution source=] |sourceToA
 Given an [=attribution rate-limit record=] |newRecord|:
 
 1. Let |max| be [=max source reporting origins per rate-limit window=].
-1. If |newRecord|'s [=attribution rate-limit record/scope=] is "<code>[=rate-limit scope/attribution=]</code>", set |max| to
+1. If |newRecord|'s [=attribution rate-limit record/scope=] is "<code>[=rate-limit scope/event-attribution=]</code>" 
+     or "<code>[=rate-limit scope/aggregatable-attribution=]</code>", set |max| to
      [=max attribution reporting origins per rate-limit window=].
 1. Let |matchingRateLimitRecords| be all [=attribution rate-limit records=] |record| in the [=attribution rate-limit cache=] where all of the following are true:
      * |record|'s [=attribution rate-limit record/scope=] and |newRecord|'s [=attribution rate-limit record/scope=] are equal
```
linnan-github, line 2875: This should be updated to allow both event-attribution and aggregatable-attribution to be considered.

index.bs
linnan-github: Good catch, please fix,  thanks.

index.bs
linnan-github: Attribution rate-limit records are deleted at source time + rate limit window. Since rate-limit window is no shorter than the expiry duration, I think we can assert that such an element always exists IIUC.

linnan-github (changes requested):

index.bs
giladbarkan-github: Doesn't that make an assumption about the relationship between expiry duration and rate limit window?

giladbarkan-github: That might be hard to catch if that relationship changes, unless there's a definition for that relationship.

linnan-github: Both are constants, we may add a note or assert this relation.

https://wicg.github.io/attribution-reporting-api/#valid-source-expiry-range
https://wicg.github.io/attribution-reporting-api/#attribution-rate-limit-window

index.bs
giladbarkan-github: To clarify -- we do insert one but elsewhere in the spec.

index.bs
```diff
@@ -2857,10 +2866,13 @@ Given an [=attribution trigger=] |trigger| and [=attribution source=] |sourceToA
 Given an [=attribution rate-limit record=] |newRecord|:
 
 1. Let |max| be [=max source reporting origins per rate-limit window=].
-1. If |newRecord|'s [=attribution rate-limit record/scope=] is "<code>[=rate-limit scope/attribution=]</code>", set |max| to
-     [=max attribution reporting origins per rate-limit window=].
+1. Let |scopeSet| be a [=set=] of [=rate-limit scope=] containing "<code>[=rate-limit scope/source=]</code>"
```
giladbarkan-github, line 2869: rate-limit scopes [plural] (e.g., we wouldn't say "a set of apple").

index.bs
```diff
@@ -2857,10 +2866,13 @@ Given an [=attribution trigger=] |trigger| and [=attribution source=] |sourceToA
...
+1. If |newRecord|'s [=attribution rate-limit record/scope=] is "<code>[=rate-limit scope/event-attribution=]</code>" 
+     or "<code>[=rate-limit scope/aggregatable-attribution=]</code>":
+     1. Set |max| to [=max attribution reporting origins per rate-limit window=].
+     1. Set |scopeSet| be a [=set=] containing "<code>[=rate-limit scope/event-attribution=]</code>" and "<code>[=rate-limit scope/aggregatable-attribution=]</code>".
```
giladbarkan-github, line 2873: "set scopeSet to a set..." rather than "set scopeSet be..."

index.bs
giladbarkan-github: Re-upping

giladbarkan-github (approved):
LGTM overall. A couple of nitpicks.

linnan-github (changes requested):

index.bs
vikassahu29: I don't think there is any harm in keeping this check.

linnan-github (approved):
Please also wait for Andrew to approve, thanks.

apasel422 (commented):
Holding off on my review until the open comments are addressed.

flexible_event_config.md
vikassahu29: Does `max reported-attributions per rate limit window` sound good to everyone? @linnan-github , @apasel422 & @giladbarkan-github

giladbarkan-github: Fine with me. Renaming it all here could be a big change, though. I assume we'd do it after?

vikassahu29: Ack. Doing it in a follow-up sounds good to me.

apasel422 (changes requested):

index.bs
vikassahu29: Ack, we can revisit this when we implement Full Flex. Also, this field will be useful for #1217 regardless of Flex implementation.

index.bs
```diff
@@ -1611,10 +1615,13 @@ This algorithm will return either a non-negative 128-bit integer or an error.
 Given an [=attribution rate-limit record=] |newRecord|:
 
 1. Let |max| be [=max source reporting origins per rate-limit window=].
-1. If |newRecord|'s [=attribution rate-limit record/scope=] is "<code>[=rate-limit scope/attribution=]</code>", set |max| to
-     [=max attribution reporting origins per rate-limit window=].
+1. Let |scopeSet| be «"<code>[=rate-limit scope/source=]</code>"».
+1. If |newRecord|'s [=attribution rate-limit record/scope=] is "<code>[=rate-limit scope/event-attribution=]</code>" 
+     or "<code>[=rate-limit scope/aggregatable-attribution=]</code>":
+     1. Set |max| to [=max attribution reporting origins per rate-limit window=].
+     1. Set |scopeSet| to «"<code>[=rate-limit scope/event-attribution=]</code>", "<code>[=rate-limit scope/aggregatable-attribution=]</code>"».
```
apasel422, line 1622:
```suggestion
1. Let |scopeSet| be « "<code>[=rate-limit scope/source=]</code>" ».
1. If |newRecord|'s [=attribution rate-limit record/scope=] is "<code>[=rate-limit scope/event-attribution=]</code>" 
     or "<code>[=rate-limit scope/aggregatable-attribution=]</code>":
     1. Set |max| to [=max attribution reporting origins per rate-limit window=].
     1. Set |scopeSet| to « "<code>[=rate-limit scope/event-attribution=]</code>", "<code>[=rate-limit scope/aggregatable-attribution=]</code>" ».
```

index.bs
```diff
@@ -1036,6 +1037,8 @@ An <dfn>attribution rate-limit record</dfn> is a [=struct=] with the following i
 :: A [=moment=].
 : <dfn>expiry time</dfn>
 :: Null or a [=moment=].
+: <dfn>entity ID</dfn>
```
apasel422, line 1040: I think we should call this `event-level report ID` for clarity.

apasel422 (approved):

index.bs
vikassahu29: Added a note for this relation.

apasel422 (approved):

Issue #1202, opened 2024-03-19
Spec unintentionally coalesces ARA registration headers


apasel422:
In https://wicg.github.io/attribution-reporting-api/#process-an-attribution-eligible-response we call https://fetch.spec.whatwg.org/#concept-header-list-get, but that algorithm "coalesces" the given header when it appears multiple times in a response. In particular, the individual values are joined by `, `.

Because our headers contain JSON, this almost always results in an invalid value, though it ends up having weird edge cases like

```http
Attribution-Reporting-Register-Source: {"destination":["https://a.test"
Attribution-Reporting-Register-Source: "https://b.test"]}
```
that end up working.

We never intended this, and we should consider making our headers non-coalescing, and either explicitly allowing multiple instances per response, allowing only the first and dropping any additional, or rejecting all if there are multiple.

domenic:
I am worried that this change works against the semantics of HTTP, which say that the multi-header and single-header-with-comma forms are equivalent. And in particular HTTP assumes that intermediaries can split or combine as necessary.

The reason the Fetch spec created high-level helpers which normalize this for you, is to ensure that the whole web platform behaves uniformly in this regard. Adding a special exception for your API seems problematic.

/cc @annevk @mnot to give more context.

mnot:
Yes - HTTP software (intermediaries, client APIs, server APIs) can and will coalesce headers with the same name, and there isn't a way to guarantee this won't happen on the open Internet. This is one of the reasons JSON doesn't work well in headers.

mnot:
`Set-Cookie` is an exception in the protocol; there won't be any more added. See:

  https://httpwg.org/specs/rfc9110.html#field.lines

In particular:

> A recipient MAY combine multiple field lines within a field section that have the same field name into one field line, without changing the semantics of the message, by appending each subsequent field line value to the initial field line value in order, separated by a comma (",") and optional whitespace ([OWS](https://httpwg.org/specs/rfc9110.html#whitespace), defined in [Section 5.6.3](https://httpwg.org/specs/rfc9110.html#whitespace)). For consistency, use comma SP.

Because a recipient may not know the semantics of your field (for example, in an intermediary or generic server library), you can't stop this from happening reliably. 

You _can_ specify that a field only has one value -- if there's more than one `Date`, for example, it's syntactically invalid and so is ignored.

[...]

PR #1212, opened 2024-03-22
Prevent coalescing for web source/trigger headers

agarant:
fixes: https://github.com/WICG/attribution-reporting-api/issues/1202

- Prevent header coalescing when receiving the web registration headers.
- Reject the headers if more than 1 is received.

***
<a href="https://pr-preview.s3.amazonaws.com/agarant/attribution-reporting-api/pull/1212.html" title="Last updated on Mar 27, 2024, 3:39 PM UTC (30f5805)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/attribution-reporting-api/1212/7e8b4c9...agarant:30f5805.html" title="Last updated on Mar 27, 2024, 3:39 PM UTC (30f5805)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -67,7 +67,6 @@ spec: structured header; type: dfn; urlPrefix: https://httpwg.org/specs/rfc8941;
         text: token; url: #token
 spec: infra; type: dfn; urlPrefix: https://infra.spec.whatwg.org/
     text: starts with; url: #string-starts-with
-    text: byte case insensitive; url: #byte-case-insensitive
 spec: fenced-frame; type: dfn; urlPrefix: https://wicg.github.io/fenced-frame/;
     for: navigable
         text: top-level traversable; url: #navigable-top-level-traversable
@@ -1839,17 +1838,16 @@ a [=header value=] or null |webTriggerHeader|, a [=header value=] or null
 
      </dl>
 
-To <dfn export>get the first non-coalesced header</dfn> given a [=response/header list=] |headers|,
+To <dfn>get the first non-coalesced header</dfn> given a [=response/header list=] |headers|
 and a [=header name=] header:
 1. Let |non-coalesced-headers| be the result of [=get non-coalesced headers=] with |headers| and |header|.
 1. If |non-coalesced-headers| [=list/is empty=], return null.
 1. Return |non-coalesced-headers|[0].
 
-To <dfn export>get non-coalesced headers</dfn> given a [=response/header list=] |headers|,
+To <dfn>get non-coalesced headers</dfn> given a [=response/header list=] |headers|
 and a [=header name=] header:
-1. If |headers| does not [=header list/contain=] |header|, return a new [=list=].
 1. Return the [=header value|values=] of all [=header|headers=] in |headers| whose [=header name|name=]
-    is a [=byte case insensitive=] match for |header|, in order.
+    is a [=byte-case-insensitive=] match for |header|, in order.
 
 To <dfn export>process an attribution eligible response</dfn> given a [=suitable origin=]
 |contextOrigin|, an [=eligibility=] |eligibility|, and a [=response=] |response|:
--- a/index.bs
+++ b/index.bs
@@ -1840,10 +1840,10 @@ a [=header value=] or null |webTriggerHeader|, a [=header value=] or null
 
      </dl>
 
-To <dfn>get the first non-coalesced header</dfn> given a [=response/header list=] |headers|
+To <dfn>get non-coalesced header</dfn> given a [=response/header list=] |headers|
 and a [=header name=] header:
 1. Let |non-coalesced-headers| be the result of [=get non-coalesced headers=] with |headers| and |header|.
-1. If |non-coalesced-headers| [=list/is empty=], return null.
+1. If |non-coalesced-headers|'s [=list/size=] is not equal to 1, return null.
 1. Return |non-coalesced-headers|[0].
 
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -1838,6 +1839,18 @@ a [=header value=] or null |webTriggerHeader|, a [=header value=] or null
...
+and a [=header name=] header:
+1. Let |non-coalesced-headers| be the result of [=get non-coalesced headers=] with |headers| and |header|.
+1. If |non-coalesced-headers| [=list/is empty=], return null.
+1. Return |non-coalesced-headers|[0].
+
+To <dfn export>get non-coalesced headers</dfn> given a [=response/header list=] |headers|,
+and a [=header name=] header:
+1. If |headers| does not [=header list/contain=] |header|, return a new [=list=].
+1. Return the [=header value|values=] of all [=header|headers=] in |headers| whose [=header name|name=]
+    is a [=byte case insensitive=] match for |header|, in order.
```
apasel422, line 1852:
```suggestion
    is a [=byte-case-insensitive=] match for |header|, in order.
```

index.bs
```diff
@@ -1838,6 +1839,18 @@ a [=header value=] or null |webTriggerHeader|, a [=header value=] or null
 
      </dl>
 
+To <dfn export>get the first non-coalesced header</dfn> given a [=response/header list=] |headers|,
+and a [=header name=] header:
+1. Let |non-coalesced-headers| be the result of [=get non-coalesced headers=] with |headers| and |header|.
+1. If |non-coalesced-headers| [=list/is empty=], return null.
+1. Return |non-coalesced-headers|[0].
+
+To <dfn export>get non-coalesced headers</dfn> given a [=response/header list=] |headers|,
```
apasel422, line 1848: I don't think we want to export these algorithms; ideally they'd ultimately be part of the `fetch` spec itself.

index.bs
```diff
@@ -1838,6 +1839,18 @@ a [=header value=] or null |webTriggerHeader|, a [=header value=] or null
 
      </dl>
 
+To <dfn export>get the first non-coalesced header</dfn> given a [=response/header list=] |headers|,
```
apasel422, line 1842:
```suggestion
To <dfn export>get the first non-coalesced header</dfn> given a [=response/header list=] |headers|
```

index.bs
apasel422, line 1848:
```suggestion
To <dfn export>get non-coalesced headers</dfn> given a [=response/header list=] |headers|
```

index.bs
```diff
@@ -1838,6 +1839,18 @@ a [=header value=] or null |webTriggerHeader|, a [=header value=] or null
...
+and a [=header name=] header:
+1. Let |non-coalesced-headers| be the result of [=get non-coalesced headers=] with |headers| and |header|.
+1. If |non-coalesced-headers| [=list/is empty=], return null.
+1. Return |non-coalesced-headers|[0].
+
+To <dfn export>get non-coalesced headers</dfn> given a [=response/header list=] |headers|,
+and a [=header name=] header:
+1. If |headers| does not [=header list/contain=] |header|, return a new [=list=].
```
apasel422, line 1850:
```suggestion
```
There's no reason to explicitly handle this case, since the subsequent step will return an empty list.

apasel422 (changes requested):
Can we rename the PR to something like `Prevent coalescing for web source/trigger headers`?

index.bs
agarant: It is defined at the top with "byte case insensitive".

apasel422: Right, but that's not how it's spelled in the fetch spec: https://infra.spec.whatwg.org/#byte-case-insensitive

agarant: I removed the mapping and use the link directly instead.

index.bs
```diff
@@ -1840,10 +1840,10 @@ a [=header value=] or null |webTriggerHeader|, a [=header value=] or null
 
      </dl>
 
-To <dfn>get the first non-coalesced header</dfn> given a [=response/header list=] |headers|
+To <dfn>get non-coalesced header</dfn> given a [=response/header list=] |headers|
```
apasel422, line 1843: `get non-coalesced header if unique` might be more clear.

index.bs
```diff
@@ -1860,10 +1871,10 @@ To <dfn export>process an attribution eligible response</dfn> given a [=suitable
     "<code>[=eligibility/event-source-or-trigger=]</code>".
 1. Let |reportingOrigin| be |response|'s [=response/URL=]'s [=url/origin=].
 1. If |reportingOrigin| is not [=check if an origin is suitable|suitable=], return.
-1. Let |sourceHeader| be the result of [=header list/get|getting=]
+1. Let |sourceHeader| be the result of [=get non-coalesced header|getting=]
     "`Attribution-Reporting-Register-Source`" from |response|'s
     [=response/header list=].
-1. Let |triggerHeader| be the result of [=header list/get|getting=]
+1. Let |triggerHeader| be the result of [=get non-coalesced header|getting=]
```
apasel422, line 1877: I don't think this properly handles:

```http
Attribution-Reporting-Register-Source: {...}
Attribution-Reporting-Register-Source: {...}
Attribution-Reporting-Register-Trigger: {...}
```
In this case, the two sources will be ignored but the trigger will be handled. IMO this should be an error that causes all the headers to be ignored instead.

apasel422 (changes requested):

apasel422 (approved):

PR #1217, opened 2024-03-26
Deactivate earliest destinations if exceeding distinct destination limit for unexpired sources

### LINKED ISSUE #1228: Consider allowing the destination limit to favor new sources
The [max destinations for unexpired sources limit](https://wicg.github.io/attribution-reporting-api/#max-destinations-covered-by-unexpired-sources) is currently spec'd such that:

once a publisher site registers sources for 100 different destinations, all subsequent source registrations will fail

Registrations will fail until these sources expire or until they are attributed. This has the downside that it locks out registration on a site.

We may want to consider, instead of blocking new sources, to kick out old sources to make space. This could be done by formulating this limit as a FIFO queue.

linnan-github:
Fixes #1228 

To preserve the protection of user’s browser history, and also to mitigate utility concern with dropping new sources, we propose to change to FIFO+priority for the destination limit for unexpired sources. This change backwards incompatible.

We also propose a new limit on the max destinations per reporting site per day to mitigate the history reconstruction attack for FIFO.

***
<a href="https://pr-preview.s3.amazonaws.com/linnan-github/conversion-measurement-api/pull/1217.html" title="Last updated on Jun 10, 2024, 2:32 PM UTC (fd1758b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/attribution-reporting-api/1217/221f474...linnan-github:fd1758b.html" title="Last updated on Jun 10, 2024, 2:32 PM UTC (fd1758b)">Diff</a>

### DIFF
--- a/EVENT.md
+++ b/EVENT.md
@@ -1148,8 +1148,7 @@ attack, the browser should maintain these limits per reporting site. This
 effectively limits the number of unique sites covered per {source site, reporting site} applied to all unexpired sources regardless of type at source time.
 
 The browser can also limit the number of `destination` sites per {source site, reporting site, 1 day}
-to mitigate the history reconstruction attack as new `destination` sites can be
-registered when the `priority_fifo` algorithm is used.
+to mitigate the history reconstruction attack.
 
 #### Limiting the number of unique destinations per source site
 
--- a/EVENT.md
+++ b/EVENT.md
@@ -1115,30 +1115,23 @@ sites represented by unexpired sources for a source-site.
 
 The browser can place a limit on the number of a source site's unexpired source's
 unique `destination` sites. Source registrations will accept an optional field
-`destination_limit` to allow developers to select the behavior when an attribution
-source is registered for a site that is not already registered by any of the
-unexpired sources and a source site is at its limit.
+`destination_limit_priority` to allow developers to prioritize the destinations
+registered with this source with respect to other destinations for the purpose
+of source deactivation.
 
 ```jsonc
 {
   ..., // existing fields
-  "destination_limit": {
-    "algorithm": "lifo", // or "priority_fifo", defaults to "lifo" if not present
-    "priority": "[64-bit signed integer]" // defaults to 0 if not present
-  }
+  "destination_limit_priority": "[64-bit signed integer]" // defaults to 0 if not present
 }
 ```
 
-The `priority` field is used to prioritize the destinations registered with this
-source with respect to other destinations for the purpose of source deactivation.
-
-The `algorithm` field is used to select the algorithm to determine the final
-`destination` sites that the browser selects. When "lifo" is used, the browser
-will drop the new source if the source site is at its limit. When "priority_fifo"
-is used, the browser will sort the `destination` sites registered by unexpired
-sources, including the new source, by `priority` in descending order and by
-the registration time in descending order. The browser will then select the
-first few `destination` sites within this limit, and delete pending sources and
+When an attribution source is registered for a site that is not already in the
+unexpired sources and a source site is at its limit, the browser will sort the
+`destination` sites registered by unexpired sources, including the new source,
+by `destination_limit_priority` in descending order and by the registration
+time in descending order. The browser will then select the first few
+`destination` sites within this limit, and delete pending sources and
 aggregatable reports associated with the unselected `destination` sites. The
 event-level reports are not deleted as the leak of user's browsing history is
 mitigated by fake reports within differential privacy.
--- a/EVENT.md
+++ b/EVENT.md
@@ -1132,8 +1132,8 @@ unexpired sources and a source site is at its limit, the browser will sort the
 by `destination_limit_priority` in descending order and by the registration
 time in descending order. The browser will then select the first few
 `destination` sites within this limit, and delete pending sources and
-aggregatable reports associated with the unselected `destination` sites. The
-event-level reports are not deleted as the leak of user's browsing history is
+aggregatable reports associated with the unselected `destination` sites. Any
+event-level reports are not deleted, as the leak of user's browsing history is
 mitigated by fake reports within differential privacy.
 
 The lower this value, the harder it is for a reporting origin to use the API to
--- a/EVENT.md
+++ b/EVENT.md
@@ -1114,9 +1114,34 @@ trying to measure user visits on, the browser can limit the number `destination`
 sites represented by unexpired sources for a source-site.
 
 The browser can place a limit on the number of a source site's unexpired source's
-unique `destination` sites. When an attribution source is registered for a site
-that is not already in the unexpired sources and a source site is at its limit,
-the browser will drop the new source.
+unique `destination` sites. Source registrations will accept an optional field
+`destination_limit` to allow developers to select the behavior when an attribution
+source is registered for a site that is not already registered by any of the
+unexpired sources and a source site is at its limit.
+
+```jsonc
+{
+  ..., // existing fields
+  "destination_limit": {
+    "algorithm": "lifo", // or "priority_fifo", defaults to "lifo" if not present
+    "priority": "[64-bit signed integer]" // defaults to 0 if not present
+  }
+}
+```
+
+The `priority` field is used to prioritize the destinations registered with this
+source with respect to other destinations for the purpose of source deactivation.
+
+The `algorithm` field is used to select the algorithm to determine the final
+`destination` sites that the browser selects. When "lifo" is used, the browser
+will drop the new source if the source site is at its limit. When "priority_fifo"
+is used, the browser will sort the `destination` sites registered by unexpired
+sources, including the new source, by `priority` in descending order and by
+the registration time in descending order. The browser will then select the
+first few `destination` sites within this limit, and delete pending sources and
+aggregatable reports associated with the unselected `destination` sites. The
+event-level reports are not deleted as the leak of user's browsing history is
+mitigated by fake reports within differential privacy.
 
 The lower this value, the harder it is for a reporting origin to use the API to
 try and measure user browsing activity not associated with ads being shown.
@@ -1128,6 +1153,10 @@ origin on a site to push the other attribution sources out of the browser. See
 the [denial of service](#denial-of-service) for more details. To prevent this
 attack, the browser should maintain these limits per reporting site. This
 effectively limits the number of unique sites covered per {source site, reporting site} applied to all unexpired sources regardless of type at source time.
+
+The browser can also limit the number of `destination` sites per {source site, reporting site, 1 day}
+to mitigate the history reconstruction attack as new `destination` sites can be
+registered when the `priority_fifo` algorithm is used.
 
 #### Limiting the number of unique destinations per source site
 
--- a/index.bs
+++ b/index.bs
@@ -2375,7 +2375,7 @@ A <dfn>source-registration JSON key</dfn> is one of the following:
 <li>"<dfn><code>debug_key</code></dfn>"
 <li>"<dfn><code>debug_reporting</code></dfn>"
 <li>"<dfn><code>destination</code></dfn>"
-<li>"<dfn><code>destination_limit</code></dfn>"
+<li>"<dfn><code>destination_limit_priority</code></dfn>"
 <li>"<dfn><code>end_times</code></dfn>"
 <li>"<dfn><code>event_level_epsilon</code></dfn>"
 <li>"<dfn><code>event_report_window</code></dfn>"
@@ -2681,19 +2681,10 @@ To <dfn noexport>parse source-registration JSON</dfn> given a [=byte sequence=]
     [=parse an optional 64-bit signed integer=] with |value|, "<code>[=source-registration JSON key/priority=]</code>", and
     0.
 1. If |priority| is an error, return null.
-1. Let |destinationLimitAlgorithm| be "<code>[=destination limit algorithm/lifo=]</code>".
-1. Let |destinationLimitPriority| be 0.
-1. If |value|["<code>[=source-registration JSON key/destination_limit=]</code>"] [=map/exists=]:
-    1. Let |destinationLimit| be |value|["<code>[=source-registration JSON key/destination_limit=]</code>"].
-    1. If |destinationLimit| is not a [=map=], return null.
-    1. Set |destinationLimitPriority| to the result of running
-        [=parse an optional 64-bit signed integer=] with |destinationLimit|,
-        "<code>[=source-registration JSON key/priority=]</code>", and 0.
-    1. If |destinationLimitPriority| is an error, return null.
-    1. If |destinationLimit|["<code>[=source-registration JSON key/algorithm=]</code>"] [=map/exists=]:
-        1. If |destinationLimit|["<code>[=source-registration JSON key/algorithm=]</code>"] is not
-            a [=destination limit algorithm=], return null.
-        1. Set |destinationLimitAlgorithm| to |destinationLimit|["<code>[=source-registration JSON key/algorithm=]</code>"].
+1. Let |destinationLimitPriority| be the result of running
+    [=parse an optional 64-bit signed integer=] with |value|,
+    "<code>[=source-registration JSON key/destination_limit_priority=]</code>", and 0.
+1. If |destinationLimitPriority| is an error, return null.
 1. Let |filterData| be a new [=filter map=].
 1. If |value|["<code>[=source-registration JSON key/filter_data=]</code>"] [=map/exists=]:
     1. Set |filterData| to the result of running [=parse filter data=] with
@@ -3018,8 +3009,7 @@ optional [=boolean=] |isNoised| (default false), and an optional [=boolean=]
         [=serialize an integer|serialized=].
 
     Note: The "`source_destination_limit`" field may be included to indicate that
-    [=max destinations covered by unexpired sources=] was hit and
-    the [=destination limit algorithm/priority_fifo=] algorithm was applied, which is not
+    [=max destinations covered by unexpired sources=] was hit, which is not
     reported as "<code>[=source debug data type/source-destination-limit=]</code>" to prevent side-channel
     leakage of cross-origin data.
 
@@ -3133,16 +3123,7 @@ To <dfn>process an attribution source</dfn> given an [=attribution source=] |sou
 1. Let |sourcesToDeleteForDestinationLimit| be the result of running [=get sources to delete for the unexpired destination limit=]
     with |source|.
 1. Let |destinationLimitHit| be false if |sourcesToDeleteForDestinationLimit| [=set/is empty=], true otherwise.
-1. Let |dropNewSource| be false.
-1. If |source|'s [=attribution source/destination limit algorithm=] is:
-    <dl class="switch">
-    : "<code>[=destination limit algorithm/priority_fifo=]</code>"
-    :: Set |dropNewSource| to true if |sourcesToDeleteForDestinationLimit| [=set/contains=] |source|'s [=attribution source/source identifier=].
-    : "<code>[=destination limit algorithm/lifo=]</code>"
-    :: Set |dropNewSource| to |destinationLimitHit|.
-
-    </dl>
-1. If |dropNewSource| is true:
+1. If |sourcesToDeleteForDestinationLimit| [=set/contains=] |source|'s [=attribution source/source identifier=]:
     1. Run [=obtain and deliver a verbose debug report on source registration=] with "[=source debug data type/source-destination-limit=]</code>" and |source|.
     1. Return.
 1. Run [=delete sources for unexpired destination limit=] with |sourcesToDeleteForDestinationLimit| and |source|'s [=attribution source/source time=].
@@ -3150,7 +3131,7 @@ To <dfn>process an attribution source</dfn> given an [=attribution source=] |sou
 1. Let |isNoised| be true if |source|'s [=attribution source/randomized response=]
     is not null, otherwise false.
 1. If |destinationRateLimitResult| is "<code>[=destination rate-limit result/hit global limit=]</code>":
-    1. Run [=obtain and deliver a verbose debug report on source registration=] with "<code>[=source debug data type/source-success=]</code>",
+    1. Run [=obtain and deliver a verbose debug report on source registration=] with "<code>[=source debug data type/source-destination-global-rate-limit=]</code>",
         |source|, |isNoised|, and |destinationLimitHit|.
     1. Return.
 1. Let |newRateLimitRecords| be a new [=set=].
@@ -3174,7 +3155,7 @@ To <dfn>process an attribution source</dfn> given an [=attribution source=] |sou
         :: |source|'s [=attribution source/destination limit priority=]
     1. If the result of running [=should processing be blocked by reporting-origin limit=] with
         |rateLimitRecord| is <strong>blocked</strong>:
-        1. Run [=obtain and deliver a verbose debug report on source registration=] with "<code>[=source debug data type/source-success=]</code>",
+        1. Run [=obtain and deliver a verbose debug report on source registration=] with "<code>[=source debug data type/source-reporting-origin-limit=]</code>",
             |source|, |isNoised|, and |destinationLimitHit|.
         1. Return.
     1. [=set/Append=] |rateLimitRecord| to |newRateLimitRecords|.
--- a/index.bs
+++ b/index.bs
@@ -3093,7 +3093,7 @@ and a [=boolean=] |isNoised|:
 
 To <dfn>obtain and deliver debug reports on source registration</dfn>
 given a [=set=] of [=source debug data types=] |dataTypes|, an [=attribution source=] |source|,
-, and an optional [=boolean=] |isNoised| (default false):
+and an optional [=boolean=] |isNoised| (default false):
 
 1. Run [=obtain and deliver a verbose debug report on source registration=]
     with |dataTypes|, |source|, and |isNoised|.
@@ -3226,7 +3226,7 @@ To <dfn>process an attribution source</dfn> given an [=attribution source=] |sou
 1. [=set/Append=] "<code>[=source debug data type/source-success=]</code>" to
     |debugDataTypes|.
 1. Run [=obtain and deliver debug reports on source registration=] with
-    |debugDataTypes|, |source| and |isNoised|.
+    |debugDataTypes|, |source|, and |isNoised|.
 1. [=set/Append=] |source| to |cache|.
 
 Note: Because a fake report does not have a "real" effective destination, we need to subtract from the
--- a/index.bs
+++ b/index.bs
@@ -2858,12 +2858,12 @@ To <dfn>delete sources for unexpired destination limit</dfn> given a [=set=] of
 1. [=set/iterate|For each=] [=attribution source=] |source| of the [=attribution source cache=]:
     1. [=set/Remove=] |source| from the [=attribution source cache=] if |sourcesToDelete|
         [=set/contains=] |source|'s [=attribution source/source identifier=].
+1. Let |deletedEventLevelReports| be a new [=set=].
 1. [=set/iterate|For each=] [=event-level report=] |report| of the [=event-level report cache=]:
-    1. If |report|'s [=event-level report/trigger time=] is greater than or equal to |now|,
-        [=set/remove=] |report| from the [=event-level report cache=].
-
-        Note: This would only delete [=obtain a fake report|fake reports=] that would never be
-        stored if they were true [=event-level reports=].
+    1. If |sourcesToDelete| [=set/contains=] |report|'s [=event-level report/source identifier=]
+        and |report|'s [=event-level report/trigger time=] is greater than or equal to |now|:
+        1. [=set/Append=] |report|'s [=event-level report/report ID=] to |deletedEventLevelReports|.
+        1. [=set/Remove=] |report| from the [=event-level report cache=].
 1. Let |deletedAggregatableReports| be a new [=set=].
 1. [=set/iterate|For each=] [=aggregatable attribution report=] |report| of the [=aggregatable attribution report cache=]:
     1. If |sourcesToDelete| [=set/contains=] |report|'s [=aggregatable attribution report/source identifier=]:
@@ -2875,6 +2875,9 @@ To <dfn>delete sources for unexpired destination limit</dfn> given a [=set=] of
         : "<code>[=rate-limit scope/source=]</code>"
         :: Set |record|'s [=attribution rate-limit record/deactivated for unexpired destination limit=] to
             true if |sourcesToDelete| [=set/contains=] |record|'s [=attribution rate-limit record/entity ID=].
+        : "<code>[=rate-limit scope/event-attribution=]</code>"
+        :: [=set/Remove=] |record| from the [=attribution rate-limit cache=] if
+            |deletedEventLevelReports| [=set/contains=] |record|'s [=attribution rate-limit record/entity ID=].
         : "<code>[=rate-limit scope/aggregatable-attribution=]</code>"
         :: [=set/Remove=] |record| from the [=attribution rate-limit cache=] if
             |deletedAggregatableReports| [=set/contains=] |record|'s [=attribution rate-limit record/entity ID=].
--- a/index.bs
+++ b/index.bs
@@ -843,6 +843,8 @@ An attribution source is a [=struct=] with the following items:
 :: Number of [=aggregatable debug reports=] created for this [=attribution source=].
 : <dfn>aggregatable debug reporting config</dfn>
 :: An [=aggregatable debug reporting config=].
+: <dfn>destination limit priority</dfn>
+:: A 64-bit integer.
 
 </dl>
 
@@ -995,7 +997,9 @@ An <dfn>attribution debug info</dfn> is a [=tuple=] with the following items:
 : <dfn>source debug key</dfn>
 :: Null or a [=string=].
 : <dfn>trigger debug key</dfn>
-:: Null or a [=string=].
+:: Null or a non-negative 64-bit integer.
+: <dfn>source identifier</dfn>
+:: A [=string=].
 
 </dl>
 
@@ -1016,8 +1020,6 @@ An event-level report is an [=attribution report=] with the following additional
 :: A 64-bit integer.
 : <dfn>trigger time</dfn>
 :: A [=moment=].
-: <dfn>source identifier</dfn>
-:: A string.
 : <dfn>attribution destinations</dfn>
 :: A [=set=] of [=sites=].
 : <dfn>attribution debug info</dfn>
@@ -1094,8 +1096,14 @@ An <dfn>attribution rate-limit record</dfn> is a [=struct=] with the following i
 :: A [=moment=].
 : <dfn>expiry time</dfn>
 :: Null or a [=moment=].
-: <dfn>event-level report ID</dfn>
-:: Null or an [=event-level report=]'s [=event-level report/report ID=].
+: <dfn>entity ID</dfn>
+:: Null for [=obtain a fake report|fake reports=] or an [=event-level report=]'s [=event-level report/report ID=] or an
+    [=aggregatable report=]'s [=aggregatable report/report ID=] or an
+    [=attribution source=]'s [=attribution source/source identifier=].
+: <dfn>deactivated for unexpired destination limit</dfn> (default false)
+:: A [=boolean=].
+: <dfn>destination limit priority</dfn> (default null)
+:: Null or a 64-bit integer.
 
 </dl>
 
@@ -1128,6 +1136,7 @@ Possible values are:
 <li>"<dfn><code>source-destination-global-rate-limit</code></dfn>"
 <li>"<dfn><code>source-destination-limit</code></dfn>"
 <li>"<dfn><code>source-destination-rate-limit</code></dfn>"
+<li>"<dfn><code>source-destination-per-day-rate-limit</code></dfn>"
 <li>"<dfn><code>source-noised</code></dfn>"
 <li>"<dfn><code>source-reporting-origin-limit</code></dfn>"
 <li>"<dfn><code>source-reporting-origin-per-site-limit</code></dfn>"
@@ -1433,6 +1442,11 @@ for [=attribution sources=] with a given [=attribution source/source site=] per 
 The second controls the maximum number of distinct [=sites=] across all [=attribution source/attribution destinations=]
 for [=attribution sources=] with a given ([=attribution source/source site=], [=attribution source/reporting origin=] [=site=])
 per [=destination rate-limit window=].
+
+<dfn>Max destinations per source reporting site per day</dfn> is an integer controls
+the maximum number of distinct [=sites=] across all [=attribution source/attribution destinations=]
+for [=attribution sources=] with a given ([=attribution source/source site=], [=attribution source/reporting origin=] [=site=])
+per day.
 
 <dfn>Max source reporting origins per rate-limit window</dfn> is a positive
 integer that controls the maximum number of distinct
@@ -2361,6 +2375,7 @@ A <dfn>source-registration JSON key</dfn> is one of the following:
 <li>"<dfn><code>debug_key</code></dfn>"
 <li>"<dfn><code>debug_reporting</code></dfn>"
 <li>"<dfn><code>destination</code></dfn>"
+<li>"<dfn><code>destination_limit</code></dfn>"
 <li>"<dfn><code>end_times</code></dfn>"
 <li>"<dfn><code>event_level_epsilon</code></dfn>"
 <li>"<dfn><code>event_report_window</code></dfn>"
@@ -2666,6 +2681,19 @@ To <dfn noexport>parse source-registration JSON</dfn> given a [=byte sequence=]
     [=parse an optional 64-bit signed integer=] with |value|, "<code>[=source-registration JSON key/priority=]</code>", and
     0.
 1. If |priority| is an error, return null.
+1. Let |destinationLimitAlgorithm| be "<code>[=destination limit algorithm/lifo=]</code>".
+1. Let |destinationLimitPriority| be 0.
+1. If |value|["<code>[=source-registration JSON key/destination_limit=]</code>"] [=map/exists=]:
+    1. Let |destinationLimit| be |value|["<code>[=source-registration JSON key/destination_limit=]</code>"].
+    1. If |destinationLimit| is not a [=map=], return null.
+    1. Set |destinationLimitPriority| to the result of running
+        [=parse an optional 64-bit signed integer=] with |destinationLimit|,
+        "<code>[=source-registration JSON key/priority=]</code>", and 0.
+    1. If |destinationLimitPriority| is an error, return null.
+    1. If |destinationLimit|["<code>[=source-registration JSON key/algorithm=]</code>"] [=map/exists=]:
+        1. If |destinationLimit|["<code>[=source-registration JSON key/algorithm=]</code>"] is not
+            a [=destination limit algorithm=], return null.
+        1. Set |destinationLimitAlgorithm| to |destinationLimit|["<code>[=source-registration JSON key/algorithm=]</code>"].
 1. Let |filterData| be a new [=filter map=].
 1. If |value|["<code>[=source-registration JSON key/filter_data=]</code>"] [=map/exists=]:
     1. Set |filterData| to the result of running [=parse filter data=] with
@@ -2772,6 +2800,8 @@ To <dfn noexport>parse source-registration JSON</dfn> given a [=byte sequence=]
     :: |aggregatableDebugBudget|
     : [=attribution source/aggregatable debug reporting config=]
     :: |aggregatableDebugReportingConfig|
+    : [=attribution source/destination limit priority=]
+    :: |destinationLimitPriority|
 1. Return |source|.
 
 Issue: Determine proper charset-handling for the JSON header value.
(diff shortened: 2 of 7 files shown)


index.bs
```diff
@@ -1036,6 +1036,10 @@ An <dfn>attribution rate-limit record</dfn> is a [=struct=] with the following i
 :: A [=moment=].
 : <dfn>expiry time</dfn>
 :: Null or a [=moment=].
+: <dfn>entity ID</dfn>
+:: Null or an ID.
```
apasel422, line 1040: What is an `ID`?

index.bs
```diff
@@ -1036,6 +1036,10 @@ An <dfn>attribution rate-limit record</dfn> is a [=struct=] with the following i
...
+: <dfn>deactivated</dfn> (default false)
```
apasel422, line 1041: The spec to this point has avoided the concept of source activation. Is it really necessary now?

apasel422 (changes requested):
Please provide motivation for this and/or a linked issue in the PR description.

index.bs
linnan-github: This PR is to be rebased on https://github.com/WICG/attribution-reporting-api/pull/1211 and this is just a placeholder.

index.bs
linnan-github: It's a different concept of "deactivation", it's to deactivate a source rate-limit record for the purpose of couting the distinct destinations for unexpired sources.

apasel422: Should we give the field a more specific name if it's only deactivated with respect to a particular limit? Otherwise it seems like every rate-limit algorithm needs to inspect this field.

linnan-github: Yeah I’m thinking about a more specific name as well, like “deactivated for unique destination limit for unexpired sources”. Please feel free to suggest a better name if any. Thanks.

index.bs
```diff
@@ -2385,19 +2389,66 @@ Note: We do not emit an explicit [=source debug data type=] for "<code>[=destina
...
+An [=attribution rate-limit record=] |a| <dfn for="attribution rate-limit record">is less than</dfn>
+an [=attribution rate-limit record=] |b| if any of the following are true:
+
+* |a|'s [=attribution rate-limit record/scope=] is less than |b|'s [=attribution rate-limit record/scope=].
+* |a|'s [=attribution rate-limit record/scope=] is equal to |b|'s [=attribution rate-limit record/scope=]
+    and |a|'s [=attribution rate-limit record/time=] is less than |b|'s [=attribution rate-limit record/time=].
+* |a|'s [=attribution rate-limit record/scope=] is equal to |b|'s [=attribution rate-limit record/scope=],
+    |a|'s [=attribution rate-limit record/time=] is equal to |b|'s [=attribution rate-limit record/time=],
+    and |a|'s <a href="https://html.spec.whatwg.org/multipage/origin.html#serialization-of-a-site">serialized</a> [=attribution rate-limit record/attribution destination=]
+    is less than |b|'s <a href="https://html.spec.whatwg.org/multipage/origin.html#serialization-of-a-site">serialized</a> [=attribution rate-limit record/attribution destination=].
```
apasel422, line 2401: I think this would be easier to understand if it were written with early returns, e.g.

```
1. If |a|'s scope is less than |b|'s scope, return true.
2. If |a|'s scope is greater than |b|'s scope, return false.
3. ...
```

index.bs
```diff
@@ -2385,19 +2389,66 @@ Note: We do not emit an explicit [=source debug data type=] for "<code>[=destina
...
+
+To <dfn>delete sources for unexpired destination limit</dfn> given a [=set=] of [=attribution source/source identifiers=] |deletedSources|:
+
+1. Let |deletedEventLevelReports| be a new [=set/is empty|empty=] [=set=].
+1. Let |deletedAggregatableReports| be a new [=set/is empty|empty=] [=set=].
+1. [=set/iterate|For each=] |destination| of |deletedDestinations|:
+    1. [=list/iterate|For each=] [=event-level report=] |report| of the [=event-level report cache=]:
```
apasel422:
```suggestion
    1. [=set/iterate|For each=] [=event-level report=] |report| of the [=event-level report cache=]:
```

index.bs
```diff
@@ -2385,19 +2389,66 @@ Note: We do not emit an explicit [=source debug data type=] for "<code>[=destina
...
+        1. If |deletedSources| [=set/contains=] |report|'s [=event-level report/source identifier=]:
+            1. [=set/Append=] |report|'s [=attribution report/report ID=] to |deletedEventLevelReports|.
+            1. [=list/Remove=] |report| from the [=event-level report cache=].
```
apasel422:
```suggestion
            1. [=set/Remove=] |report| from the [=event-level report cache=].
```

index.bs
```diff
@@ -2385,19 +2389,66 @@ Note: We do not emit an explicit [=source debug data type=] for "<code>[=destina
...
+    1. [=list/iterate|For each=] [=aggregatable report=] |report| of the [=aggregatable report cache=]:
```
apasel422:
```suggestion
    1. [=set/iterate|For each=] [=aggregatable report=] |report| of the [=aggregatable report cache=]:
```

apasel422 (changes requested):

index.bs
linnan-github: Updated as suggested, thanks.

EVENT.md
```diff
@@ -1110,7 +1110,10 @@ sites represented by unexpired sources for a source-site.
 The browser can place a limit on the number of a source site's unexpired source's
 unique `destination` sites. When an attribution source is registered for a site
 that is not already in the unexpired sources and a source site is at its limit,
-the browser will drop the new source.
+the browser will only count the most recent `destination` sites within this
+limit, including those registered with the new source. The browser will delete
+pending sources and reports associated with the `destination` sites that are no
```
giladbarkan-github, line 1115: If reports were already delivered for the source/s with pending reports, does this still address the history reconstruction attacks?

Why is no consideration given to the volume of sources and reports that may be deleted with strict FIFO?

linnan-github: @vikassahu29 may have better explanation, but with FIFO we ensure that at any point there are no more than 100 destinations sent in the reports for unexpired sources.

We would like to keep this limit purely based on time to be privacy neutral. Utility-wise, we expect that ad-techs most of time would favor more recent sources.

giladbarkan-github: Utility-wise, I would also expect API consumers to favour receiving reports. More recent sources don't necessarily guarantee that, and the volume of potential report deletion seems unaccounted for.

linnan-github: The volume of potential report deletion definitely contains cross-site leak, and I don't think we would consider this as a rule for FIFO.

linnan-github: Sources would be used for future attributions/reports though, and that's what we would like to limit.

giladbarkan-github: My understanding is we want to "ensure that at any point there are no more than 100 destinations sent in the reports." Regardless of what sources and triggers the user-agent gets, can't this be controlled by the reports themselves, theoretically?

EVENT.md
```diff
@@ -783,10 +783,10 @@ included in the request body as a JSON list of objects, e.g.:
 
 ```jsonc
 [{
-  "type": "<report type>", // e.g. "source-destination-limit"
+  "type": "<report type>", // e.g. "source-destination-rate-limit"
   "body": {
     "attribution_destination": "https://destination.example",
-    "limit": "100", // the browser's limit
+    "limit": "50", // the browser's limit
```
apasel422, line 789: Why are we changing this? It's just an example.

index.bs
```diff
@@ -1324,12 +1317,17 @@ for unexpired [=attribution sources=] with a given ([=attribution source/source
 <dfn>Destination rate-limit window</dfn> is a positive [=duration=] that
 controls the rate-limiting window for destinations.
 
-<dfn>Max destinations per rate-limit window</dfn> is a [=tuple=] consisting of two integers. The first
+<dfn>Max destinations per rate-limit window</dfn> is an integer that
```
apasel422, line 1320:
```suggestion
<dfn>Max destinations per rate-limit window</dfn> is a positive integer that
```

index.bs
```diff
@@ -1324,12 +1317,17 @@ for unexpired [=attribution sources=] with a given ([=attribution source/source
...
 controls the maximum number of distinct [=sites=] across all [=attribution source/attribution destinations=]
 for [=attribution sources=] with a given [=attribution source/source site=] per [=destination rate-limit window=].
-The second controls the maximum number of distinct [=sites=] across all [=attribution source/attribution destinations=]
+
+<dfn>Destination per reporting site rate-limit window</dfn> is a positive [=duration=] that
```
apasel422, line 1324:
```suggestion
<dfn>Destinations-per-reporting-site rate-limit window</dfn> is a positive [=duration=] that
```

index.bs
```diff
@@ -1324,12 +1317,17 @@ for unexpired [=attribution sources=] with a given ([=attribution source/source
...
+controls the per reporting site rate-limiting window for destinations.
```
apasel422, line 1325:
```suggestion
controls the per-reporting-site rate-limiting window for destinations.
```

index.bs
```diff
@@ -1324,12 +1317,17 @@ for unexpired [=attribution sources=] with a given ([=attribution source/source
...
+
+<dfn>Max destinations per reporting site per rate-limit window</dfn> is an integer that
```
apasel422, line 1327:
```suggestion
<dfn>Max destinations per reporting site per rate-limit window</dfn> is a positive integer that
```

index.bs
```diff
@@ -2410,45 +2408,91 @@ To <dfn>check if an [=attribution source=] exceeds the time-based destination li
...
+
+1. Let |matchingSources| be all [=attribution rate-limit records=] |record| in the [=attribution rate-limit cache=] where all of the following are true:
+
+     * |record|'s [=attribution rate-limit record/scope=] is "<code>[=rate-limit scope/source=]</code>"
+     * |record|'s [=attribution rate-limit record/source site=] and |source|'s [=attribution source/source site=] are equal
+     * |record|'s [=attribution rate-limit record/reporting origin=] and |source|'s [=attribution source/reporting origin=] are [=same site=]
+     * |record|'s [=attribution rate-limit record/expiry time=] is greater than |source|'s [=attribution source/source time=]
+     * The [=duration from=] |record|'s [=attribution rate-limit record/time=] and |source|'s [=attribution source/source time=]
+        is less than [=destination per reporting site rate-limit window=]
+1. Let |destinations| be the [=set=] of every [=attribution rate-limit record/attribution destination=] in |matchingSources|,
```
apasel422, line 2426:
```suggestion
1. Let |destinations| be the [=set/union=] of all [=attribution rate-limit record/attribution destination=] in |matchingSources|,
```

index.bs
```diff
@@ -2410,45 +2408,91 @@ To <dfn>check if an [=attribution source=] exceeds the time-based destination li
...
+1. Let |newDestinations| be the result of [=set/cloning=] |source|'s [=attribution source/attribution destinations=].
+1. Let |sourcesToDelete| be a new [=set=].
+1. [=set/iterate|For each=] [=attribution rate-limit record=] |unexpiredRecord| of |unexpiredSources|:
+    1. Let |destination| be |unexpiredRecord|'s [=attribution rate-limit record/attribution destination=].
+    1. If |newDestinations|'s [=set/size=] is less than the user agent's [=max destinations covered by unexpired sources=],
+        [=set/append=] |destination| to |newDestinations|.
+    1. Otherwise, if |newDestinations| does not [=set/contain=] |destination|:
+        1. [=set/Append=] |unexpiredRecord|'s [=attribution rate-limit record/entity ID=] to |sourcesToDelete|.
+1. Run [=delete sources for unexpired destination limit=] with |sourcesToDelete|.
+1. If |sourcesToDelete| [=set/is empty=] return false, otherwise return true.
```
apasel422, line 2495:
```suggestion
1. If |sourcesToDelete| [=set/is empty=], return false; otherwise, return true.
```

index.bs
```diff
@@ -2493,11 +2538,12 @@ To <dfn>obtain and deliver a debug report on source registration</dfn> given a
 
 1. If |dataType| is:
     <dl class="switch">
-    : "<code>[=source debug data type/source-destination-limit=]</code>"
-    :: [=map/Set=] |body|["`limit`"] to the user agent's [=max destinations covered by unexpired sources=],
-         [=serialize an integer|serialized=].
+    : "<code>[=source debug data type/source-success=]</code>"
+    : "<code>[=source debug data type/source-noised=]</code>"
+    :: If |destinationLimitHit| is true, [=map/set=] |body|["`source_destination_limit`"] to the user agent's [=max destinations covered by unexpired sources=],
```
apasel422, line 2543: Please add a non-normative note to explain why we're adding this field when the debug data type is `success` or `noised`.

index.bs
```diff
@@ -2619,10 +2664,10 @@ To <dfn>process an attribution source</dfn> given an [=attribution source=] |sou
             :: |source|'s [=attribution source/source time=]
             : [=attribution rate-limit record/expiry time=]
             :: null
-            : [=attribution rate-limit record/event-level report ID=]
+            : [=attribution rate-limit record/entity ID=]
             :: null
         1. [=set/Append=] |rateLimitRecord| to the [=attribution rate-limit cache=].
-1. Run [=obtain and deliver a debug report on source registration=] with |debugDataType| and |source|.
+1. Run [=obtain and deliver a debug report on source registration=] with |debugDataType|, |source| and |destinationLimitHit|.
```
apasel422, line 2670:
```suggestion
1. Run [=obtain and deliver a debug report on source registration=] with |debugDataType|, |source|, and |destinationLimitHit|.
```

index.bs
```diff
@@ -2619,10 +2664,10 @@ To <dfn>process an attribution source</dfn> given an [=attribution source=] |sou
             :: |source|'s [=attribution source/source time=]
             : [=attribution rate-limit record/expiry time=]
             :: null
-            : [=attribution rate-limit record/event-level report ID=]
+            : [=attribution rate-limit record/entity ID=]
             :: null
```
apasel422, line 2668: Is the `entity ID` field only `null` for fake event-level reports? If so, maybe add note that in the struct-field description for clarity.

params/chromium-params.md
```diff
@@ -16,7 +16,9 @@ Chromium's implementation assigns the following values:
 | [Max aggregatable reports per source][] | [20][max aggregatable reports per source value] |
 | [Max destinations covered by unexpired sources][] | [100][max destinations covered by unexpired sources value] |
 | [Destination rate-limit window][] | [1 minute][destination rate-limit window value]
-| [Max destinations per rate-limit window][] | [50][max destinations per rate-limit window per reporting site] per reporting site, [200][max destinations per rate-limit window total] total
+| [Max destinations per rate-limit window][] | [200][max destinations per rate-limit window value]
+| [Destination per reporting site rate-limit window][] | [1 day]
```
apasel422, line 20:
```suggestion
| [Destinations-per-reporting-site rate-limit window][] | [1 day]
```

params/chromium-params.md
```diff
@@ -45,8 +47,10 @@ Chromium's implementation assigns the following values:
...
 [Destination rate-limit window value]: https://source.chromium.org/chromium/chromium/src/+/refs/heads/main:content/browser/attribution_reporting/destination_throttler.h;l=30;drc=1890f3f74c8100eb1a3e945d34d6fd576d2a9061
 [Max destinations per rate-limit window]: https://wicg.github.io/attribution-reporting-api/#max-destinations-per-rate-limit-window
-[Max destinations per rate-limit window per reporting site]: https://source.chromium.org/chromium/chromium/src/+/refs/heads/main:content/browser/attribution_reporting/destination_throttler.h;l=29;drc=1890f3f74c8100eb1a3e945d34d6fd576d2a9061
-[Max destinations per rate-limit window total]: https://source.chromium.org/chromium/chromium/src/+/refs/heads/main:content/browser/attribution_reporting/destination_throttler.h;l=28;drc=1890f3f74c8100eb1a3e945d34d6fd576d2a9061
+[Max destinations per rate-limit window value]: https://source.chromium.org/chromium/chromium/src/+/refs/heads/main:content/browser/attribution_reporting/destination_throttler.h;l=28;drc=1890f3f74c8100eb1a3e945d34d6fd576d2a9061
+[Destination per reporting site rate-limit window]: https://wicg.github.io/attribution-reporting-api/#destination-per-site-rate-limit-window
```
apasel422, line 51:
```suggestion
[Destinations-per-reporting-site rate-limit window]: https://wicg.github.io/attribution-reporting-api/#destination-per-site-rate-limit-window
```

verbose_debugging_reports.md
```diff
@@ -113,10 +110,10 @@ each object has a string field `type` and a dictionary field `body`, e.g.:
 
 ```jsonc
 [{
-  "type": "source-destination-limit",
+  "type": "source-destination-rate-limit",
   "body": {
     "attribution_destination": "https://destination.example",
-    "limit": "100",
+    "limit": "50",
```
apasel422, line 116: This is just an example. It doesn't need to be changed.

verbose_debugging_reports.md
```diff
@@ -134,12 +131,15 @@ otherwise the dictionary may include the following fields:
 * `source_site`: The site on which source was registered, e.g. `"https://source.example"`.
 * `trigger_debug_key`: The debug key in the trigger registration, omitted if not set.
 
+If `type` is [`source-success`](#source-success) or [`source-noised`](#source-nosied), the `body`
```
apasel422, line 134:
```suggestion
If `type` is [`source-success`](#source-success) or [`source-noised`](#source-noised), the `body`
```

verbose_debugging_reports.md
```diff
@@ -134,12 +131,15 @@ otherwise the dictionary may include the following fields:
...
+dictionary may include `source_destination_limit` field if [destination limit](https://github.com/WICG/attribution-reporting-api/blob/main/EVENT.md#limiting-the-number-of-unique-destinations-covered-by-unexpired-sources)
```
apasel422, line 135:
```suggestion
dictionary may include a `source_destination_limit` field if the [destination limit](https://github.com/WICG/attribution-reporting-api/blob/main/EVENT.md#limiting-the-number-of-unique-destinations-covered-by-unexpired-sources)
```

verbose_debugging_reports.md
```diff
@@ -134,12 +131,15 @@ otherwise the dictionary may include the following fields:
...
+was exceeded.
+
 This table defines the fields in the `body` dictionary.
 
 | `type` | `attribution_destination`| `limit` | `source_debug_key` | `source_event_id` | `source_site` | `trigger_debug_key` |
```
apasel422, line 140: Shouldn't there be a column now for `source_destination_limit`?

apasel422 (changes requested):

EVENT.md
linnan-github: As we need to change the report types (`source-destination-limit` deprecated`), I think it's clearer to update the limit as well.

apasel422: But this is just an example. The value is meaningless.

verbose_debugging_reports.md
linnan-github: This field is only applicable to two fields, and is already noted separately.

apasel422: IMO it's harder to use the table if it's incomplete.

linnan-github: Updated, thanks.

index.bs
linnan-github: Added, thanks.

index.bs
```diff
@@ -2493,11 +2538,16 @@ To <dfn>obtain and deliver a debug report on source registration</dfn> given a
...
     <dl class="switch">
-    : "<code>[=source debug data type/source-destination-limit=]</code>"
-    :: [=map/Set=] |body|["`limit`"] to the user agent's [=max destinations covered by unexpired sources=],
-         [=serialize an integer|serialized=].
+    : "<code>[=source debug data type/source-success=]</code>"
+    : "<code>[=source debug data type/source-noised=]</code>"
+    :: If |destinationLimitHit| is true, [=map/set=] |body|["`source_destination_limit`"] to the user agent's [=max destinations covered by unexpired sources=],
+        [=serialize an integer|serialized=].
+
+    Note: The "`source_destiantion_limit`" field may be included to indicate that
```
apasel422, line 2546:
```suggestion
    Note: The "`source_destination_limit`" field may be included to indicate that
```

index.bs
```diff
@@ -2493,11 +2538,16 @@ To <dfn>obtain and deliver a debug report on source registration</dfn> given a
...
+    [=max destinations covered by unexpired sources=] was hit.
```
apasel422, line 2547: I think this note could use a bit more explanation. My immediate question on reading it is: "Why don't we just return a `source-destination-limit-hit` debug report instead of adding this field?" If it's to avoid leaking the exact error that occurred, we should say so, no?

apasel422 (changes requested):

index.bs
linnan-github: Elaborated a bit, PTAL, thanks.

apasel422 (approved):

vikassahu29 (approved):

index.bs
```diff
@@ -2410,45 +2408,91 @@ To <dfn>check if an [=attribution source=] exceeds the time-based destination li
...
-    : (true, false)
-    :: Return "<code>[=destination rate-limit result/hit global limit=]</code>".
-    : (true, true)
-    :: Return "<code>[=destination rate-limit result/hit reporting limit=]</code>".
+1. Return whether |destinations|'s [=set/size=] is greater than [=max destinations per reporting site per rate-limit window=].
+
+An [=attribution rate-limit record=] |a| <dfn for="attribution rate-limit record">is less than</dfn>
+an [=attribution rate-limit record=] |b| if the following steps return true:
+
+1. If |a|'s [=attribution rate-limit record/scope=] is less than |b|'s [=attribution rate-limit record/scope=],
```
giladbarkan-github, line 2433: The spec says
```
A rate-limit scope is one of the following:
"source"
"event-attribution"
"aggregatable-attribution"
```
Just to clarify -- is this comparing these? Why?

linnan-github: In the algorithm, the comparison is done only for `source` scope, so it's actually irrelevant, but just to make `less than` more complete.

index.bs
```diff
@@ -2410,45 +2408,91 @@ To <dfn>check if an [=attribution source=] exceeds the time-based destination li
...
 
-To <dfn>check if an [=attribution source=] exceeds the unexpired destination limit</dfn> given an
-[=attribution source=] |source|, run the following steps:
+        </dl>
+
+To <dfn>check if an [=attribution source=] exceeds the unexpired destination limit and maybe delete sources</dfn>
+given an [=attribution source=] |source|, run the following steps:
 
-1. Let |unexpiredSources| be all [=attribution rate-limit records=] |record| in the [=attribution rate-limit cache=] where all of the following are true:
+1. Let |unexpiredSources| be all [=attribution rate-limit records=] |record| in the [=attribution rate-limit cache=]
```
giladbarkan-github, line 2479: This seems confusing -- we're calling them "unexpired sources" but they are attribution rate-limit records.

index.bs
giladbarkan-github: If it's irrelevant, I'd argue not to include it.

index.bs
linnan-github: They are rate-limit records for sources, so unexpired sources seems reasonable to me.

apasel422:
Should this PR reference #1228?

index.bs
```diff
@@ -753,6 +753,18 @@ To <dfn>find a matching trigger spec</dfn> given an [=attribution source=]
...
 
+<h3 id="destination-limit-algorithm-header">Destination limit algorithm</h3>
+
+A <dfn>destination limit algorithm</dfn> is one of the following:
+
+<dl dfn-for="destination limit algorithm">
+: "<dfn><code>fifo</code></dfn>"
+:: Counts the highest prioritized and most recent [=attribution source/attribution destinations=].
+: "<dfn><code>lifo</code></dfn>"
+:: Counts the earliest [=attribution source/attribution destinations=].
```
apasel422, line 764: Are `fifo` and `lifo` opposites of each other in this scheme? If not, the naming is a bit misleading.

index.bs
```diff
@@ -2430,45 +2466,90 @@ To <dfn>check if an [=attribution source=] exceeds the time-based destination li
...
+1. If |sourcesToDelete| [=set/is empty=], return.
+1. [=set/iterate|For each=] [=attribution source=] |source| of the [=attribution source cache=]:
+    1. [=set/Remove=] |source| from the [=attribution source cache=] if |sourcesToDelete|
+        [=set/contains=] |source|'s [=attribution source/source identifier=].
+1. [=set/iterate|For each=] [=event-level report=] |report| of the [=event-level report cache=]:
+    1. If |report|'s [=event-level report/trigger time=] is greater than or equal to |now|,
+        [=set/remove=] |report| from the [=event-level report cache=].
+
+        Note: This would only delete [=obtain a fake report|fake reports=] that would never be
+        scheduled if they were true [=event-level reports=].
```
apasel422, line 2500: What does `scheduled` mean here? Do we mean `sent` instead?

index.bs
```diff
@@ -2516,8 +2597,19 @@ To <dfn>obtain and deliver a debug report on source registration</dfn> given a
...
     :: [=map/Set=] |body|["`limit`"] to the user agent's [=max destinations covered by unexpired sources=],
          [=serialize an integer|serialized=].
+    : "<code>[=source debug data type/source-success=]</code>"
+    : "<code>[=source debug data type/source-noised=]</code>"
+    :: If |destinationLimitHit| is true, [=map/set=] |body|["`source_destination_limit`"] to the user agent's [=max destinations covered by unexpired sources=],
+        [=serialize an integer|serialized=].
+
+    Note: The "`source_destination_limit`" field may be included to indicate that
+    [=max destinations covered by unexpired sources=] was hit and
+    [=destination limit algorithm/fifo=] algorithm was applied, which is not
```
apasel422, line 2607:
```suggestion
    the [=destination limit algorithm/fifo=] algorithm was applied, which is not
```

EVENT.md
```diff
@@ -1132,7 +1154,7 @@ effectively limits the number of unique sites covered per {source site, reportin
 To further reduce the possibility of a history reconstruction attack, the browser can also limit the number of `destination` sites registered per {source-site, 1 minute}.
 
 
-Additionally, to prevent one origin from using up the budget in the limit above, the browser can also limit the number of `destination` sites per {source site, reporting site, 1 minute}.
+Additionally, to prevent one origin from using up the budget in the limit above, the browser can also limit the number of `destination` sites per {source site, reporting site, 1 day}.
```
apasel422, line 1157: According to the PR description, this change is being made for better privacy with FIFO. Is there a reason we should keep it as is for LIFO?

apasel422 (changes requested):
Please clarify in the PR description whether this is backwards compatible if a registration entirely omits the new `destination_limit` source-registration field.

EVENT.md
```diff
@@ -1112,9 +1112,31 @@ trying to measure user visits on, the browser can limit the number `destination`
...
+
+The `priority` field is used to prioritize the destinations registered with this
+source with respect to other destinations.
+
+The `algorithm` field is used to select the algorithm to determine the final
+`destination` sites that the browser counts. When "lifo" is used, the browser
+will drop the new source if the source site is at its limit. When "fifo" is
+used, the browser will only count the highest priority and most recent
+`destination` sites within this limit, including those registered with the new
+source. The browser will delete pending sources and aggregatable reports
```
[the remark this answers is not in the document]
linnan-github: Yes, that's intentional. Event-level reports don't have to be deleted as the leak is mitigated by the fake reports.

index.bs
linnan-github: It's not strictly opposite, fifo uses priority but lifo is not.

index.bs
linnan-github: `scheduled` means scheduled but not actually sent yet. We only delete fake reports that haven't been delivered yet.

EVENT.md
linnan-github: The change is necessary for FIFO, and to avoid complexity, we will change this in either algorithm.

EVENT.md
```diff
@@ -1112,9 +1112,31 @@ trying to measure user visits on, the browser can limit the number `destination`
...
+  ..., // existing fields
+  "destination_limit": {
+    "algorithm": "lifo", // or "fifo", defaults to "lifo" if not present
+    "priority": "[64-bit signed integer]" // defaults to 0 if not present
+  }
+}
+```
+
+The `priority` field is used to prioritize the destinations registered with this
+source with respect to other destinations.
```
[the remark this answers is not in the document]
linnan-github: Updated, thanks.

EVENT.md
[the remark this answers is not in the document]
linnan-github: Updated to mention this in the explainer.

EVENT.md
```diff
@@ -1112,9 +1112,31 @@ trying to measure user visits on, the browser can limit the number `destination`
 sites represented by unexpired sources for a source-site.
 
 The browser can place a limit on the number of a source site's unexpired source's
-unique `destination` sites. When an attribution source is registered for a site
-that is not already in the unexpired sources and a source site is at its limit,
-the browser will drop the new source.
+unique `destination` sites. Source registration will accept an optional field
+`destination_limit` to allow developers to select the behavior when an attribution
+source is registered for a site that is not already in the unexpired sources and a
```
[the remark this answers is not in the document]
linnan-github: Updated to "not already registered. by any of the unexpired sources". Hopefully it's a bit easier to understand.

index.bs
apasel422: Maybe `never be stored` then?

index.bs
apasel422: I wonder if we should have some more precise names, then. IIUC, in `fifo`, sources are sorted by priority and *then* by recency, so maybe it should be `priority-fifo`?

And should it be an error to specify the `priority` field if the algorithm is `lifo`? At the very least we should probably warn about this in the header validator.

index.bs
linnan-github: SG, updated, thanks.

index.bs
linnan-github: Yeah renaming to priority-fifo SGTM. The priority can still be set if the algorithm is lifo, and will be stored as is for future fifo use.

EVENT.md
```diff
@@ -1112,9 +1112,33 @@ trying to measure user visits on, the browser can limit the number `destination`
...
+source with respect to other destinations for the purpose of source deactivation.
+
+The `algorithm` field is used to select the algorithm to determine the final
+`destination` sites that the browser counts. When "lifo" is used, the browser
+will drop the new source if the source site is at its limit. When "priority_fifo"
+is used, the browser will only count the highest prioritized and most recent
+`destination` sites within this limit, including those registered with the new
+source. The browser will delete pending sources and aggregatable reports
+associated with the `destination` sites that are no longer counted. The event-level
+reports are not deleted as the leak is mitigated by fake reports within
```
apasel422, line 1140: Let's clarify what leak this is.

EVENT.md
```diff
@@ -1112,9 +1112,33 @@ trying to measure user visits on, the browser can limit the number `destination`
...
+}
+```
+
+The `priority` field is used to prioritize the destinations registered with this
+source with respect to other destinations for the purpose of source deactivation.
+
+The `algorithm` field is used to select the algorithm to determine the final
+`destination` sites that the browser counts. When "lifo" is used, the browser
+will drop the new source if the source site is at its limit. When "priority_fifo"
+is used, the browser will only count the highest prioritized and most recent
```
apasel422, line 1136: `will only count` is a bit ambiguous: Does it mean that only the destination sites whose priority equals the max are counted?

In general I think these algorithm descriptions could be clarified a bit more, perhaps with examples.

index.bs
```diff
@@ -753,6 +753,18 @@ To <dfn>find a matching trigger spec</dfn> given an [=attribution source=]
 
     </dl>
 
+<h3 id="destination-limit-algorithm-header">Destination limit algorithm</h3>
+
+A <dfn>destination limit algorithm</dfn> is one of the following:
+
+<dl dfn-for="destination limit algorithm">
+: "<dfn><code>priority_fifo</code></dfn>"
+:: Counts the highest prioritized and most recent [=attribution source/attribution destinations=].
```
apasel422, line 762: Similar to my comment in `EVENT.md`, I find `counts...` to be hard to understand. IIUC all destinations are *counted*, but the algorithm and priority are used to determine which ones to evict.

index.bs
```diff
@@ -2430,45 +2466,90 @@ To <dfn>check if an [=attribution source=] exceeds the time-based destination li
...
+1. [=set/iterate|For each=] [=attribution rate-limit record=] |record| of the [=attribution rate-limit cache=]:
+    1. If |record|'s [=attribution rate-limit record/scope=] is not "<code>[=rate-limit scope/source=]</code>", [=iteration/continue=].
+    1. If |record|'s [=attribution rate-limit record/deactivated for unexpired destination limit=] is true, [=iteration/continue=].
+    1. If |record|'s [=attribution rate-limit record/source site=] and |source|'s [=attribution source/source site=] are not equal, [=iteration/continue=].
+    1. If |record|'s [=attribution rate-limit record/reporting origin=] and |source|'s [=attribution source/reporting origin=] are not [=same site=], [=iteration/continue=].
+    1. If |record|'s [=attribution rate-limit record/expiry time=] is less than or equal to |source|'s [=attribution source/source time=], [=iteration/continue=].
+    1. [=Assert=]: |record|'s [=attribution rate-limit record/destination limit priority=] is not null.
+    1. [=Assert=]: |record|'s [=attribution rate-limit record/entity ID=] is not null.
+    1. [=list/Append=] the [=tuple=] (|record|'s [=attribution rate-limit record/destination limit priority=],
```
apasel422, line 2529: IMO this tuple has enough fields that it would be better to use a struct with named fields.

index.bs
```diff
@@ -2430,45 +2466,90 @@ To <dfn>check if an [=attribution source=] exceeds the time-based destination li
...
+1. [=set/iterate|For each=] [=site=] |destination| of |source|'s [=attribution source/attribution destinations=]:
+    1. [=list/Append=] the [=tuple=] (|source|'s [=attribution source/destination limit priority=],
+        |source|'s [=attribution source/source time=], |destination|, |source|'s [=attribution source/source identifier=])
+        to |records|.
+1. [=list/sort in descending order|Sort=] |records| in descending order, with |a| less than |b| if the following steps return true:
+    1. If |a|[0] is less than |b|[0], return true.
+    1. If |a|[0] is greater than |b|[0], return false.
+    1. If |a|[1] is less than |b|[1], return true.
+    1. If |a|[1] is greater than |b|[1], return false.
+    1. If <a href="https://html.spec.whatwg.org/multipage/origin.html#serialization-of-a-site">serialized</a> |a|[2]
```
apasel422, line 2541: I was surprised to see that we need to explicitly serialize the site here in order to be able to compare it with another site, because I thought that our use of sets of sites elsewhere in the spec implied that sites already had an ordering operation.

But then I noticed that Infra sets (which are always "ordered sets") don't actually enforce any ordering beyond the ordering of their underlying list, based on [this example](https://infra.spec.whatwg.org/#example-set-replace). In other words, the Infra set and map types aren't, in a Java analogy, `TreeSet` or `TreeMap`, but rather `LinkedHashSet` or `LinkedHashMap`.

So my comment here isn't an issue with this PR, but I think we need to audit our use of sets and maps again. Related to #1193.

index.bs
```diff
@@ -2516,8 +2597,19 @@ To <dfn>obtain and deliver a debug report on source registration</dfn> given a
...
          [=serialize an integer|serialized=].
+    : "<code>[=source debug data type/source-success=]</code>"
+    : "<code>[=source debug data type/source-noised=]</code>"
+    :: If |destinationLimitHit| is true, [=map/set=] |body|["`source_destination_limit`"] to the user agent's [=max destinations covered by unexpired sources=],
+        [=serialize an integer|serialized=].
+
+    Note: The "`source_destination_limit`" field may be included to indicate that
+    [=max destinations covered by unexpired sources=] was hit and
+    the [=destination limit algorithm/priority_fifo=] algorithm was applied, which is not
+    reported as [=source debug data type/source-destination-limit=] to prevent side-channel
```
apasel422, line 2608:
```suggestion
    reported as "<code>[=source debug data type/source-destination-limit=]</code>" to prevent side-channel
```

index.bs
```diff
@@ -2574,23 +2666,29 @@ To <dfn>process an attribution source</dfn> given an [=attribution source=] |sou
...
-1. If |destinationRateLimitResult| is "<code>[=destination rate-limit result/hit reporting limit=]</code>":
+1. If the result of running [=check if an attribution source exceeds the time-based destination per reporting site limit=] with |source| is true:
     1. Run [=obtain and deliver a debug report on source registration=] with "<code>[=source debug data type/source-destination-rate-limit=]</code>" and |source|.
     1. Return.
+1. Let |sourcesToDeleteForDestinationLimit| be the result of running [=get sources to delete for the unexpired destination limit=]
+    with |source|.
+1. Let |destinationLimitHit| be false if |sourcesToDeleteForDestinationLimit| [=set/is empty=], true otherwise.
+1. Let |dropNewSource| be false.
+1. If |source|'s [=attribution source/destination limit algorithm=] is:
+    <dl class="switch">
```
apasel422, line 2677: This is missing a `</dl>`.

apasel422 (changes requested):

index.bs
linnan-github: Updated with a struct, thanks.

EVENT.md
linnan-github: Updated, thanks.

EVENT.md
linnan-github: Updated the description for clarity.

index.bs
```diff
@@ -2526,29 +2546,48 @@ To <dfn>get sources to delete for the unexpired destination limit</dfn> given an
...
+        :: |source|'s [=attribution source/source time=]
+        : [=destination limit record/source identifier=]
+        :: |record|'s [=attribution source/source identifier=]
+
+    1. [=list/Append=] |destinationRecord| to |destinationRecords|.
+1. [=list/sort in descending order|Sort=] |destinationRecords| in descending order, with |a| less than |b| if the following steps return true:
+    1. If |a|'s [=destination limit record/priority=] is less than |b|'s [=destination limit record/priority=], return true.
+    1. If |a|'s [=destination limit record/priority=] is greater than |b|'s [=destination limit record/priority=], return false.
+    1. If |a|'s [=destination limit record/time=] is less than |b|'s [=destination limit record/time=], return true.
+    1. If |a|'s [=destination limit record/time=]] is greater than |b|'s [=destination limit record/time=], return false.
```
apasel422, line 2578:
```suggestion
    1. If |a|'s [=destination limit record/time=] is greater than |b|'s [=destination limit record/time=], return false.
```

ts/src/header-validator/source.test.ts
```diff
@@ -1285,6 +1285,21 @@ const testCases: TestCase[] = [
       },
     ],
   },
+  {
+    name: 'destination-limit-lifo-non-default-priority',
+    json: `{
+      "destination": "https://a.test",
+      "destination_limit": {
+	"priority": "123"
```
apasel422, line 1293: Let's fix the indentation here.

ts/src/header-validator/source.test.ts
```diff
@@ -1285,6 +1285,21 @@ const testCases: TestCase[] = [
...
+      }
+    }`,
+    expectedWarnings: [
+      {
+        path: ['destination_limit', 'priority'],
+        msg: 'non-default value (123) for algorithm lifo',
```
apasel422, line 1299: To make sure I'm understanding this: A source might still want to specify a non-0 priority even if it's using `lifo` because a future source may use `priority-fifo` and then the first source's priority would be considered? If so, we may want to say something to that effect in this message and make this a note instead of a warning. Maybe something like:

```
... (value may still be used by future priority-fifo source)
```
?

ts/src/header-validator/validate-json.ts
```diff
@@ -1172,6 +1172,14 @@ function destinationLimit(ctx: Context, j: Json): Maybe<DestinationLimit> {
     ),
 
     ...priorityField,
+  }).peek((s) => {
+    ctx.scope('priority', () => {
+      if (s.algorithm == DestinationLimitAlgorithm.lifo && s.priority != 0n) {
```
apasel422, line 1177:
```suggestion
      if (s.algorithm === DestinationLimitAlgorithm.lifo && s.priority !== 0n) {
```

apasel422 (changes requested):

ts/src/header-validator/source.test.ts
linnan-github: yeah, that's correct, updated as suggested, thanks.

ts/src/header-validator/source.test.ts
linnan-github: The indentation seems to be right?

ts/src/header-validator/source.test.ts
apasel422: I still think it should be a note, not a warning, as it's not really something that the developer should take action on. And let's surround the `may still be used...` part with parentheses.

ts/src/header-validator/validate-json.ts
```diff
@@ -1154,6 +1154,35 @@ function warnInconsistentMaxEventLevelReportsAndTriggerSpecs(
...
+      (ctx, j) => enumerated(ctx, j, DestinationLimitAlgorithm),
+      DestinationLimitAlgorithm.lifo
+    ),
+
+    ...priorityField,
+  }).peek((s) => {
+    ctx.scope('priority', () => {
+      if (s.algorithm === DestinationLimitAlgorithm.lifo && s.priority !== 0n) {
+        ctx.warning(
+          `non-default value (${s.priority}) for algorithm ${s.algorithm} may still be used by future ${DestinationLimitAlgorithm.priority_fifo} source`
```
apasel422, line 1179:
```suggestion
          `non-default pririty (${s.priority}) for algorithm ${s.algorithm} may still be used by future ${DestinationLimitAlgorithm.priority_fifo} source`
```

apasel422 (approved):
LGTM from a spec/validator perspective. Please wait for approval from a utility perspective.

EVENT.md
```diff
@@ -1114,9 +1114,34 @@ trying to measure user visits on, the browser can limit the number `destination`
...
+
+The `priority` field is used to prioritize the destinations registered with this
+source with respect to other destinations for the purpose of source deactivation.
+
+The `algorithm` field is used to select the algorithm to determine the final
+`destination` sites that the browser selects. When "lifo" is used, the browser
+will drop the new source if the source site is at its limit. When "priority_fifo"
+is used, the browser will sort the `destination` sites registered by unexpired
+sources, including the new source, by `priority` in descending order and by
+the registration time in descending order. The browser will then select the
```
apasel422, line 1140: I'm a bit confused: The algorithm is called `priority_fifo` but the sources are ordered by registration time *descending*?

EVENT.md
```diff
@@ -1129,6 +1154,10 @@ the [denial of service](#denial-of-service) for more details. To prevent this
 attack, the browser should maintain these limits per reporting site. This
 effectively limits the number of unique sites covered per {source site, reporting site} applied to all unexpired sources regardless of type at source time.
 
+The browser can also limit the number of `destination` sites per {source site, reporting site, 1 day}
+to mitigate the history reconstruction attack as new `destination` sites can be
+registered when "priority_fifo" algorithm is used.
```
apasel422, line 1159:
```suggestion
registered when the `priority_fifo` algorithm is used.
```

verbose_debugging_reports.md
```diff
@@ -29,6 +29,9 @@ is rejected due to the following limits to mitigate security concerns:
 #### `source-destination-rate-limit`
 A source is rejected due to the [destinations per source and reporting site rate limit](https://github.com/WICG/attribution-reporting-api/blob/main/EVENT.md#limiting-the-number-of-unique-destinations-per-source-site).
 
+#### `source-destination-wider-rate-limit`
```
apasel422, line 32: IMO `wider` is not descriptive enough, and IIUC the value in question isn't even necessarily wider than any other limit, since these are vendor-specific values.

CC @johnivdel for thoughts on naming.

apasel422 (changes requested):

EVENT.md
linnan-github: fifo -> first in first out, so the earliest destinations will be out first, therefore the sources are ordered by registration time desending.

verbose_debugging_reports.md
linnan-github: Yeah I agree that `wider` may not be good, alternatively we can change this to per day and also make the new rate limit window constant.

EVENT.md
apasel422: Ah, this algorithm is saying which ones to keep, not delete. OK.

verbose_debugging_reports.md
linnan-github: Gave it a try, hopefully this is clearer.

EVENT.md
```diff
@@ -1114,9 +1114,34 @@ trying to measure user visits on, the browser can limit the number `destination`
...
+source with respect to other destinations for the purpose of source deactivation.
+
+The `algorithm` field is used to select the algorithm to determine the final
+`destination` sites that the browser selects. When "lifo" is used, the browser
+will drop the new source if the source site is at its limit. When "priority_fifo"
+is used, the browser will sort the `destination` sites registered by unexpired
+sources, including the new source, by `priority` in descending order and by
+the registration time in descending order. The browser will then select the
+first few `destination` sites within this limit, and delete pending sources and
+aggregatable reports associated with the unselected `destination` sites. The
```
akashnadan, line 1142: @linnan-github should this be "associated with the **selected** `destination` sites."

linnan-github: The selected destinations are the ones we keep, so we delete sources/reports for unselected destinations.

index.bs
```diff
@@ -2604,23 +2774,36 @@ To <dfn>process an attribution source</dfn> given an [=attribution source=] |sou
...
+    : "<code>[=destination limit algorithm/priority_fifo=]</code>"
+    :: Set |dropNewSource| to true if |sourcesToDeleteForDestinationLimit| [=set/contains=] |source|'s [=attribution source/source identifier=].
+    : "<code>[=destination limit algorithm/lifo=]</code>"
+    :: Set |dropNewSource| to |destinationLimitHit|.
+
+    </dl>
+1. If |dropNewSource| is true:
+    1. Run [=obtain and deliver a debug report on source registration=] with "[=source debug data type/source-destination-limit=]</code>" and |source|.
+    1. Return.
+1. Run [=delete sources for unexpired destination limit=] with |sourcesToDeleteForDestinationLimit| and |source|'s [=attribution source/source time=].
```
vikassahu29, line 2800: Do we need to a conditional deletion here based on the `dropNewSource` flag?

It seems that the newest source won't be deleted in case of LIFO, or am I missing something?

Let's say the limit is 1 and there are 2 sources: oldSource{destination: d1} & newSource {destination: d2}.

- Step 4 in `get sources to delete for the unexpired destination limit` the sorted list will look like [d2, d1].
- Step 6 in  get sources to delete for the unexpired destination limit` will add oldSource to `sourcesToDelete` instead of newSource (in case of LIFO)

linnan-github: If `|dropNewSource|` is true, it will early return. Therefore this is only reachable when `|dropNewSource|` is false, that is, we only delete sources if the new source is not to be dropped by this limit.

linnan-github: And in case of FIFO, dropNewSource will be true as long as the returning set is non empty, so new source will be deleted.

vikassahu29: Ack. I missed the early return.

vikassahu29 (approved):

EVENT.md
```diff
@@ -1114,9 +1114,34 @@ trying to measure user visits on, the browser can limit the number `destination`
...
+    "priority": "[64-bit signed integer]" // defaults to 0 if not present
+  }
+}
+```
+
+The `priority` field is used to prioritize the destinations registered with this
+source with respect to other destinations for the purpose of source deactivation.
+
+The `algorithm` field is used to select the algorithm to determine the final
+`destination` sites that the browser selects. When "lifo" is used, the browser
```
johnivdel, line 1136: I think a naming like "ignore"/"replace" is more clear than lifo/priority_fifo. Where lifo ignores sources once the limit is hit, and the fifo algorithm will replace them.

index.bs
```diff
@@ -1347,6 +1379,11 @@ The second controls the maximum number of distinct [=sites=] across all [=attrib
 for [=attribution sources=] with a given ([=attribution source/source site=], [=attribution source/reporting origin=] [=site=])
 per [=destination rate-limit window=].
 
+<dfn>Max destinations per reporting site per day</dfn> is an integer controls
```
johnivdel, line 1382: Should this be "per source reporting site"? Otherwise this seems to be across source sites by naming.

index.bs
```diff
@@ -1292,6 +1321,9 @@ controls the maximum [=map/size=] of a [=trigger spec map=] for an
 the maximum [=string/length=] of an [=attribution trigger=]'s [=attribution trigger/trigger context ID=].
 Its value is 64.
 
+<dfn>Destination per day rate-limit window</dfn> is a positive [=duration=] that
```
johnivdel, line 1324: This is a bit strange given the name of the limit already encodes day.

One option is to just not make this paramterizable, or potentially consider a different naming although I agree it's hard to come up with a name.

One potential idea would be to instead include this in [=Max destinations per rate-limit window=] by extending the tuple to include a third integer, and ensuring a different rate limit window for the new value. wdyt?

linnan-github: I would lean towards not making this parametrizable. Including this in the existing tuple may be a bit confusing as they have different rate limit windows.

EVENT.md
linnan-github: I don’t feel strongly either way. @akashnadan @vikassahu29 Any thoughts/preferences for the naming? Thanks.

index.bs
linnan-github: Updated to "per source reporting site", thanks.

index.bs
linnan-github: Not needed anymore.

EVENT.md
```diff
@@ -1114,9 +1114,27 @@ trying to measure user visits on, the browser can limit the number `destination`
...
+```
+
+When an attribution source is registered for a site that is not already in the
+unexpired sources and a source site is at its limit, the browser will sort the
+`destination` sites registered by unexpired sources, including the new source,
+by `destination_limit_priority` in descending order and by the registration
+time in descending order. The browser will then select the first few
+`destination` sites within this limit, and delete pending sources and
+aggregatable reports associated with the unselected `destination` sites. The
+event-level reports are not deleted as the leak of user's browsing history is
```
apasel422, line 1136:
```suggestion
event-level reports are not deleted, as the leak of user's browsing history is
```

EVENT.md
```diff
@@ -1114,9 +1114,27 @@ trying to measure user visits on, the browser can limit the number `destination`
...
+}
+```
+
+When an attribution source is registered for a site that is not already in the
+unexpired sources and a source site is at its limit, the browser will sort the
+`destination` sites registered by unexpired sources, including the new source,
+by `destination_limit_priority` in descending order and by the registration
+time in descending order. The browser will then select the first few
+`destination` sites within this limit, and delete pending sources and
+aggregatable reports associated with the unselected `destination` sites. The
```
apasel422, line 1135:
```suggestion
aggregatable reports associated with the unselected `destination` sites. Any
```

EVENT.md
```diff
@@ -1129,6 +1147,10 @@ the [denial of service](#denial-of-service) for more details. To prevent this
 attack, the browser should maintain these limits per reporting site. This
 effectively limits the number of unique sites covered per {source site, reporting site} applied to all unexpired sources regardless of type at source time.
 
+The browser can also limit the number of `destination` sites per {source site, reporting site, 1 day}
+to mitigate the history reconstruction attack as new `destination` sites can be
+registered when the `priority_fifo` algorithm is used.
```
apasel422, line 1152: This mention of an algorithm is obsolete IIUC.

index.bs
```diff
@@ -1094,6 +1102,7 @@ Possible values are:
 <li>"<dfn><code>source-destination-global-rate-limit</code></dfn>"
 <li>"<dfn><code>source-destination-limit</code></dfn>"
 <li>"<dfn><code>source-destination-rate-limit</code></dfn>"
+<li>"<dfn><code>source-destination-per-day-rate-limit</code></dfn>"
```
apasel422, line 1105:
```suggestion
<li>"<dfn><code>source-destination-per-day-rate-limit</code></dfn>"
<li>"<dfn><code>source-destination-rate-limit</code></dfn>"
```

index.bs
```diff
@@ -1380,6 +1389,11 @@ The second controls the maximum number of distinct [=sites=] across all [=attrib
 for [=attribution sources=] with a given ([=attribution source/source site=], [=attribution source/reporting origin=] [=site=])
 per [=destination rate-limit window=].
 
+<dfn>Max destinations per source reporting site per day</dfn> is an integer controls
```
apasel422, line 1392:
```suggestion
<dfn>Max destinations per source reporting site per day</dfn> is an integer that controls
```

index.bs
```diff
@@ -2643,19 +2664,119 @@ To <dfn>check if an [=attribution source=] exceeds the time-based destination li
...
+        |source|'s [=attribution source/reporting origin=]
      * |record|'s [=attribution rate-limit record/expiry time=] is greater than |source|'s [=attribution source/source time=]
-1. Let |unexpiredDestinations| be a new [=set=].
-1. For each [=attribution rate-limit record=] |unexpiredRecord| of |unexpiredSources|:
-    1. [=set/Append=] |unexpiredRecord|'s [=attribution rate-limit record/attribution destination=] to |unexpiredDestinations|.
-1. Let |newDestinations| be the result of taking the [=set/union=] of |unexpiredDestinations| and |source|'s [=attribution source/attribution destinations=].
-1. Return whether |newDestinations|'s [=set/size=] is greater than the user agent's [=max destinations covered by unexpired sources=].
+     * The [=duration from=] |record|'s [=attribution rate-limit record/time=] and |source|'s [=attribution source/source time=]
+        is less than 1 day
+1. Let |destinations| be the [=set=] of every [=attribution rate-limit record/attribution destination=] in |matchingSources|,
```
apasel422, line 2678:
```suggestion
1. Let |destinations| be the [=set=] of all [=attribution rate-limit record/attribution destination=] in |matchingSources|,
```

index.bs
```diff
@@ -2643,19 +2664,119 @@ To <dfn>check if an [=attribution source=] exceeds the time-based destination li
...
+    [=set/union|unioned=] with |source|'s [=attribution source/attribution destinations=].
+1. Return whether |destination|'s [=set/size=] is greater than [=max destinations per source reporting site per day=].
```
apasel422, line 2680:
```suggestion
1. Return whether |destinations|'s [=set/size=] is greater than [=max destinations per source reporting site per day=].
```

index.bs
```diff
@@ -2643,19 +2664,119 @@ To <dfn>check if an [=attribution source=] exceeds the time-based destination li
...
+    1. If |a|'s [=destination limit record/time=] is greater than |b|'s [=destination limit record/time=], return false.
+    1. If |a|'s <a href="https://html.spec.whatwg.org/multipage/origin.html#serialization-of-a-site">serialized</a>
+        [=destination limit record/attribution destination=]
+        is less than |b|'s <a href="https://html.spec.whatwg.org/multipage/origin.html#serialization-of-a-site">serialized</a>
+        [=destination limit record/attribution destination=], return true.
+    1. Return false.
+1. Let |sourcesToDelete| be a new [=set=].
+1. [=set/iterate|For each=] [=destination limit record=] |record| of |destinationRecords|:
+    1. Let |destination| be |record|'s [=destination limit record/attribution destination=].
+    1. If |newDestinations|'s [=set/size=] is less than the user agent's [=max destinations covered by unexpired sources=],
```
apasel422, line 2775: Where is `newDestinations` declared?

apasel422 (changes requested):

EVENT.md
linnan-github: Removed, thanks.

index.bs
linnan-github: Hmm this may be accidentally removed in a merge, thanks for catching.

EVENT.md
linnan-github: Not needed, resolving.

apasel422 (approved):

index.bs
```diff
@@ -2909,39 +3065,40 @@ a [=boolean=] |isNoised|:
...
     |source|'s [=attribution source/source site=], |source|'s [=attribution source/reporting origin=],
     |source|, |source|'s [=attribution source/attribution destinations=][0],
     |config|'s [=aggregatable debug reporting config/aggregation coordinator=],
     and |source|'s [=attribution source/source time=].
 
 To <dfn>obtain and deliver debug reports on source registration</dfn>
-given a [=source debug data type=] |dataType|, an [=attribution source=] |source|,
-and an optional [=boolean=] |isNoised| (default false):
+given a [=set=] of [=source debug data types=] |dataTypes|, an [=attribution source=] |source|,
+, and an optional [=boolean=] |isNoised| (default false):
```
apasel422, line 3096:
```suggestion
and an optional [=boolean=] |isNoised| (default false):
```

apasel422 (approved):

PR #1220, opened 2024-03-28
Add note to spec about check orders in source registration for security concerns

linnan-github:
Add non-normative note about the security implications with the check order.

***
<a href="https://pr-preview.s3.amazonaws.com/linnan-github/conversion-measurement-api/pull/1220.html" title="Last updated on Mar 29, 2024, 12:47 PM UTC (b6647ef)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/attribution-reporting-api/1220/c0e269d...linnan-github:b6647ef.html" title="Last updated on Mar 29, 2024, 12:47 PM UTC (b6647ef)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2538,7 +2538,7 @@ privacy budget of all possible destinations.
 
 Note: The limits that are not reported as <code>[=source debug data type/source-success=]</code> should be
 checked before any limits that are reported as <code>[=source debug data type/source-success=]</code>
-to prevent side-channel attacks. 
+to prevent side-channel leakage of cross-origin data.
 
 # Triggering Algorithms # {#trigger-algorithms}
 
--- a/index.bs
+++ b/index.bs
@@ -2536,6 +2536,10 @@ To <dfn>process an attribution source</dfn> given an [=attribution source=] |sou
 Note: Because a fake report does not have a "real" effective destination, we need to subtract from the
 privacy budget of all possible destinations.
 
+Note: The limits that are reported as explicit [=source debug data type=] should
+be checked before any limits that are reported implicitly as <code>[=source debug data type/source-success=]</code>
+to prevent side-channel attacks.
+
 # Triggering Algorithms # {#trigger-algorithms}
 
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -2536,6 +2536,10 @@ To <dfn>process an attribution source</dfn> given an [=attribution source=] |sou
 Note: Because a fake report does not have a "real" effective destination, we need to subtract from the
 privacy budget of all possible destinations.
 
+Note: The limits that are reported as explicit [=source debug data type=] should
+be checked before any limits that are reported implicitly as <code>[=source debug data type/source-success=]</code>
+to prevent side-channel attacks.
```
apasel422, line 2541: I'm not sure what `explicit` and `implicit` mean here. Could we clarify that a bit?

apasel422 (changes requested):

index.bs
linnan-github: Implicit -> reported as source-success, and explicit should be opposite. I think it's clear?

apasel422: Instead of explicit and implicit, can we can `source-success` and not `source-success`? 

It might also be useful simply to list all of the verbose debug report types in the order they should be reported in for privacy purposes.

giladbarkan-github: How is a limit reported "implicitly," and from a privacy/security perspective why should we care how limit is reported if it is in fact reported?

linnan-github: Updated as suggested, PTAL, thanks.

index.bs
```diff
@@ -2536,6 +2536,10 @@ To <dfn>process an attribution source</dfn> given an [=attribution source=] |sou
 Note: Because a fake report does not have a "real" effective destination, we need to subtract from the
 privacy budget of all possible destinations.
 
+Note: The limits that are not reported as <code>[=source debug data type/source-success=]</code> should be
+checked before any limits that are reported as <code>[=source debug data type/source-success=]</code>
+to prevent side-channel attacks.
```
apasel422, line 2541: Maybe something like

```suggestion
to prevent side-channel leakage of cross-origin data.
```
?

PR #1221, opened 2024-03-29
Determine debug data regardless of security concerning checks

linnan-github:
To better mitigate security concerns with security concerning checks, e.g. max source reporting origins per rate-limit window and max destinations per reporting window, the debug data sent in the report should not change depend on the result of those checks.

This PR fixes the issue with sources with randomized response, and `source-noised` may be sent if those security concerning checks fail.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2478,8 +2478,11 @@ To <dfn>process an attribution source</dfn> given an [=attribution source=] |sou
 1. If |destinationRateLimitResult| is "<code>[=destination rate-limit result/hit reporting limit=]</code>":
     1. Run [=obtain and deliver a debug report on source registration=] with "<code>[=source debug data type/source-destination-rate-limit=]</code>" and |source|.
     1. Return.
+1. Let |debugDataType| be "<code>[=source debug data type/source-success=]</code>".
+1. Set |debugDataType| to "<code>[=source debug data type/source-noised=]</code>"
+    if |source|'s [=attribution source/randomized response=] is not null.
 1. If |destinationRateLimitResult| is "<code>[=destination rate-limit result/hit global limit=]</code>":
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -2478,8 +2478,11 @@ To <dfn>process an attribution source</dfn> given an [=attribution source=] |sou
 1. If |destinationRateLimitResult| is "<code>[=destination rate-limit result/hit reporting limit=]</code>":
     1. Run [=obtain and deliver a debug report on source registration=] with "<code>[=source debug data type/source-destination-rate-limit=]</code>" and |source|.
     1. Return.
+1. Let |debugDataType| be "<code>[=source debug data type/source-success=]</code>".
+1. Set |debugDataType| to "<code>[=source debug data type/source-noised=]</code>"
+    if |source|'s [=attribution source/randomized response=] is not null.
```
giladbarkan-github, line 2483: Did you mean, "is null," rather than "is not null?"

linnan-github: The source is noised if the response "is not null" right?

giladbarkan-github: Right. I thought the intention if "hit global limit" is to send a source-noised debug report even if the source was not noised.

linnan-github: The intention is actually a "source-noised" debug report would be sent no matter whether the global limit is hit or not. It should be just based on the response config.

linnan-github: Yes, that's the intention. The same debug report should be sent no matter global limit hit or not.

giladbarkan-github: Gotcha

giladbarkan-github (approved):

johnivdel (approved):
LGTM

PR #1245, opened 2024-04-16
Add missing Makefile deps

apasel422:
The existing wildcard didn't properly account for files in nested directories.


johnivdel (approved):
lgtm

PR #1246, opened 2024-04-16
Add TypeScript implementation of filtering

apasel422:
This change adds an interactive form like the header validator to make it easier to experiment with this logic, which is non-trivial.

filters

### DIFF
--- a/ts/src/header-validator/filters-index.ts
+++ b/ts/src/header-validator/filters-index.ts
@@ -1,0 +1,112 @@
+import { SourceType } from '../source-type'
+import { Context, Issue, PathComponent } from './context'
+import * as filters from './filters'
+import { validateJSON, filterPair, filterData } from './validate-json'
+
+const form = document.querySelector('form')! as HTMLFormElement
+const sourceAgeInput = form.elements.namedItem(
+  'source-age'
+)! as HTMLInputElement
+const filterDataInput = form.elements.namedItem(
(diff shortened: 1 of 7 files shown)


linnan-github:
Could you please add a screenshot, thanks!

ts/src/header-validator/filters-index.ts
```diff
@@ -0,0 +1,112 @@
...
+  const sourceAge = sourceAgeInput.valueAsNumber
+  if (!(sourceAge >= minSourceAge && sourceAge <= maxSourceAge)) {
+    sourceErrorList.append(
+      makeLi({
+        msg: `source age must be in the range [${minSourceAge}, ${maxSourceAge}]`,
+      })
+    )
+  }
+
+  const [filterDataResult, parsedFilterData] = validateJSON(
```
linnan-github, line 77: It's probably duplicate and irrelevant of the filtering logic, but wdyt about also validating the filters value here, not just valid JSON?

apasel422: It is validated. Any errors will be shown in the UI.

linnan-github: Thanks, yeah missed that the validation function is passed in.

apasel422:
> Could you please add a screenshot, thanks!

Done.

linnan-github (approved):

PR #1247, opened 2024-04-17
Parse OS items in preparation for showing effective value

apasel422:
In a followup change, we will show the effective value of the header (i.e. with invalid values removed and defaults applied) in the interactive form.


apasel422:
> LGTM, but of course please wait for John's approval

Probably not necessary.

PR #1250, opened 2024-04-17
Show effective values in header validator

apasel422:
This allows users to see how omitted values are defaulted, and how adjusted values are clamped.

### DIFF
--- a/ts/src/header-validator/to-json.ts
+++ b/ts/src/header-validator/to-json.ts
@@ -1,0 +1,304 @@
+import * as parsed from './validate-json'
+
+type MaybeHasField<K extends string, V> = {
+  [key in K]?: V
+}
+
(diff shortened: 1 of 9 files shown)


ThomasQuesadilla (approved):
High Level LGTM

ts/src/header-validator/to-json.ts
```diff
@@ -0,0 +1,304 @@
...
+  f: (u: U) => V
+): MaybeHasField<K, V> {
+  const obj: MaybeHasField<K, V> = {}
+  if (u !== null) {
+    obj[key] = f(u)
+  }
+  return obj
+}
+
+export type CommonDebug = {
```
linnan-github, line 19: Overall LGTM. One comment is that these types are very similar to those defined in validate-json.ts, and it would be nice if they could be reused in some way.

PR #1268, opened 2024-05-03
Skip event-level-reports-per-destination check for replaced reports

### LINKED ISSUE #1267: "max event-level reports per attribution destination" check may be wrong with respect to report replacement
[This check](https://wicg.github.io/attribution-reporting-api/#ref-for-max-event-level-reports-per-attribution-destination%E2%91%A0) happens before we determine whether the new trigger will cause an existing report to be replaced, in which case the number of reports per attribution destination may remain the same but the trigger to be rejected.

apasel422:
Without this, when an event-level report is being replaced, the limit ends up being `n - 1`.

Fixes #1267

***
<a href="https://pr-preview.s3.amazonaws.com/apasel422/attribution-reporting-api/pull/1268.html" title="Last updated on May 3, 2024, 5:58 PM UTC (4299151)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/attribution-reporting-api/1268/1a9478a...apasel422:4299151.html" title="Last updated on May 3, 2024, 5:58 PM UTC (4299151)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -3259,13 +3259,6 @@ To <dfn>trigger event-level attribution</dfn> given an [=attribution trigger=] |
         |trigger|, |sourceToAttribute| and [=obtain debug data on trigger registration/report=] set to null.
     1. Return the [=triggering result=] ("<code>[=triggering status/dropped=]</code>", |debugData|).
 1. [=Assert=]: |windowResult| is <strong>falls within</strong>.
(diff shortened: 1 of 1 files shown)


johnivdel (approved):
lgtm

linnan-github (approved):

PR #1285, opened 2024-05-16
Refactor trigger verbose debug report generation in spec

linnan-github:
Moves the debug data generation to `triggering attribution` algorithm for simplification.

***
<a href="https://pr-preview.s3.amazonaws.com/linnan-github/conversion-measurement-api/pull/1285.html" title="Last updated on May 17, 2024, 6:14 PM UTC (4f5d359)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/attribution-reporting-api/1285/4565ada...linnan-github:4f5d359.html" title="Last updated on May 17, 2024, 6:14 PM UTC (4f5d359)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1185,13 +1185,23 @@ A <dfn>triggering status</dfn> is one of the following:
 Note: "<code>[=triggering status/noised=]</code>" only applies for [=triggering event-level attribution=] when it is attributed
 successfully but dropped as the noise was applied to the source.
 
+A <dfn>trigger debug data</dfn> is [=tuple=] with the following items:
+
+<dl dfn-for="trigger debug data">
+: <dfn>data type</dfn>
+:: A [=trigger debug data type=].
+: <dfn>report</dfn>
+:: Null or an [=attribution report=].
+
+</dl>
+
 A triggering result is a [=tuple=] with the following items:
 
 <dl dfn-for="triggering result">
 : <dfn>status</dfn>
 :: A [=triggering status=].
 : <dfn>debug data</dfn>
-:: Null or an [=attribution debug data=].
+:: Null or a [=trigger debug data=].
 
 </dl>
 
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -3448,15 +3423,19 @@ To <dfn>trigger aggregatable attribution</dfn> given an [=attribution trigger=]
 
 <h3 algorithm id="triggering-attribution">Triggering attribution</h3>
 
-To <dfn>obtain and deliver a debug report on trigger registration</dfn> given a [=trigger debug data type=] |dataType|,
-an [=attribution trigger=] |trigger| and an optional [=attribution source=]
+To <dfn>obtain and deliver a debug report on trigger registration</dfn>
+given a [=set=] of [=trigger debug data=] |dataSet|, an [=attribution trigger=] |trigger|,
```
apasel422, line 3427: What do you think about making this a list instead of a set? I ask because it's not clear to me how the equality operation is supposed to work for `trigger debug data`, particularly with respect to its `report` field.

index.bs
```diff
@@ -3448,15 +3423,19 @@ To <dfn>trigger aggregatable attribution</dfn> given an [=attribution trigger=]
...
-1. Let |debugData| be the result of running [=obtain debug data on trigger registration=]
-    with |dataType|, |trigger|, |sourceToAttribute| and
-    [=obtain debug data on trigger registration/report=] set to null.
-1. If |debugData| is null, return.
-1. Run [=obtain and deliver a debug report=] with « |debugData| », |trigger|'s [=attribution trigger/reporting origin=],
+1. Let |debugDataList| be a new [=list=].
+1. [=set/iterate|For each=] |data| of |dataSet|:
+    1. Let |debugData| be the result of running [=obtain debug data on trigger registration=] with
+        |data|'s [=trigger debug data/data type=], |trigger|, |sourceToAttribute|,
+        and |data|'s [=trigger debug data/report=].
```
apasel422, line 3435: Instead of passing the data type and report as separate arguments, should we just pass `data` itself?

index.bs
```diff
@@ -3489,8 +3468,9 @@ To <dfn noexport>trigger attribution</dfn> given an [=attribution trigger=] |tri
     [=set/is empty=] and |hasAggregatableData| is false, return.
 1. Let |matchingSources| be the result of running [=find matching sources=] with |trigger|.
 1. If |matchingSources| [=list/is empty=]:
-    1. Run [=obtain and deliver a debug report on trigger registration=]
-        with "<code>[=trigger debug data type/trigger-no-matching-source=]</code>", |trigger| and [=obtain and deliver a debug report on trigger registration/sourceToAttribute=] set to null.
+    1. Run [=obtain and deliver a debug report on trigger registration=] with
+        « ("<code>[=trigger debug data type/trigger-no-matching-source=]</code>", null) »,
+        |trigger| and [=obtain and deliver a debug report on trigger registration/sourceToAttribute=] set to null.
```
apasel422, line 3473:
```suggestion
        |trigger|, and [=obtain and deliver a debug report on trigger registration/sourceToAttribute=] set to null.
```

apasel422 (changes requested):

index.bs
linnan-github: The comparison is field-by-field, including the report field, which is mostly null except for two event errors. So basically it’s just comparing types.

index.bs
linnan-github: Passing the data type and report is more flexible if data generation is different.

index.bs
apasel422: Is there actually a benefit to using a set though? As written, it seems like implementations will need to compare every single field in the report in this algorithm.

linnan-github: As event-level and aggregate may have the same debug type, so using a set will deduplicate them and only results in one payload. Alternately, we can pass as a list and deduplicate based on types in the report generation algorithm. Wdyt?

apasel422: Could you clarify when they can have the same debug type? I thought that the common debug types were only generated before calling the aggregatable-specific and event-level-specific algorithms.

apasel422: I think it would be clearer if we refactored [that algorithm](https://pr-preview.s3.amazonaws.com/linnan-github/conversion-measurement-api/pull/1285.html#check-if-attribution-should-be-blocked-by-rate-limits) to perform [should processing be blocked by reporting-origin limit](https://pr-preview.s3.amazonaws.com/linnan-github/conversion-measurement-api/pull/1285.html#should-block-processing-for-reporting-origin-limit) only once per attribution, instead of once for event-level and once for aggregatable, since IIUC the result should be the same.

If that's not possible because the checks are done in a particular order for privacy reasons, then I would expect that algorithm to have a note making that clear.

linnan-github: Added an issue to considering refactoring to make this PR no behavioral change.

apasel422: Could you also add an issue for changing the type from `set` to `list`?

apasel422 (changes requested):
LGTM once the issue is added.

index.bs
linnan-github: Added an issue, thanks.

apasel422 (approved):

PR #1290, opened 2024-05-20
Spec Flexible Contributions Filtering

agarant:
Fixes: https://github.com/WICG/attribution-reporting-api/issues/1291

***
<a href="https://pr-preview.s3.amazonaws.com/agarant/attribution-reporting-api/pull/1290.html" title="Last updated on Jun 14, 2024, 7:02 PM UTC (6b60132)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/attribution-reporting-api/1290/bf3f812...agarant:6b60132.html" title="Last updated on Jun 14, 2024, 7:02 PM UTC (6b60132)">Diff</a>


AGGREGATE.md
```diff
@@ -507,6 +507,46 @@ A similar design was proposed for the
...
+
+```jsonc
+{
+  ..., // existing fields
+  "aggregatable_filtering_id_max_bytes": 2, // defaults to 1
+  "aggregatable_values": {
+    "campaignCounts": 32768,
+    "geoValue": {
+      "value": 1664,
+      "filtering_id": "23", // must fit within <aggregatable_filtering_id_max_bytes> bytes
```
apasel422, line 523:
```suggestion
      "filtering_id": "23", // must fit within <aggregatable_filtering_id_max_bytes> bytes
    }
```

index.bs
```diff
@@ -942,6 +952,8 @@ An attribution trigger is a [=struct=] with the following items:
 :: Null or a [=string=].
 : <dfn>fenced</dfn>
 :: A [=boolean=].
+: <dfn>aggregatable filtering ID max bytes </dfn>
```
apasel422, line 955:
```suggestion
: <dfn>aggregatable filtering ID max bytes</dfn>
```

index.bs
```diff
@@ -2771,24 +2797,64 @@ To <dfn>parse aggregatable trigger data</dfn> given a [=map=] |map|:
     1. [=list/Append=] |aggregatableTrigger| to |aggregatableTriggerData|.
 1. Return |aggregatableTriggerData|.
 
-To <dfn>parse aggregatable key-values</dfn> given a [=map=] |map|:
+To <dfn>parse aggregatable filtering id max bytes</dfn> given a [=map=] |map|:
```
apasel422, line 2800:
```suggestion
To <dfn>parse aggregatable filtering ID max bytes</dfn> given a [=map=] |map|:
```

index.bs
```diff
@@ -2771,24 +2797,64 @@ To <dfn>parse aggregatable trigger data</dfn> given a [=map=] |map|:
...
 
+1. Let |maxBytes| be [=default filtering ID max bytes=].
+1. If |map|["`aggregatable_filtering_id_max_bytes`"] [=map/exists=]:
+    1. Let |maxBytesValue| be |map|["`aggregatable_filtering_id_max_bytes`"].
+    1. If |maxBytesValue| is a non-negative integer and is [=set/contained=] in the [=valid filtering ID max bytes range=].
+        Let |maxBytes| be |maxBytesValue|.
```
apasel422, line 2806:
```suggestion
    1. If |maxBytesValue| is a non-negative integer and is [=set/contained=] in the [=valid filtering ID max bytes range=],
        set |maxBytes| to |maxBytesValue|.
```

index.bs
```diff
@@ -2771,24 +2797,64 @@ To <dfn>parse aggregatable trigger data</dfn> given a [=map=] |map|:
...
+    1. Otherwise, return null.
+1. Return |maxBytes|.
+
+To <dfn>parse aggregatable key-values value</dfn> given a |value|:
```
apasel422, line 2810: This algorithm might be more simply expressed as a validity-checker, i.e. something returning true/false, since `value` itself is returned unchanged.

index.bs
```diff
@@ -2771,24 +2797,64 @@ To <dfn>parse aggregatable trigger data</dfn> given a [=map=] |map|:
...
+    1. If |value| is an integer:
+        1. Let |val| be the result of running [=parse aggregatable key-values value=] with |value|.
+        1. If |val| is null, return null.
+        1. [=map/Set=] |out|[|key|] to a new [=aggregatable key value=] whose items are
+            : [=aggregatable key value/value=]
+            :: |val|
+            : [=aggregatable key value/filtering id=]
+            :: [=default filtering ID value=].
+        1. [=iteration/Continue=].
+    1. If |value|["`value`"] does not [=map/exists=], return null.
```
apasel422, line 2832:
```suggestion
    1. If |value|["`value`"] does not [=map/exist=], return null.
```

index.bs
```diff
@@ -2771,24 +2797,64 @@ To <dfn>parse aggregatable trigger data</dfn> given a [=map=] |map|:
...
+    1. Otherwise, return null.
+1. Return |maxBytes|.
+
+To <dfn>parse aggregatable key-values value</dfn> given a |value|:
+1. If |value| is not an integer, return null.
+1. If |value| is less than or equal to 0, return null.
+1. If |value| is greater than [=allowed aggregatable budget per source=], return null.
+1. Return |value|.
+
+To <dfn>parse aggregatable key-values</dfn> given a [=map=] |map| and integer |maxBytes|:
```
apasel422, line 2816:
```suggestion
To <dfn>parse aggregatable key-values</dfn> given a [=map=] |map| and an integer |maxBytes|:
```

index.bs
```diff
@@ -2771,24 +2797,64 @@ To <dfn>parse aggregatable trigger data</dfn> given a [=map=] |map|:
...
-To <dfn>parse aggregatable values</dfn> given a [=map=] |map|:
+    1. If |value| is not a [=map=] or an integer, return null.
+    1. If |value| is an integer:
+        1. Let |val| be the result of running [=parse aggregatable key-values value=] with |value|.
+        1. If |val| is null, return null.
+        1. [=map/Set=] |out|[|key|] to a new [=aggregatable key value=] whose items are
+            : [=aggregatable key value/value=]
+            :: |val|
+            : [=aggregatable key value/filtering id=]
+            :: [=default filtering ID value=].
```
apasel422, line 2830:
```suggestion
            :: [=default filtering ID value=]
```

index.bs
```diff
@@ -2771,24 +2797,64 @@ To <dfn>parse aggregatable trigger data</dfn> given a [=map=] |map|:
...
+        1. [=iteration/Continue=].
+    1. If |value|["`value`"] does not [=map/exists=], return null.
+    1. Let |val| be the result of running
+        [=parse aggregatable key-values value=] with |value|["`value`"]|.
+    1. If |val| is null, return null.    
+    1. Let |filteringId| be [=default filtering ID value=].
+    1. If |value|["`filtering_id`"] [=map/exists=]:
+        1. Let |filteringId| be the result of applying the
```
apasel422, line 2838:
```suggestion
        1. Set |filteringId| to the result of applying the
```

index.bs
```diff
@@ -2771,24 +2797,64 @@ To <dfn>parse aggregatable trigger data</dfn> given a [=map=] |map|:
...
+            <a spec="html">rules for parsing non-negative integers</a> to |value|[|"`filtering_id`"|].
```
apasel422, line 2839:
```suggestion
            <a spec="html">rules for parsing non-negative integers</a> to |value|["`filtering_id`"].
```

index.bs
```diff
@@ -2771,24 +2797,64 @@ To <dfn>parse aggregatable trigger data</dfn> given a [=map=] |map|:
...
+        1. If |filteringId| is an error, return null
```
apasel422, line 2840:
```suggestion
        1. If |filteringId| is an error, return null.
```

index.bs
```diff
@@ -2771,24 +2797,64 @@ To <dfn>parse aggregatable trigger data</dfn> given a [=map=] |map|:
...
+        1. If |filteringId| is not in in [=the exclusive range|the range=]
+            0 to 256<sup>|maxBytes|</sup>, exclusive, return null.    
+    1.  [=map/Set=] |out|[|key|] to a new [=aggregatable key value=] whose items are
+        : [=aggregatable key value/value=]
+        :: |val|
+        : [=aggregatable key value/filtering id=]
+        :: |filteringId|
+1. Return |out|.
+
+To <dfn>parse aggregatable values</dfn> given a [=map=] |map| and integer |maxBytes|:
```
apasel422, line 2850:
```suggestion
To <dfn>parse aggregatable values</dfn> given a [=map=] |map| and an integer |maxBytes|:
```

index.bs
apasel422, line 2816: Should we require `maxBytes` to be positive?

index.bs
```diff
@@ -2994,6 +3066,15 @@ To <dfn>match an attribution source against filters and negated filters</dfn> gi
     |source|, |notFilters|, |moment|, and [=match an attribution source against filters/isNegated=] set to true is false, return false.
 1. Return true.
 
+<h3 dfn id="should-send-a-report-unconditionally">Should send a report unconditionally</h3>
+
+To <dfn>check if an aggregatable report be unconditionally sent</dfn> given an [=attribution trigger=] |trigger|:
```
apasel422, line 3071:
```suggestion
To <dfn>check if an aggregatable report should be unconditionally sent</dfn> given an [=attribution trigger=] |trigger|:
```

index.bs
```diff
@@ -3755,8 +3841,10 @@ To <dfn>queue reports for delivery</dfn> given a [=set=] of
 
 <h3 id="encode-integer">Encode an unsigned k-bit integer</h3>
 
-To <dfn>encode an unsigned k-bit integer</dfn>, represent it as a big-endian [=byte sequence=]
-of length k / 8, left padding with zero as necessary.
+To <dfn>encode an unsigned k-bytes integer</dfn> given an integer |integerToEncode|
```
apasel422, line 3844:
```suggestion
To <dfn>encode an unsigned k-byte integer</dfn> given an integer |integerToEncode|
```

apasel422 (changes requested):
We should also update the header validator for this change.

ts/src/header-validator/validate-json.ts
```diff
@@ -1279,28 +1279,54 @@ function aggregatableTriggerData(
   )
 }
 
+export type AggregatableValuesValue =
```
apasel422, line 1282: To match [the spec](https://pr-preview.s3.amazonaws.com/agarant/attribution-reporting-api/pull/1290.html#aggregatable-key-value-filtering-id), I think this should remove the post-parsing distinction between ones with a `filtering_id` and ones without.

ts/src/header-validator/validate-json.ts
```diff
@@ -1279,28 +1279,54 @@ function aggregatableTriggerData(
...
-    .filter((n) =>
-      isInRange(ctx, n, 1, constants.allowedAggregatableBudgetPerSource)
-    )
+
+  return typeSwitch(ctx, j, {
+    number: (ctx, j) => aggregatableKeyValueValue(ctx, j),
+    object: (ctx, j) =>
+      struct(ctx, j, {
+        value: field('value', aggregatableKeyValueValue),
+        filteringId: field('filtering_id', uint64, null),
```
apasel422, line 1314: Shouldn't this default to `0n`?

apasel422, line 1314: Here we are using `uint64`, but the spec currently just says:

> Set filteringId to the result of applying the [rules for parsing non-negative integers](https://html.spec.whatwg.org/multipage/common-microsyntaxes.html#rules-for-parsing-non-negative-integers) to value["filtering_id"].

Maybe the spec should use a sized integer type?

apasel422 (changes requested):

ts/src/header-validator/to-json.ts
```diff
@@ -263,7 +261,10 @@ function serializeAggregatableValuesConfiguration(
   for (const [key, value] of c.values.entries()) {
     values[key] =
       typeof value == 'number'
-        ? value
+        ? {
+            value,
+            filtering_id: '0',
+          }
```
apasel422: IIUC this branch is impossible now.

apasel422 (changes requested):

index.bs
```diff
@@ -2799,24 +2825,63 @@ To <dfn>parse aggregatable trigger data</dfn> given a [=map=] |map|:
...
+    1. Let |filteringId| be [=default filtering ID value=].
+    1. If |value|["`filtering_id`"] [=map/exists=]:
+        1. Set |filteringId| to the result of applying the
+            <a spec="html">rules for parsing non-negative integers</a> to |value|["`filtering_id`"].
+        1. If |filteringId| is an error, return null.
+        1. If |filteringId| is not in in [=the exclusive range|the range=]
+            0 to 256<sup>|maxBytes|</sup>, exclusive, return null.    
+    1.  [=map/Set=] |out|[|key|] to a new [=aggregatable key value=] whose items are
+        : [=aggregatable key value/value=]
+        :: |val|
```
linnan-github, line 2872:
```suggestion
        :: |value|["`value`"]
```

index.bs
```diff
@@ -2799,24 +2825,63 @@ To <dfn>parse aggregatable trigger data</dfn> given a [=map=] |map|:
...
+        1. Set |val| to |value|.
+        1. [=map/Set=] |out|[|key|] to a new [=aggregatable key value=] whose items are
+            : [=aggregatable key value/value=]
+            :: |val|
+            : [=aggregatable key value/filtering id=]
+            :: [=default filtering ID value=]
+        1. [=iteration/Continue=].
+    1. If |value|["`value`"] does not [=map/exist=], return null.
+    1. If the result of running [=validate aggregatable key-values value=] with |value|["`value`"] is false, return null.
+    1. Set |val| to |value|["`value`"].
```
linnan-github, line 2862: |var| is unnecessary.

index.bs
```diff
@@ -2883,7 +2949,10 @@ a [=moment=] |triggerTime|, and a [=boolean=] |fenced|:
 1. Let |aggregatableTriggerData| be the result of running [=parse aggregatable trigger data=]
     with |value|.
 1. If |aggregatableTriggerData| is null, return null.
-1. Let |aggregatableValuesConfigurations| be the result of running [=parse aggregatable values=] with |value|.
+1. Let |filteringIdsMaxBytes| be the result of [=parsing aggregatable filtering ID max bytes=] with |value|.
+1. If |filteringIdsMaxBytes| is null, return null.
```
linnan-github, line 2953: Shall we validate against source registration time configuration to make sure that it’s not configured for to include if max bytes is non-default?

index.bs
```diff
@@ -1003,6 +1015,8 @@ An aggregatable contribution is a [=struct=] with the following items:
 :: A non-negative 128-bit integer.
 : <dfn>value</dfn>
 :: A non-negative 32-bit integer.
+: <dfn>filtering ID</dfn> (default: [=default filtering ID value=])
```
linnan-github, line 1018: Is the default value used? Similar for the default in other structs.

ts/src/header-validator/validate-json.ts
agarant: 64 bit is the maximum size for a filtering id (assuming that max bytes is set to 8). The actual limit is validated later in `areFilteringIdsWithinMaxBytes` when the max bytes property has been parsed.

In the spec, the next line validates the actual limit. I think it's clearer to validate the size directly on the actual limit.

For the validator, changing the structure to pass the limit did not seem like it'd worth the effort. In practice, the worst case is that they'd hit this limit but the actual limit is even lower (due to having a lower configured max bytes). The optimal solution being that we check on the actual limit immediately

index.bs
agarant: might as well, updated to `a positive integer |maxBytes|`.

ts/src/header-validator/aggregatable-contributions.ts
```diff
@@ -24,7 +24,10 @@ function createAggregatableContributionsFromKeysAndValues(
     if (value === undefined) {
       continue
     }
-    contributions.push({ key, value })
+    contributions.push({
+      key,
+      value: typeof value === 'number' ? value : value.value,
```
apasel422, line 29: IIUC the `true` condition can never be reached as `value` is always an object now.

ts/src/header-validator/validate-json.ts
```diff
@@ -1389,6 +1422,52 @@ function aggregatableSourceRegistrationTime(
...
+      'must be in the range [0, 255]. It exceeds the default max size of 1 byte. To increase, specify the aggregatable_filtering_id_max_bytes property.'
+  }
+  ctx.scope('aggregatable_values', () => {
+    for (const [
+      index,
+      configurations,
+    ] of t.aggregatableValuesConfigurations.entries()) {
+      const evaluate = () => {
+        for (const [key, value] of configurations.values) {
+          if (typeof value === 'number') continue
```
apasel422, line 1449: IIUC this branch can never be taken now that `value` is always an object.

index.bs
```diff
@@ -2693,6 +2719,7 @@ A <dfn>trigger-registration JSON key</dfn> is one of the following:
 <li>"<dfn><code>aggregatable_deduplication_keys</code></dfn>"
 <li>"<dfn><code>aggregatable_source_registration_time</code></dfn>"
 <li>"<dfn><code>aggregatable_trigger_data</code></dfn>"
+<li>"<dfn><code>aggregatable_filtering_id_max_bytes</code></dfn>"
```
apasel422, line 2722: Let's keep this list sorted alphabetically.

apasel422 (changes requested):

ts/src/header-validator/validate-json.ts
apasel422: Ack.

index.bs
agarant: I suppose that the default values are always set from the constants directly. As such, they can be removed. I removed them.

apasel422 (approved):

ts/src/header-validator/validate-json.ts
agarant: After reviewing, `areFilteringIdsWithinMaxBytes` also had some complexities and was fragile given that it had to "replicate" the scope. As such, I updated to pass down the maxBytes value making the validator implementation closer to the spec.

agarant:
Thanks for the comments, updated!

index.bs
```diff
@@ -2799,24 +2826,61 @@ To <dfn>parse aggregatable trigger data</dfn> given a [=map=] |map|:
...
+    1. If |value| is not a [=map=] or an integer, return null.
+    1. If |value| is an integer:
+        1. If the result of running [=validate aggregatable key-values value=] with |value| is false, return null.
+        1. [=map/Set=] |out|[|key|] to a new [=aggregatable key value=] whose items are
+            : [=aggregatable key value/value=]
+            :: |value|
+            : [=aggregatable key value/filtering ID=]
+            :: [=default filtering ID value=]
+        1. [=iteration/Continue=].
+    1. If |value|["`value`"] does not [=map/exist=], return null.
```
linnan-github, line 2860:
```suggestion
    1. If |value|["<code>[=trigger-registration JSON key/value=]</code>"] does not [=map/exist=], return null.
```

index.bs
```diff
@@ -2799,24 +2826,61 @@ To <dfn>parse aggregatable trigger data</dfn> given a [=map=] |map|:
...
+    1. If the result of running [=validate aggregatable key-values value=] with |value|["`value`"] is false, return null.
+    1. Let |filteringId| be [=default filtering ID value=].
+    1. If |value|["`filtering_id`"] [=map/exists=]:
```
linnan-github, line 2863: Please add `filtering_id` to `trigger-registration JSON key` as well.

thanatkron (commented):
Thank

index.bs
```diff
@@ -2799,24 +2827,61 @@ To <dfn>parse aggregatable trigger data</dfn> given a [=map=] |map|:
...
+    1. If |value| is an integer:
+        1. If the result of running [=validate aggregatable key-values value=] with |value| is false, return null.
+        1. [=map/Set=] |out|[|key|] to a new [=aggregatable key value=] whose items are
+            : [=aggregatable key value/value=]
+            :: |value|
+            : [=aggregatable key value/filtering ID=]
+            :: [=default filtering ID value=]
+        1. [=iteration/Continue=].
+    1. If |value|["<code>[=trigger-registration JSON key/value=]</code>"] does not [=map/exist=], return null.
+    1. If the result of running [=validate aggregatable key-values value=] with |value|["`value`"] is false, return null.
```
linnan-github, line 2862:
```suggestion
    1. If the result of running [=validate aggregatable key-values value=] with |value|["<code>[=trigger-registration JSON key/value=]</code>"] is false, return null.
```

index.bs
```diff
@@ -2917,8 +2986,10 @@ a [=moment=] |triggerTime|, and a [=boolean=] |fenced|:
     1. If |value|["<code>[=trigger-registration JSON key/trigger_context_id=]</code>"] is not a [=string=], return null.
     1. If |value|["<code>[=trigger-registration JSON key/trigger_context_id=]</code>"]'s [=string/length=] is greater than the [=max length per trigger context ID=],
         return null.
-    1. If |aggregatableSourceRegTimeConfig| is not "<code>[=aggregatable source registration time configuration/exclude=]</code>", return null.
     1. Set |triggerContextID| to |value|["<code>[=trigger-registration JSON key/trigger_context_id=]</code>"].
+1. If |aggregatableSourceRegTimeConfig| is not "<code>[=aggregatable source registration time configuration/exclude=]</code>":
+    1. If |triggerContextID| is not null, return null.
```
linnan-github, line 2991: Do you think it makes sense to reuse “check if an aggregatable report should be unconditionally sent” here?

ts/src/header-validator/validate-json.ts
```diff
@@ -1270,52 +1273,128 @@ function aggregatableTriggerData(
...
+        )
+        return false
+      }
+      if (
+        aggregatableSourceRegTime.value !==
+          AggregatableSourceRegistrationTime.exclude &&
+        n !== constants.defaultAggregatableFilteringIdMaxBytes
+      ) {
+        ctx.error(
+          `with a non-default value (higher than 1) is prohibited for aggregatable_source_registration_time ${aggregatableSourceRegTime.value}`
```
linnan-github, line 1389: Shall we reference the constant here?

akashnadan (approved):

AGGREGATE.md
```diff
@@ -507,6 +509,47 @@ A similar design was proposed for the
...
+
+The filtering IDs need to be unsigned integers limited to a small number of
+bytes, (1 byte = 8 bits) by default. We limit the size of the ID space to
+prevent unnecessarily increasing the payload size and thus storage and
+processing costs.
+
+This size can be increased via the `aggregatable_filtering_id_max_bytes` field.
+To avoid amplifying a counting attack due to the resulting different payload
+size, the browser will unconditionally send an aggregatable report on every
+trigger registration with a non default (higher than 1) max bytes. A null report
```
apasel422, line 545:
```suggestion
trigger registration with a non-default (greater than 1) max bytes. A null report
```

AGGREGATE.md
```diff
@@ -507,6 +509,47 @@ A similar design was proposed for the
...
+will be sent in the case that the trigger registration did not generate an
+attribution report. The source registration time will always be excluded from
+the aggregatable report with a non default max bytes. This behavior is the same
```
apasel422, line 548:
```suggestion
the aggregatable report with a non-default max bytes. This behavior is the same
```

ts/src/header-validator/trigger.test.ts
```diff
@@ -101,12 +102,16 @@ const testCases: jsontest.TestCase<Trigger>[] = [
...
-          sourceKeys: new Set(['x']),
+          sourceKeys: new Set(['x', 'y']),
         },
       ],
+      aggregatableFilteringIdMaxBytes: 1,
       aggregatableValuesConfigurations: [
         {
-          values: new Map([['x', 5]]),
+          values: new Map([
+            ['x', { value: 5, filteringId: 0n }] as any,
```
apasel422, line 112: Is the `as any` necessary here?

ts/src/header-validator/validate-json.ts
```diff
@@ -1259,40 +1265,107 @@ function aggregatableValue(ctx: Context, j: Json): Maybe<number> {
...
+            return isInRange(
+              ctx,
+              n,
+              0n,
+              256n ** BigInt(maxBytes.value) - 1n,
+              maxBytes.value == constants.defaultAggregatableFilteringIdMaxBytes
+                ? 'must be in the range [0, 255]. It exceeds the default max size of 1 byte. To increase, specify the aggregatable_filtering_id_max_bytes property.'
+                : undefined
+            )
+          })
```
apasel422, line 1301: I think the logic here is complex enough that it merits being put in its own top-level function.

index.bs
```diff
@@ -1361,6 +1377,16 @@ controls the maximum [=map/size=] of a [=trigger spec map=] for an
 the maximum [=string/length=] of an [=attribution trigger=]'s [=attribution trigger/trigger context ID=].
 Its value is 64.
 
+<dfn>Default filtering ID value</dfn> is a non-negative integer which allows flexible
+contribution filtering of aggregatable reports. Its value is 0.
+
+<dfn>Default filtering ID max bytes</dfn> is a positive integer controlling the
+max bytes used if none is explicitly chosen. Its value is 1.
```
apasel422, line 1384: Maybe we can say what it's used for.

apasel422 (changes requested):

apasel422 (approved):

index.bs
```diff
@@ -4287,15 +4383,23 @@ is the result of running the following steps:
...
 
         : "`bucket`"
-        :: |contribution|'s [=aggregatable contribution/key=], [=encode an unsigned k-bit integer|encoded=]
+        :: The result of [=encoding an unsigned k-byte integer=] given
+            |contribution|'s [=aggregatable contribution/key=] and 16.
         : "`value`"
-        :: |contribution|'s [=aggregatable contribution/value=], [=encode an unsigned k-bit integer|encoded=]
+        :: The result of [=encoding an unsigned k-byte integer=] given
+            |contribution|'s [=aggregatable contribution/value=] and 4.
+        : "`id`"
```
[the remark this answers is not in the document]
agarant: Updated, to 1.0. From an implementation point of view, when using the new version it expects the id to be set.

We could have the debug reports use the old version in the spec and create a different contribution structure.

But I think that it adds more complexity to the spec without really bringing value.  wdyt ?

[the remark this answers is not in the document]
linnan-github: Then do we want to update the ADR report to use version "1.0" in the implementaion?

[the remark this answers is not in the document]
agarant: I think that eventually yes, we should be updating the ADR report to use version "1.0" if only to ensure that all implementations relying on the aggregation service are updated to the latest version.

ts/src/header-validator/validate-json.ts
```diff
@@ -1156,54 +1157,137 @@ function aggregatableTriggerData(
...
 }
 
 function aggregatableKeyValues(
   j: Json,
-  ctx: Context
+  ctx: Context,
+  maxBytes: Maybe<number>
 ): Maybe<AggregatableValues> {
-  return keyValues(j, ctx, aggregatableKeyValue)
+  return keyValues(j, ctx, (j, ctx) => aggregatableKeyValue(j, ctx, maxBytes))
```
apasel422, line 1231:
```suggestion
  return keyValues(j, ctx, (j) => aggregatableKeyValue(j, ctx, maxBytes))
```

ts/src/header-validator/validate-json.ts
```diff
@@ -1156,54 +1157,137 @@ function aggregatableTriggerData(
...
     object: (j) =>
-      aggregatableKeyValues(j, ctx).map((values) => [
+      aggregatableKeyValues(j, ctx, maxBytes).map((values) => [
         { values, positive: [], negative: [] },
       ]),
     list: (j) =>
       array(j, ctx, (j) =>
         struct(j, ctx, {
-          values: field('values', aggregatableKeyValues),
+          values: field('values', (j, ctx) =>
```
apasel422, line 1247:
```suggestion
          values: field('values', (j) =>
```

apasel422 (approved):

apasel422 (approved):

index.bs
```diff
@@ -1373,6 +1389,18 @@ controls the maximum [=map/size=] of a [=trigger spec map=] for an
 the maximum [=string/length=] of an [=attribution trigger=]'s [=attribution trigger/trigger context ID=].
 Its value is 64.
 
+<dfn>Default filtering ID value</dfn> is a non-negative integer. Its value is 0.
+It is the default value for flexible contribution filtering of aggregatable
```
linnan-github, line 1393: [=aggregatable reports=]

index.bs
```diff
@@ -3358,24 +3392,62 @@ To <dfn>parse aggregatable trigger data</dfn> given a [=map=] |map|:
     1. [=list/Append=] |aggregatableTrigger| to |aggregatableTriggerData|.
 1. Return |aggregatableTriggerData|.
 
-To <dfn>parse aggregatable key-values</dfn> given a [=map=] |map|:
+To <dfn>parse aggregatable filtering ID max bytes</dfn> given a [=map=] |map|:
+
+1. Let |maxBytes| be [=default filtering ID max bytes=].
+1. If |map|["<code>[=trigger-registration JSON key/aggregatable_filtering_id_max_bytes=]</code>"] [=map/exists=]:
+    1. Let |maxBytes| be |map|["<code>[=trigger-registration JSON key/aggregatable_filtering_id_max_bytes=]</code>"].
```
linnan-github, line 3399:
```suggestion
    1.Set |maxBytes| to |map|["<code>[=trigger-registration JSON key/aggregatable_filtering_id_max_bytes=]</code>"].
```

index.bs
```diff
@@ -4211,6 +4304,8 @@ an [=attribution trigger=] |trigger|:
     :: |trigger|'s [=attribution trigger/aggregatable source registration time configuration=].
     : [=aggregatable attribution report/trigger context ID=]
     :: |trigger|'s [=attribution trigger/trigger context ID=]
+    : [=aggregatable attribution report/filtering ID max bytes=]
+    :: |trigger|'s [=attribution trigger/aggregatable filtering ID max bytes=]
```
linnan-github, line 4308: Should we update this for null reports as well?

agarant: good catch, updated

linnan-github (approved):

PR #1289, opened 2024-05-20
Explainer for Aggregate Debug Reporting

linnan-github:
Related to #705

PR #1292, opened 2024-05-20
Spec header parsing for aggregate debug reporting

linnan-github:
https://github.com/WICG/attribution-reporting-api/pull/1289


index.bs
```diff
@@ -1552,6 +1574,114 @@ To <dfn>parse a filter pair</dfn> given a [=map=] |map|:
...
+positive integer |maxValue|, a [=boolean=] |isSource|, and an [=aggregate debug reporting config=]
+|default|:
+
+1. If |map|["<code>[=aggregate-debug-reporting JSON key/key_piece=]</code>"] does not
+    [=map/exist=], return |default|.
+1. If |map|["<code>[=aggregate-debug-reporting JSON key/key_piece=]</code>"] is not a
+    [=string=], return |default|.
+1. Let |keyPiece| be the result of running [=parse an aggregation key piece=]
+    with |map|["<code>[=aggregate-debug-reporting JSON key/key_piece=]</code>"].
+1. If |keyPiece| is an error, return |default|.
```
apasel422, line 1662: Are we intentionally ignoring the errors in this algorithm? If so, could we add a non-normative note explaining why?

index.bs
```diff
@@ -760,6 +760,21 @@ To <dfn>find a matching trigger spec</dfn> given an [=attribution source=]
...
+An aggregate debug reporting config is a [=struct=] with the following items:
+
+<dl dfn-for="aggregate debug reporting config">
+: <dfn>key piece</dfn> (default 0)
+:: A non-negative 128-bit integer.
+: <dfn>debug data</dfn> (default an [=map/is empty|empty=] [=map=])
+:: A [=map=] whose [=map/keys] are [=debug data types=] and whose [=map/values=]
+    are [=aggregatable contributions=].
+: <dfn>aggregation coordinator</dfn> (default [=default aggregation coordinator=])
+:: An [=aggregation coordinator=].
```
apasel422, line 774: Are these defaults used anywhere? If not, let's remove them.

index.bs
```diff
@@ -1552,6 +1574,114 @@ To <dfn>parse a filter pair</dfn> given a [=map=] |map|:
...
+1. If |map|["<code>[=aggregate-debug-reporting JSON key/debug_data=]</code>"] [=map/exists=]:
+    1. Set |debugDataMap| to the result of running [=parse aggregate debug reporting data=]
+        with |map|["<code>[=aggregate-debug-reporting JSON key/debug_data=]</code>"], |maxValue|,
+        and |isSource|.
+    1. If |debugDataMap| is null, return |default|.
+1. Let |aggregationCoordinator| be [=default aggregation coordinator=].
+1. If |map|["<code>[=aggregate-debug-reporting JSON key/aggregation_coordinator_origin=]</code>"]
+    [=map/exists=]:
+    1. Let |aggregationCoordinatorOrigin| be the result of [=parsing aggregation coordinator=].
+    1. If |aggregationCoordinatorOrigin| is null, return |default|.
```
apasel422, line 1673: Don't we have to update `aggregationCoordinator`?

index.bs
```diff
@@ -2337,6 +2469,21 @@ To <dfn>parse trigger specs</dfn> given a [=map=] |map|, a [=moment=]
 Issue: Invoke [=parse summary buckets=] and [=parse summary window operator=]
 from this algorithm.
 
+To <dfn>parse source aggregate debug reporting config</dfn> given |value|, an
```
apasel422, line 2472:
```suggestion
To <dfn>parse source aggregate debug reporting config</dfn> given |value|, a
```

index.bs
```diff
@@ -2703,6 +2865,7 @@ A <dfn>trigger-registration JSON key</dfn> is one of the following:
 <li>"<dfn><code>aggregatable_source_registration_time</code></dfn>"
 <li>"<dfn><code>aggregatable_trigger_data</code></dfn>"
 <li>"<dfn><code>aggregatable_values</code></dfn>"
+<li>"<dfn><code>aggregate_debug_reporting</code></dfn>"
```
apasel422, line 2868: Are we intentionally using `aggregate` here even though most of these other fields use `aggregatable`?

ts/src/header-validator/validate-json.ts
```diff
@@ -725,6 +725,102 @@ const priorityField: StructFields<Priority> = {
   priority: field('priority', int64, 0n),
 }
 
+function isDebugTypeSupported(ctx: RegistrationContext, s: string) {
+  if (ctx instanceof SourceContext) {
+    return constants.sourceAggregateDebugTypes.includes(s)
+  } else {
+    return constants.triggerAggregateDebugTypes.includes(s)
+  }
```
apasel422, line 733: Instead of doing an `instanceof` check here, can we change the context type to itself have a field for the debug types?

ts/src/header-validator/validate-json.ts
```diff
@@ -725,6 +725,102 @@ const priorityField: StructFields<Priority> = {
...
+  aggregationCoordinatorOrigin: string
+  data: AggregateDebugReportingData[]
+}
+
+const aggregateDebugReportingConfig: StructFields<
+  AggregateDebugReportingConfig,
+  RegistrationContext
+> = {
+  aggregationCoordinatorOrigin: aggregationCoordinatorOriginField,
+  data: field('data', aggregateDebugReportingDataList, []),
```
apasel422, line 819: Isn't this field called `debug_data` in the spec?

ts/src/header-validator/source.test.ts
```diff
@@ -1357,6 +1379,487 @@ const testCases: TestCase[] = [
...
+    ],
+  },
+  {
+    name: 'aggregate-debug-reporting-data-elem-empty',
+    json: `{
+      "destination": "https://a.test",
+      "aggregate_debug_reporting": {
+        "budget": 123,
+        "key_piece": "0x1",
+	"data": [{}]
```
apasel422, line 1584: Please fix the indentation here and throughout this file.

apasel422 (changes requested):

ts/src/header-validator/validate-json.ts
linnan-github: Yes, also realized this and updated. Thanks.

index.bs
linnan-github: Updated, thanks.

index.bs
linnan-github: Added note.

index.bs
linnan-github: The defaults are defined so that when creating a new item we don’t have to specify each item explicitly. This is used in the source and trigger JSON parsing.

index.bs
linnan-github: “aggregate_debug_reporting” seems to read better than “aggregatable_debug_reporting”. I slightly lean towards keeping as is, but don’t feel too strongly. Cc @akashnadan

index.bs
```diff
@@ -1552,6 +1574,116 @@ To <dfn>parse a filter pair</dfn> given a [=map=] |map|:
...
+
+1. If |value| is not a [=string=], return null.
+1. Let |url| be the result of running the [=URL parser=] on |value|.
+1. If |url| is failure or null, return null.
+1. If |url|'s [=url/origin=] is not an [=aggregation coordinator=], return null.
+1. Return |url|'s [=url/origin=].
+
+<h3 id="parsing-aggregate-debug-reporting-config">Parsing aggregate debug reporting config</h3>
+
+An <dfn>aggregate-debug-reporting JSON key<dfn> is one of the following:
```
apasel422, line 1589:
```suggestion
An <dfn>aggregate-debug-reporting JSON key</dfn> is one of the following:
```

index.bs
```diff
@@ -2405,6 +2554,17 @@ To <dfn noexport>parse source-registration JSON</dfn> given a [=byte sequence=]
...
+1. Let |aggregateDebugReportingConfig| be a new [=aggregate debug reporting config=].
+1. If |value|["<code>[=source-registration JSON key/aggregate_debug_reporting=]</code>"] [=map/exists=]:
+    1. Let |aggregateDebugReportingResult| be the result of running
+        [=parse source aggregate debug reporting config=] with
+        |value|["<code>[=source-registration JSON key/aggregate_debug_reporting=]</code>"],
+        |aggregatableDebugBudget|, and |aggregateDebugReportingConfig|.
+    1. Set |aggregatableDebugBudget| to |aggregateDebugReportingResult|[0].
+    1. Set |aggregateDebugReportingConfig| to |aggregateDebugReportingResult|[1].
+1. Let |aggregatableAttributionBudget| be [=allowed aggregatable budget per source=]
+    - |aggregatableDebugBudget|.
```
apasel422, line 2567:
```suggestion
1. Let |aggregatableAttributionBudget| be [=allowed aggregatable budget per source=] - |aggregatableDebugBudget|.
```

Otherwise it is rendered as a nested list item.

index.bs
apasel422: To me the consistency is more important.

apasel422 (dismissed):

index.bs
linnan-github: That's fair, updated to use `aggregatable_debug_reporting`.

ts/src/constants.ts
```diff
@@ -50,3 +50,41 @@ export const defaultTriggerDataCardinality: Readonly<
...
+export const triggerAggregatableDebugTypes: Readonly<[string, ...string[]]> = [
+  'trigger-aggregate-attributions-per-source-destination-limit',
+  'trigger-aggregate-deduplicated',
+  'trigger-aggregate-excessive-reports',
+  'trigger-aggregate-no-contributions',
+  'trigger-aggregate-insufficient-budget',
+  'trigger-aggregate-storage-limit',
+  'trigger-aggregate-report-window-passed',
+  'trigger-event-attributions-per-source-destination-limit',
+  'trigger-event-dedupliated',
```
apasel422, line 76:
```suggestion
  'trigger-event-deduplicated',
```

ts/src/constants.ts
```diff
@@ -50,3 +50,41 @@ export const defaultTriggerDataCardinality: Readonly<
...
+  'trigger-event-excessive-reports',
+  'trigger-event-low-prioirty',
```
apasel422, line 78:
```suggestion
  'trigger-event-low-priority',
```

ts/src/header-validator/source.test.ts
```diff
@@ -39,7 +39,16 @@ const testCases: TestCase[] = [
...
       "priority": "2",
       "source_event_id": "3",
-      "max_event_level_reports": 2
+      "max_event_level_reports": 2,
+      "aggregatable_debug_reporting": {
+        "budget": 1234,
+        "key_piece": "0x2",
+        "debug_data": [ {
+          "key_piece": "0x1",
+          "types": ["source-success"],
```
apasel422, line 48: Let's also add an unknown debug data type here to ensure that it roundtrips properly.

ts/src/header-validator/validate-json.ts
```diff
@@ -725,6 +726,101 @@ const priorityField: StructFields<Priority> = {
   priority: field('priority', int64, 0n),
 }
 
+function isDebugTypeSupported(ctx: RegistrationContext, s: string) {
```
apasel422, line 729:
```suggestion
function isDebugTypeSupported(ctx: RegistrationContext, s: string): boolean {
```

ts/src/header-validator/validate-json.ts
```diff
@@ -725,6 +726,101 @@ const priorityField: StructFields<Priority> = {
...
+export type KeyPiece = {
+  keyPiece: bigint
+}
+
+const keyPieceField: StructFields<KeyPiece> = {
+  keyPiece: field('key_piece', hex128),
+}
+
+export type AggregatableDebugReportingData = KeyPiece & {
+  types: string[]
```
apasel422, line 753: Should we make this a set instead of an array?

ts/src/header-validator/validate-json.ts
apasel422, line 729: Maybe just make this a method on `RegistrationContext`?

ts/src/header-validator/validate-json.ts
```diff
@@ -725,6 +726,101 @@ const priorityField: StructFields<Priority> = {
...
+function aggregationCoordinatorOriginField(
+  ctx: RegistrationContext,
+  j: JsonDict
+): Maybe<string> {
+  return field(
+    'aggregation_coordinator_origin',
+    aggregationCoordinatorOrigin,
+    ctx.vsv.aggregationCoordinatorOrigins[0]
+  )(ctx, j)
+}
```
apasel422, line 807: Instead of making this a function that immediately invokes the function returned by `field`, can we make this work more similarly to https://github.com/WICG/attribution-reporting-api/blob/69bed9def9b95eb796d8abf9c65943b374b60d28/ts/src/header-validator/validate-json.ts#L704-L710?

Then `Trigger` and `AggregatableDebugReportingConfig` could both extend an `AggregationCoordinatorOrigin` type.

ts/src/header-validator/validate-json.ts
```diff
@@ -907,6 +1003,30 @@ export type TriggerSpec = {
...
+  j: Json
+): Maybe<SourceAggregatableDebugReportingConfig> {
+  return struct(ctx, j, {
+    budget: field('budget', aggregatableValue),
+
+    ...aggregatableDebugReportingConfig,
+  }).filter((s) => {
+    for (const d of s.debugData) {
+      if (d.value > s.budget) {
+        ctx.error(`data contains value greater than budget (${s.budget})`)
```
apasel422, line 1022: Would it be possible to use the scope corresponding to the array item's value field?

One way to do that would be to process the `budget` field first, and pass it in to the function that validates the debug data.

ts/src/header-validator/validate-json.ts
```diff
@@ -1526,17 +1657,21 @@ function trigger(ctx: RegistrationContext, j: Json): Maybe<Trigger> {
...
           'trigger_context_id',
           (ctx, j) => triggerContextID(ctx, j, aggregatableSourceRegTimeVal),
           null
         ),
+        aggregatableDebugReporting: field(
+          'aggregatable_debug_reporting',
+          (ctx, j) =>
+            struct(ctx, j, {
+              ...aggregatableDebugReportingConfig,
+            }),
```
apasel422, line 1672:
```suggestion
            struct(ctx, j, aggregatableDebugReportingConfig),
```
Note sure if that works though.

apasel422 (changes requested):

ts/src/header-validator/validate-json.ts
linnan-github: We treat supported and unsupported types differently. Unsupported types are just ignored, no matter whether it duplicates or not, but we error for duplicated supported types. Therefore, it would be easier to use an array and validate later.

apasel422: OK, that's fine for now, but I may try to clean this up later.

ts/src/header-validator/validate-json.ts
linnan-github: I thought about that as well, but I don't know what's the best way to set the default which needs ctx. Any suggestions?

apasel422: I can take a look at cleaning this up in a followup field.

ts/src/header-validator/validate-json.ts
linnan-github: Alternatively, I can process `budget` as suggested, and make that non-StructFields, similar to the aggregation coordinator origin. Wdyt?

apasel422: We do something similar in https://github.com/WICG/attribution-reporting-api/blob/d6bbe748163187f2c017a6940f5dc024dc0b6169/ts/src/header-validator/validate-json.ts#L473-L492 where we process one field before the other.

linnan-github: yeah, but then it can't be a `StructFields` IIUC?

apasel422: Feel free to leave it for now. I'll come in and see what can be added later. Maybe just add a TODO for giving more precise path information.

linnan-github: Added TODO, thanks.

index.bs
```diff
@@ -1552,6 +1574,120 @@ To <dfn>parse a filter pair</dfn> given a [=map=] |map|:
...
+<ul dfn-for="aggregatable-debug-reporting JSON key">
+<li>"<dfn><code>aggregation_coordinator_origin</code></dfn>"
+<li>"<dfn><code>debug_data</code></dfn>"
+<li>"<dfn><code>key_piece</code></dfn>"
+<li>"<dfn><code>types</code></dfn>"
+<li>"<dfn><code>value</code></dfn>"
+</ul>
+
+To <dfn>parse aggregatable debug reporting data</dfn> given a |dataList|, a
+positive integer |maxValue|, and a [=boolean=] |isSource|:
```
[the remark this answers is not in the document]
linnan-github: Makes sense, updated, thanks.

apasel422 (approved):

apasel422 (approved):

PR #1299, opened 2024-05-23
Rename integer regex

agarant:
The regex are not validating the sizes of the integers. Renaming to make this clearer.

### DIFF
--- a/ts/src/header-validator/validate-json.ts
+++ b/ts/src/header-validator/validate-json.ts
@@ -251,7 +251,12 @@ function matchesPattern(
 function uint64(ctx: Context, j: Json): Maybe<bigint> {
   return string(ctx, j)
     .filter((s) =>
-      matchesPattern(ctx, s, uintRegex, 'must be a non-negative integer')
+      matchesPattern(
+        ctx,
+        s,
+        uintRegex,
+        'string must represent a non-negative integer'
+      )
     )
     .map(BigInt)
     .filter((n) =>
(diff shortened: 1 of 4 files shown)


ts/src/header-validator/validate-json.ts
```diff
@@ -250,7 +250,9 @@ function matchesPattern(
 
 function uint64(ctx: Context, j: Json): Maybe<bigint> {
   return string(ctx, j)
-    .filter((s) => matchesPattern(ctx, s, uint64Regex, 'must be a uint64'))
+    .filter((s) =>
+      matchesPattern(ctx, s, uintRegex, 'must be a non-negative integer')
```
apasel422, line 254: This PR LGTM overall.

I wonder if we should just get rid of the extra message here and instead only show the regex that must be matched, since saying it must be an integer might be interpreted as must be a JSON number.

agarant: good idea!

apasel422: We can also change https://github.com/WICG/attribution-reporting-api/blob/d6bbe748163187f2c017a6940f5dc024dc0b6169/ts/src/header-validator/validate-json.ts#L321.

apasel422: Although, that regex actually does impose a size limit, so never mind.

apasel422: Sounds good!

apasel422 (approved):

PR #1300, opened 2024-05-24
Share struct/field logic for JSON and structured headers

apasel422:
This allows the structured header validators to parse values into well-typed objects without ad-hoc logic.

`validate.ts` is effectively a module parameterized over a map type (`object` for JSON, `Map` for structured headers) and a value type (the `Json` type union for JSON, `Item | InnerList` for structured headers).


agarant (approved):
LGTM, the setup becomes a bit more complex, but the actual validation is more streamlined.

PR #1311, opened 2024-05-31
Update attribution_scopes.md - update to when sources are deactivated

akashnadan:
Updating the Attribution Scopes explainer to correct when sources are deactivated

### DIFF
--- a/attribution_scopes.md
+++ b/attribution_scopes.md
@@ -101,7 +101,7 @@ The `attribution_scope_limit` value can be updated during source registration at
 
 ### Deletion logic
 
-If the current trigger is attributed to a source, then all other eligible sources considered for attribution (across all `attribution_scopes`) will be deleted.
+If the current trigger passes the top-level context filter check during the attribution process, then all other eligible sources considered for attribution (across all `attribution_scopes`) will be deleted.
 
 ## Attribution Scope Examples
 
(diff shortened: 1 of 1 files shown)


attribution_scopes.md
```diff
@@ -101,7 +101,7 @@ The `attribution_scope_limit` value can be updated during source registration at
 
 ### Deletion logic
 
-If the current trigger is attributed to a source, then all other eligible sources considered for attribution (across all `attribution_scopes`) will be deleted.
+If the current trigger passes the top-level context filter check during the attribution process, then all other eligible sources considered for attribution (across all `attribution_scopes`) will be deleted.
```
apasel422, line 104:
```suggestion
If the current trigger passes the top-level filter check during the attribution process, then all other eligible sources considered for attribution (across all `attribution_scopes`) will be deleted.
```

attribution_scopes.md
```diff
@@ -158,7 +158,7 @@ The user then converts at a later time on the destination site by purchasing a p
 }
 ```
 
-The API automatically performs attribution between any sources that have `attribution_scopes` that are not disjoint with the trigger `attribution_scopes`. Any sources that do not have an `attribution_scopes` that matches at least one of the trigger registration `attribution_scopes` are deleted (assuming there is a source that gets attributed and a report is generated; if no report is generated then no sources are deleted). In this example, the API caller would receive an attribution report attributing the trigger registration to advertiser1’s second source registration.
+The API automatically performs attribution between any sources that have `attribution_scopes` that are not disjoint with the trigger `attribution_scopes`. Any sources that do not have an `attribution_scopes` that matches at least one of the trigger registration `attribution_scopes` are deleted (assuming the source that is chosen passes the top-level context filter checks; if it does not then no sources are deleted). In this example, the API caller would receive an attribution report attributing the trigger registration to advertiser1’s second source registration.
```
apasel422, line 161:
```suggestion
The API automatically performs attribution between any sources that have `attribution_scopes` that are not disjoint with the trigger `attribution_scopes`. Any sources that do not have an `attribution_scopes` that matches at least one of the trigger registration `attribution_scopes` are deleted (assuming the source that is chosen passes the top-level filter check; if it does not then no sources are deleted). In this example, the API caller would receive an attribution report attributing the trigger registration to advertiser1’s second source registration.
```

apasel422 (changes requested):

giladbarkan-github:
Looks good to me with Andrew's suggestions. The other statement that I commented seeming like a mistake is "Allow finer-grained filtering before attribution, trading off additional filtering with noise levels." Noising is still calculated in the same way as it is currently.

giladbarkan-github (approved):

PR #1282, opened 2024-05-13
Attribution Scopes - New Explainer

akashnadan:
Related to #1215 

Draft Explainer for Attribution Scopes Proposal.

This proposal also helps to address feedback from #1229

PR #1318, opened 2024-06-04
Update attribution_scopes.md - addressing feedback comments

akashnadan:
Addressing feedback comments from #1282


attribution_scopes.md
```diff
@@ -89,9 +89,9 @@ If there are multiple sources whose `attribution_scopes` contains at least one o
...
 
-If a source registration is specified with a configuration that has a higher number of event states than the most recent `max_event_states` for the same reporting origin, then the source will be rejected and the registration will fail. Additionally, if the `max_event_states` field is changed in a future source registration, then all other previous pending source registrations with a different `max_event_states` will be ignored in subsequent attribution report generation flows, but will still count towards rate limits. 
+For sources that use the `attribution_scopes` feature, if the source registration is specified with a configuration that has a higher number of event states than the most recent `max_event_states` for the same reporting origin, then the source will be rejected and the registration will fail. Additionally, if the `max_event_states` field is changed in a future source registration, then all other previous pending source registrations with a different `max_event_states` will be ignored in subsequent attribution report generation flows, but will still count towards rate limits.
```
dmdabbs, line 94: >...then the source will be rejected and the registration will fail.

Will an associated verbose debugging failure mode get reported?

giladbarkan-github: By "previous pending source registrations," do you mean sources that use the attribution scope feature? I guess I'm not completely clear on if we're looking at the explicit field, `max_event_states`, or the actual trigger states for the source.

akashnadan: @dmdabbs yes, we do plan to support a verbose debug report that will notify of this failure.

@giladbarkan-github that statement refers to only sources that use the attribution scope feature. I will clarify in the text

attribution_scopes.md
```diff
@@ -101,7 +101,7 @@ The `attribution_scope_limit` value can be updated during source registration at
 
 ### Deletion logic
 
-If the current trigger passes the top-level filter check during the attribution process, then all other eligible sources considered for attribution (across all `attribution_scopes`) will be deleted.
+If the current trigger passes the top-level filter check during the attribution process, then all other eligible sources considered for attribution (across all `attribution_scopes` and including eligible sources that do not use the `attribution_scopes` feature) will be deleted.
```
apasel422, line 104: What does `eligible` mean here?

akashnadan: @apasel422 in this case `eligible` refers to all sources in storage that match <reporting origin, `destination` site> based on this [section of the Event explainer](https://github.com/WICG/attribution-reporting-api/blob/main/EVENT.md#trigger-attribution-algorithm)

apasel422 (approved):

attribution_scopes.md
```diff
@@ -89,9 +89,9 @@ If there are multiple sources whose `attribution_scopes` contains at least one o
 
 If the trigger registration's `attribution_scopes` is empty, then all sources are considered for attribution.
 
-Once an `attribution_scope_limit` is set, the last K values (where K = `attribution_scope_limit`) of `attribution_scopes` will be considered the final set of `attribution_scopes` values and any source with additional `attribution_scopes` values will be treated as if the attribution scopes were empty.
+Once an `attribution_scope_limit` is set, the last K values (where K = `attribution_scope_limit`) of `attribution_scopes` will be considered the final set of `attribution_scopes` values per reporting origin per destination site, and any source with additional `attribution_scopes` values will be treated as if the attribution scopes were empty.
```
linnan-github, line 92: Per recent discussion, sources with those scopes will be deleted instead. Also shall we mention the following recent change?

1. If the new source uses the attribution scopes feature, all previous pending source registrations that do not use the feature will be deleted.
2. If the new source doesn't use the attribution scopes feature, all previous pending source registrations will be treated as if the feature is not used.

akashnadan: I have updated the text to reflect this behavior @linnan-github

linnan-github:
Please resolve the merge conflict, thanks.

linnan-github (approved):

attribution_scopes.md
```diff
@@ -222,7 +222,7 @@ In this example the API will start by finding any sources that have an `attribut
 
 ## Alternatives Considered
 
-One alternative that was considered was instead of using `attribution_scopes` the API would perform attribution matching on a finer granularity than site (e.g. origin or path). This approach would allow the rest of the API logic to continue working as it does currently. However, the `attribution_scopes` proposal gives users additional flexibility in case their URL structures do not match their logical scope structures. Additionally, matching on finer granularity than site may open additional potential attack vectors and would require additional rate limits.
+One alternative that was considered was instead of using `attribution_scopes` the API would perform attribution matching on a finer granularity than site (e.g. origin or path). This approach would allow the rest of the API logic to continue working as it does currently. However, the `attribution_scopes` proposal gives users additional flexibility in case their URL structures do not match their logical scope structures. , matching on finer granularity than site may open additional potential attack vectors and would require additional rate limits.
```
apasel422, line 225:
```suggestion
One alternative that was considered was instead of using `attribution_scopes` the API would perform attribution matching on a finer granularity than site (e.g. origin or path). This approach would allow the rest of the API logic to continue working as it does currently. However, the `attribution_scopes` proposal gives users additional flexibility in case their URL structures do not match their logical scope structures. Matching on finer granularity than site may open additional potential attack vectors and would require additional rate limits.
```
?

apasel422 (changes requested):

apasel422 (approved):

PR #1319, opened 2024-06-04
Allow Maybe methods to invoke callbacks with additional arguments

apasel422:
Combined with swapping the order of `CtxFunc`'s parameters, this allows a number of lambdas and lambda-captures to be eliminated.

### DIFF
--- a/ts/src/header-validator/validate-json.ts
+++ b/ts/src/header-validator/validate-json.ts
@@ -78,12 +78,12 @@ type StructFields<
 > = validate.StructFields<T, JsonDict, C>
 
 function struct<T extends object, C extends Context>(
-  ctx: C,
   d: Json,
+  ctx: C,
   fields: StructFields<T, C>,
   warnUnknown: boolean = true
 ): Maybe<T> {
-  return object(ctx, d).map((d) => structInternal(ctx, d, fields, warnUnknown))
+  return object(d, ctx).map((d) => structInternal(d, ctx, fields, warnUnknown))
 }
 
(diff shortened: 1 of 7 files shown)


ts/src/header-validator/validate-json.ts
```diff
@@ -285,61 +277,66 @@ function suitableScope(
...
 function suitableOrigin(
-  ctx: Context,
   j: Json,
+  ctx: Context,
   rejectExtraComponents: boolean = false
 ): Maybe<string> {
-  return string(ctx, j).map((s) =>
-    suitableScope(ctx, s, 'origin', (u) => u.origin, rejectExtraComponents)
+  return string(j, ctx).map(
+    suitableScope,
```
agarant, line 285: (general comment) I'm not 100% convinced that the approach is an improvement.

It looks good when you know what's going on.

But it is quite opaque at first sight.

Have you considered going the other way around meaning that we'd have the function receive their extra properties and return a function that receives JSON and the context object, partially applying properties when needed.

```ts
return string(j, ctx).map(
    suitableScope('origin', (u) => u.origin, rejectExtraComponents)
)
  ```

apasel422: I did think about that, but it means that every function that wants to be used this way needs to be return a function (effectively currying itself), whereas this approach works with the existing lambda approach *and* this binding shorthand.

If you think it's too hard, that's fine, and I can abandon.

agarant: good with me to proceed with the suggested approach.

agarant (approved):

PR #1322, opened 2024-06-06
Prevent duplicates in event-level-report destination field

apasel422:
Since destinations are deduplicated during source registration, they can never appear in the report.

### DIFF
--- a/ts/src/header-validator/validate-json.ts
+++ b/ts/src/header-validator/validate-json.ts
@@ -1576,9 +1576,9 @@ function reportDestination(ctx: Context, j: Json): Maybe<string | string[]> {
         maxLength: 3,
       }).filter((v) => {
         for (let i = 1; i < v.length; ++i) {
-          if (v[i]! < v[i - 1]!) {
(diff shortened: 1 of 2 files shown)


ts/src/header-validator/validate-json.ts
```diff
@@ -1576,9 +1576,9 @@ function reportDestination(ctx: Context, j: Json): Maybe<string | string[]> {
         maxLength: 3,
       }).filter((v) => {
         for (let i = 1; i < v.length; ++i) {
-          if (v[i]! < v[i - 1]!) {
+          if (v[i]! <= v[i - 1]!) {
```
linnan-github, line 1579: Not related to this PR, but looks like warning is reported for`https://A.test`: URL components other than site (https://a.test) will be ignored

linnan-github (approved):

ts/src/header-validator/validate-json.ts
apasel422: I can look into that separately.

Issue #1309, opened 2024-05-29
Allow max attributions per rate-limit window during event report replacement


giladbarkan-github:
In step 15 of https://wicg.github.io/attribution-reporting-api/#trigger-event-level-attribution, we exit early if the number of persisted attribution rate limit records is equal to the max allowed; which means we reach step 17, replacing event-level reports, only if the number of persisted attribution rate limit records is less than the max allowed, even though we keep the number of attribution rate limit records in line with the number of reports by removing attribution rate limit records (step 12 of https://wicg.github.io/attribution-reporting-api/#maybe-replace-event-level-report).

Since having (max-1) persisted attribution rate limit records results in report replacements unlimited by max attributions per rate-limit window, it seems arbitrary to limit such report replacements when we keep max persisted attribution rate limit records.

PR #1326, opened 2024-06-08
Allow max attribution rate limit when replacing an event report

giladbarkan-github:
Resolves issue https://github.com/WICG/attribution-reporting-api/issues/1309

In step 15 of https://wicg.github.io/attribution-reporting-api/#trigger-event-level-attribution, we exit early if the number of persisted attribution rate limit records is equal to the max allowed; which means we reach step 17, replacing event-level reports, only if the number of persisted attribution rate limit records is less than the max allowed, even though we keep the number of attribution rate limit records in line with the number of reports by removing attribution rate limit records (step 12 of https://wicg.github.io/attribution-reporting-api/#maybe-replace-event-level-report).

Since having (max-1) persisted attribution rate limit records results in report replacements unlimited by max attributions per rate-limit window, it seems arbitrary to limit such report replacements when we keep max persisted attribution rate limit records.

***
<a href="https://pr-preview.s3.amazonaws.com/giladbarkan-github/attribution-reporting-api/pull/1326.html" title="Last updated on Jul 2, 2024, 8:55 PM UTC (0e9ed97)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/attribution-reporting-api/1326/312deee...giladbarkan-github:0e9ed97.html" title="Last updated on Jul 2, 2024, 8:55 PM UTC (0e9ed97)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -3943,6 +3943,25 @@ To <dfn>trigger event-level attribution</dfn> given an [=attribution trigger=] |
 1. Let |report| be the result of running [=obtain an event-level report=] with |sourceToAttribute|,
     |trigger|'s [=attribution trigger/trigger time=], |trigger|'s [=attribution trigger/debug key=],
     |matchedConfig|'s [=event-level trigger configuration/priority=], and |specEntry|.
+1. If |sourceToAttribute|'s [=attribution source/event-level attributable=] value
+    is false:
+     1. Return the [=triggering result=] ("<code>[=triggering status/dropped=]</code>",
+         ("<code>[=trigger debug data type/trigger-event-excessive-reports=]</code>", |report|)).
+1. If the result of running [=maybe replace event-level report=] with |sourceToAttribute| and |report| is:
+    <dl class="switch">
+    : "<code>[=event-level-report-replacement result/add-new-report=]</code>"
+    ::
+         1. Do nothing.
+    : "<code>[=event-level-report-replacement result/drop-new-report-none-to-replace=]</code>"
+    ::
+         1. Return the [=triggering result=] ("<code>[=triggering status/dropped=]</code>",
+             ("<code>[=trigger debug data type/trigger-event-excessive-reports=]</code>", |report|)).
+    : "<code>[=event-level-report-replacement result/drop-new-report-low-priority=]</code>"
+    ::
+         1. Return the [=triggering result=] ("<code>[=triggering status/dropped=]</code>",
+             ("<code>[=trigger debug data type/trigger-event-low-priority=]</code>", |report|)).
+
+    </dl>
 1. Let |rateLimitRecord| be a new [=attribution rate-limit record=] with the items:
     : [=attribution rate-limit record/scope=]
     :: "<code>[=rate-limit scope/event-attribution=]</code>"
@@ -3960,29 +3979,11 @@ To <dfn>trigger event-level attribution</dfn> given an [=attribution trigger=] |
     :: |report|'s [=event-level report/report ID=]
 1. If the result of running [=check if attribution should be blocked by rate limits=]
     with |trigger|, |sourceToAttribute|, and |rateLimitRecord| is not null, return it.
-1. If |sourceToAttribute|'s [=attribution source/event-level attributable=] value
-    is false:
-     1. Return the [=triggering result=] ("<code>[=triggering status/dropped=]</code>",
-         ("<code>[=trigger debug data type/trigger-event-excessive-reports=]</code>", |report|)).
-1. If the result of running [=maybe replace event-level report=] with |sourceToAttribute| and |report| is:
-    <dl class="switch">
-    : "<code>[=event-level-report-replacement result/add-new-report=]</code>"
-    ::
-         1. Let |numMatchingReports| be the number of entries in the [=event-level report cache=] whose
-             [=event-level report/attribution destinations=] [=set/contains=] |trigger|'s [=attribution trigger/attribution destination=].
-         1. If |numMatchingReports| is greater than or equal to the user agent's [=max event-level reports per attribution destination=]:
-             1. Return the [=triggering result=] ("<code>[=triggering status/dropped=]</code>",
-                 ("<code>[=trigger debug data type/trigger-event-storage-limit=]</code>", null)).
-    : "<code>[=event-level-report-replacement result/drop-new-report-none-to-replace=]</code>"
-    ::
-         1. Return the [=triggering result=] ("<code>[=triggering status/dropped=]</code>",
-             ("<code>[=trigger debug data type/trigger-event-excessive-reports=]</code>", |report|)).
-    : "<code>[=event-level-report-replacement result/drop-new-report-low-priority=]</code>"
-    ::
-         1. Return the [=triggering result=] ("<code>[=triggering status/dropped=]</code>",
-             ("<code>[=trigger debug data type/trigger-event-low-priority=]</code>", |report|)).
-
-    </dl>
+1. Let |numMatchingReports| be the number of entries in the [=event-level report cache=] whose
+    [=event-level report/attribution destinations=] [=set/contains=] |trigger|'s [=attribution trigger/attribution destination=].
+1. If |numMatchingReports| is greater than or equal to the user agent's [=max event-level reports per attribution destination=]:
+    1. Return the [=triggering result=] ("<code>[=triggering status/dropped=]</code>",
+        ("<code>[=trigger debug data type/trigger-event-storage-limit=]</code>", null)).
 1. Let |triggeringStatus| be "<code>[=triggering status/attributed=]</code>".
 1. Let |debugData| be null.
 1. If |sourceToAttribute|'s [=attribution source/randomized response=] is:


index.bs
```diff
@@ -3680,31 +3676,33 @@ To <dfn>trigger event-level attribution</dfn> given an [=attribution trigger=] |
     :: null
     : [=attribution rate-limit record/event-level report ID=]
     :: |report|'s [=event-level report/report ID=]
-1. If the result of running [=check if attribution should be blocked by rate limits=]
-    with |trigger|, |sourceToAttribute|, and |rateLimitRecord| is not null, return it.
 1. If |sourceToAttribute|'s [=attribution source/event-level attributable=] value
```
apasel422, line 3679: Let's move this above the `Let |rateLimitRecord| be...` step since that variable isn't needed yet.

index.bs
```diff
@@ -3612,7 +3608,7 @@ To <dfn>maybe replace event-level report</dfn> given an [=attribution source=]
     If this assumption does not hold then |rateLimitRecord| might be null.
 
 1. [=set/Remove=] |rateLimitRecord| from the [=attribution rate-limit cache=].
-1. Return "<code>[=event-level-report-replacement result/add-new-report=]</code>".
+1. Return null.
```
apasel422, line 3611: Why are we changing this? The `add-new-report` value is now unused.

apasel422 (changes requested):

index.bs
giladbarkan-github: Can we just remove references to `add-new-report`?

index.bs
giladbarkan-github: This PR is about conditioning this check with `If sourceToAttribute’s number of event-level reports is less than sourceToAttribute’s max number of event-level reports:`

Should we move both? Then what about the "Otherwise?"

apasel422: Prior to this PR, `|rateLimitRecord|` was needed in the step immediately following its creation. Now it is not.

giladbarkan-github: I'm confused -- the section highlighted, `...with |trigger|, |sourceToAttribute|, and |rateLimitRecord| is not null...`, uses the variable, `rateLimitRecord`. How can we move it "above the `Let |rateLimitRecord| be...` step" ?

index.bs
giladbarkan-github: It seems odd to me to name a value when there is no handling for it. I don't feel strongly about it. Would you rather have the return, `add-new-report`, when we don't need to handle it explicitly in this sequence?

index.bs
apasel422: [Step 14](https://pr-preview.s3.amazonaws.com/giladbarkan-github/attribution-reporting-api/pull/1326.html#ref-for-attribution-rate-limit-record%E2%91%A0%E2%91%A4) creates it but it isn't needed until Step 17.1. Please move creation of `|rateLimitRecord|` directly above what is currently Step 17.

index.bs
apasel422: Since we have to return a value from the algorithm in all paths, a named value is more descriptive than `null` and easier to search for. You can even add an assertion after the switch that the value is equal to that value.

index.bs
giladbarkan-github: Done!

index.bs
giladbarkan-github: Done!

giladbarkan-github:
Does this change now hide or makes an assumption about the relationship between a source's `max number of event-level reports`  and attribution rate limit?

index.bs
```diff
@@ -3680,31 +3682,29 @@ To <dfn>trigger event-level attribution</dfn> given an [=attribution trigger=] |
...
-     1. Return the [=triggering result=] ("<code>[=triggering status/dropped=]</code>",
-         ("<code>[=trigger debug data type/trigger-event-excessive-reports=]</code>", |report|)).
-1. If the result of running [=maybe replace event-level report=] with |sourceToAttribute| and |report| is:
-    <dl class="switch">
-    : "<code>[=event-level-report-replacement result/add-new-report=]</code>"
-    ::
-         1. Let |numMatchingReports| be the number of entries in the [=event-level report cache=] whose
-             [=event-level report/attribution destinations=] [=set/contains=] |trigger|'s [=attribution trigger/attribution destination=].
-         1. If |numMatchingReports| is greater than or equal to the user agent's [=max event-level reports per attribution destination=]:
+1. If |sourceToAttribute|'s [=attribution source/number of event-level reports=] is less than
```
linnan-github, line 3685: Can we still have these checks in "maybe replace event-level report", and perform the attribution rate-limit check after that?  I understand that in the case of replacement, it's guaranteed to be pass, but it seems to be clearer. Otherwise, we probably need a note why we don't check attribution rate-limits in that case.

giladbarkan-github: I thought about it, but it seems to complicate the return result and parameters passed to "maybe replace event-level report" and doesn't seem directly related to that procedure. I'm happy to add a note because the assumption about the relationship between max number of event-level reports and attribution rate limit would still seem to be there and perhaps should be explicit.

giladbarkan-github: Actually, attribution rate limit is vendor specific and they could choose to make it lower than max reports, in which case this change wouldn't work, right?

giladbarkan-github: I'm confusing myself thinking about the scenarios for the limits :). @linnan-github could you please explain why you think a note is necessary? Is there a scenario where this logic breaks?

linnan-github: I’m a little confused, what I meant to say is just move attribution rate-limit check later and only  check this when the result is add new report from “maybe replace the event-level report”. This aligns with how we deal with the storage limit. The old rate-limit record would have been deleted in case of replacement and the check should pass.

I think at least adding a note to explain why we don’t need this check when the number of event-level reports equals the limit.

giladbarkan-github: @linnan-github what would happen under your proposal if, as a vendor, I set attribution rate limit to 10 (assuming the constant, max event reports, stays at 20)?

giladbarkan-github: Ah, that seems a simpler change than what I've done here :) Thanks!

giladbarkan-github: Pushed. Is this what you had in mind?

index.bs
```diff
@@ -3705,6 +3688,27 @@ To <dfn>trigger event-level attribution</dfn> given an [=attribution trigger=] |
              ("<code>[=trigger debug data type/trigger-event-low-priority=]</code>", |report|)).
 
     </dl>
+1. Let |rateLimitRecord| be a new [=attribution rate-limit record=] with the items:
```
linnan-github, line 3691: To be consistent, shall we move the storage-limit check in line 3676 here as well, and we just break for `add-new-report`.

index.bs
```diff
@@ -3705,6 +3688,27 @@ To <dfn>trigger event-level attribution</dfn> given an [=attribution trigger=] |
...
+    : [=attribution rate-limit record/time=]
+    :: |sourceToAttribute|'s [=attribution source/source time=]
+    : [=attribution rate-limit record/expiry time=]
+    :: null
+    : [=attribution rate-limit record/event-level report ID=]
+    :: |report|'s [=event-level report/report ID=]
+1. If the result of running [=check if attribution should be blocked by rate limits=]
+    with |trigger|, |sourceToAttribute|, and |rateLimitRecord| is not null, return it.
+
+    Note: When [=max attributions per rate-limit window=] has already been
```
linnan-github, line 3709: Do we still need this note which seems to be self-explained by the algorithm?

index.bs
giladbarkan-github: Done -- is this what you had in mind?

index.bs
giladbarkan-github: Removed.

index.bs
linnan-github: yes, thanks.

linnan-github:
@pasin30055 Could you PTAL, thanks!

linnan-github (approved):

vikassahu29 (approved):

apasel422 (approved):

index.bs
```diff
@@ -3968,11 +3951,7 @@ To <dfn>trigger event-level attribution</dfn> given an [=attribution trigger=] |
     <dl class="switch">
     : "<code>[=event-level-report-replacement result/add-new-report=]</code>"
     ::
-         1. Let |numMatchingReports| be the number of entries in the [=event-level report cache=] whose
-             [=event-level report/attribution destinations=] [=set/contains=] |trigger|'s [=attribution trigger/attribution destination=].
-         1. If |numMatchingReports| is greater than or equal to the user agent's [=max event-level reports per attribution destination=]:
-             1. Return the [=triggering result=] ("<code>[=triggering status/dropped=]</code>",
-                 ("<code>[=trigger debug data type/trigger-event-storage-limit=]</code>", null)).
+         1. [=iteration/Break=].
```
giladbarkan-github, line 3954: @apasel422 does break seem the correct action to skip from here to the next step (16)?

apasel422: Actually no, because there is no ongoing loop. I'd just say:

```
1. Do nothing.
```

giladbarkan-github: Done!

PR #1333, opened 2024-06-11
Check size of sets after deduplication instead of before

### LINKED ISSUE #1321: Header validator inconsistent with spec re size checks
Per https://wicg.github.io/attribution-reporting-api/#parse-attribution-destinations, the size of the `destination` field is supposed to be checked after URL-to-site transformation and deduplication, but the validator checks the size of the JSON list itself, causing some registrations that would be allowed by the spec to be prohibited by the validator.

This also applies to [`filter_data` values](https://wicg.github.io/attribution-reporting-api/#parse-filter-data).

apasel422:
To match the spec.

Fixes #1321.

### DIFF
--- a/ts/src/header-validator/source.test.ts
+++ b/ts/src/header-validator/source.test.ts
@@ -210,13 +210,59 @@ const testCases: TestCase[] = [
     ],
   },
   {
-    name: 'destination-list-too-long',
+    name: 'destination-list-size-ok-after-dedup',
     json: `{"destination": [
       "https://a.test",
       "https://b.test/1",
       "https://b.test/2",
       "https://c.test/3"
     ]}`,
+    expectedWarnings: [
+      {
+        msg: 'URL components other than site (https://b.test) will be ignored',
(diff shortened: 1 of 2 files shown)


ts/src/header-validator/source.test.ts
```diff
@@ -2108,7 +2177,7 @@ const testCases: TestCase[] = [
     name: 'trigger-data-too-many-within',
     json: JSON.stringify({
       destination: 'https://a.test',
-      trigger_data: Array.from({ length: 33 }, () => 0),
+      trigger_data: Array.from({ length: 33 }, (_, i) => i),
```
linnan-github, line 2180: Do we need to update spec for `trigger_data`: https://wicg.github.io/attribution-reporting-api/#parse-trigger-data-into-a-trigger-spec-map?

linnan-github: `trigger_data` doesn't allow duplication, so should be fine.

PR #1337, opened 2024-06-13
Remove Maybe.flatten function

apasel422:
It was unsound because it could be called with a Maybe<T> but its type parameter T instantiated with Maybe<T>, resulting in an apparent return type of Maybe<Maybe<T>>, but the instanceof check would cause the original Maybe<T> to be returned.

For example, the following code logs "number" despite y's type indicating that its value should be another Maybe:

```ts
const x = Maybe.some(5)
const y: Maybe<Maybe<number>> = Maybe.flatten<Maybe<number>>(x)
console.log(typeof y.value)
```

### DIFF
--- a/ts/src/header-validator/maybe.ts
+++ b/ts/src/header-validator/maybe.ts
@@ -1,14 +1,8 @@
-export type Maybeable<T> = T | Maybe<T>
-
 export class Maybe<T> {
   static readonly None = new Maybe<never>()
 
   static some<T>(this: void, t: T): Maybe<T> {
     return new Maybe(t)
-  }
-
-  static flatten<T>(this: void, t: Maybeable<T>): Maybe<T> {
-    return t instanceof Maybe ? t : Maybe.some(t)
   }
 
   private constructor(private readonly t?: T) {}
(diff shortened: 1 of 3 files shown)


ts/src/header-validator/maybe.ts
```diff
@@ -1,16 +1,10 @@
-export type Maybeable<T> = T | Maybe<T>
-
 export class Maybe<T> {
   static readonly None = new Maybe<never>()
 
   static some<T>(this: void, t: T): Maybe<T> {
     return new Maybe(t)
   }
 
-  static flatten<T>(this: void, t: Maybeable<T>): Maybe<T> {
```
agarant, line 10: Just to make sure I understand correctly.

The edge case is that it doesn't support nested Maybe types. 

If T is a Maybe<number>, it wont return Maybe<Maybe<number>> as expected by the type and will instead return a Maybe<number>. Correct ?

Do we have a use case for these ? 

Intuitively, I would be using .flatten to avoid that case and ensure that I always get back to a Maybe<T>.

apasel422: > If T is a `Maybe<number>`, it wont return `Maybe<Maybe<number>>` as expected by the type and will instead return a Maybe. Correct ?

Correct.

I think it is possible to prevent this using conditional types with `never`, but then all code that transitively calls `flatten` needs to propagate those bounds as well. So I do think it's simpler to just have separate `map` and `flatMap`.

agarant: Yeah, I'm not sure it's possible to statically type it and prevent incorrect inference.

agarant (approved):

PR #1338, opened 2024-06-17
Simplify collection parsing and explicitly mark variance of Maybe<T>

apasel422:


ts/src/header-validator/validate.ts
```diff
@@ -168,26 +168,26 @@ export enum ItemErrorAction {
   earlyExit,
 }
 
-function isCollection<P extends PathComponent, V, C extends Context>(
+function makeCollection<Col, P extends PathComponent, V, C extends Context>(
```
agarant, line 171:
```suggestion
function collection<Col, P extends PathComponent, V, C extends Context>(
```

I'm wondering if just `collection` would be better aligned with the other naming.

agarant (approved):

PR #1339, opened 2024-06-18
Require explicit handling of optional and required fields

apasel422:
This allow us to merge the `field` and `fieldMaybeDefault` functions and makes it harder to accidentally make a field required (or optional).


agarant (approved):
Nice!

PR #1352, opened 2024-06-27
Parsing changes for Pre-Attribution Filtering: Attribution Scope

feifeiji89:
The current attribution logic in the Attribution Reporting API may not be ideal for use-cases where an ad-tech needs more fine grain control over the attribution granularity (i.e. campaign, product, conversion ID, etc.) before a source is chosen for attribution. Currently available features such as top-level filters are not fully sufficient for this use-case because they happen after a source has been selected (i.e. after attribution). We can optionally support this use-case by allowing callers of the API to specify predefined attribution scopes that will be considered for filtering before attributing a source, in order to more efficiently extract utility out of the API.

This PR covers the parsing logic for the new API surface.

***
<a href="https://pr-preview.s3.amazonaws.com/feifeiji89/attribution-reporting-api/pull/1352.html" title="Last updated on Jul 8, 2024, 4:14 PM UTC (3841d5f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/attribution-reporting-api/1352/bacfb84...feifeiji89:3841d5f.html" title="Last updated on Jul 8, 2024, 4:14 PM UTC (3841d5f)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2736,8 +2736,8 @@ To <dfn>parse attribution scopes for source</dfn> from a [=map=] |map| and a pos
 1. Return |result|.
 
 Note: Empty attribution scopes are not allowed if |attributionScopeLimit| is set,
-to prevent the selection of both sources with and without scopes 
-if |attributionScopeLimit| is set to 1, which would effectively result in two scopes.
+to prevent the selection of both sources with and without scopes which would effectively
+result in |attributionScopeLimit| + 1 scopes.
 
 To <dfn noexport>parse source-registration JSON</dfn> given a [=byte sequence=]
 |json|, a [=suitable origin=] |sourceOrigin|, a [=suitable origin=] |reportingOrigin|, a
--- a/index.bs
+++ b/index.bs
@@ -845,6 +845,12 @@ An attribution source is a [=struct=] with the following items:
 :: An [=aggregatable debug reporting config=].
 : <dfn>destination limit priority</dfn>
 :: A 64-bit integer.
+: <dfn>attribution scope limit</dfn>
+:: Null or a positive 32-bit integer representing the number of distinct [=attribution source/attribution scopes=] allowed per [=attribution source/attribution destinations|attribution destination=] for the source's [=attribution source/reporting origin=].
+: <dfn>attribution scopes</dfn>
+:: A [=set=] of [=strings=].
+: <dfn>max event states</dfn>
+:: A positive integer representing the maximum number of [=trigger states=] for [=source type/event=] sources per [=attribution source/attribution destinations|attribution destination=] for the source's [=attribution source/reporting origin=].
 
 </dl>
 
@@ -985,6 +991,8 @@ An attribution trigger is a [=struct=] with the following items:
 :: A positive integer.
 : <dfn>aggregatable debug reporting config</dfn>
 :: An [=aggregatable debug reporting config=].
+: <dfn>attribution scopes</dfn>
+:: A [=set=] of [=strings=].
 
 </dl>
 
@@ -1155,6 +1163,7 @@ Possible values are:
 <li>"<dfn><code>source-destination-limit-replaced</code></dfn>"
 <li>"<dfn><code>source-destination-per-day-rate-limit</code></dfn>"
 <li>"<dfn><code>source-destination-rate-limit</code></dfn>"
+<li>"<dfn><code>source-max-event-states-limit</code></dfn>"
 <li>"<dfn><code>source-noised</code></dfn>"
 <li>"<dfn><code>source-reporting-origin-limit</code></dfn>"
 <li>"<dfn><code>source-reporting-origin-per-site-limit</code></dfn>"
@@ -1413,6 +1422,17 @@ per [=aggregatable debug rate-limit window=]. The second controls the total
 [=aggregatable debug report/required aggregatable budget=] of all [=aggregatable debug reports=]
 with a given ([=aggregatable debug rate-limit record/context site=], [=aggregatable debug rate-limit record/reporting site=])
 per [=aggregatable debug rate-limit window=]. Its value is (2<sup>20</sup>, 65536).
+
+<dfn>Default max event states</dfn> is a positive integer that controls the
+default [=attribution source/max event states=]. Its value is 3.
+
+<dfn>Max length of attribution scope for source</dfn> is a positive integer that controls the
+maximum [=string/length=] of an attribution scope from an [=attribution source=]'s [=attribution source/attribution scopes=].
+Its value is 50.
+
+<dfn>Max attribution scopes per source</dfn> is a positive integer that controls the
+maximum [=set/size=] of an [=attribution source=]'s [=attribution source/attribution scopes=].
+Its value is 20.
 
 # Vendor-Specific Values # {#vendor-specific-values}
 
@@ -2401,6 +2421,8 @@ A <dfn>source-registration JSON key</dfn> is one of the following:
 <li>"<dfn><code>aggregatable_report_window</code></dfn>"
 <li>"<dfn><code>aggregation_keys</code></dfn>"
 <li>"<dfn><code>budget</code></dfn>"
+<li>"<dfn><code>attribution_scope_limit</code></dfn>"
+<li>"<dfn><code>attribution_scopes</code></dfn>"
 <li>"<dfn><code>debug_key</code></dfn>"
 <li>"<dfn><code>debug_reporting</code></dfn>"
 <li>"<dfn><code>destination</code></dfn>"
@@ -2412,6 +2434,7 @@ A <dfn>source-registration JSON key</dfn> is one of the following:
 <li>"<dfn><code>expiry</code></dfn>"
 <li>"<dfn><code>filter_data</code></dfn>"
 <li>"<dfn><code>max_event_level_reports</code></dfn>"
+<li>"<dfn><code>max_event_states</code></dfn>"
 <li>"<dfn><code>priority</code></dfn>"
 <li>"<dfn><code>source_event_id</code></dfn>"
 <li>"<dfn><code>start_time</code></dfn>"
@@ -2687,6 +2710,32 @@ non-negative integer |defaultBudget|, and an [=aggregatable debug reporting conf
     with |value|, |budget|, |supportedTypes|, and |defaultConfig|.
 1. Return the [=tuple=] (|budget|, |config|).
 
+To <dfn>parse max event states</dfn> from a [=map=] |map| and
+a possibly null 32-bit positive integer |attributionScopeLimit|:
+1. Let |maxEventStates| be [=default max event states=].
+1. If |map|["<code>[=source-registration JSON key/max_event_states=]</code>"] [=map/exists=], set |maxEventStates| to |map|["<code>[=source-registration JSON key/max_event_states=]</code>"].
+1. If |maxEventStates| is not an integer or is less than or equal to 0, return an error.
+1. If |maxEventStates| is greater than the user agent's [=max trigger-state cardinality=], return an error.
+1. If |maxEventStates| is not equal to [=default max event states=] and |attributionScopeLimit| is null, return an error.
+1. Return |maxEventStates|.
+
+To <dfn>parse attribution scopes for source</dfn> from a [=map=] |map| and a possibly null 32-bit positive integer |attributionScopeLimit|:
+1. Let |result| be a new [=set=].
+1. If |map|["<code>[=source-registration JSON key/attribution_scopes=]</code>"] does not [=map/exist=], return |result|.
+1. Let |values| be |map|["<code>[=source-registration JSON key/attribution_scopes=]</code>"].
+1. If |values| is not a [=list=], return an error.
+1. [=list/iterate|For each=] |value| of |values|:
+    1. If |value| is not a [=string=], return an error.
+    1. If |value|'s [=string/length=] is greater than [=max length of attribution scope for source=], return an error.
+    1. [=set/Append=] |value| to |result|.
+1. If |attributionScopeLimit| is null:
+    1. If |result| is not [=set/is empty|empty=], return an error.
+    1. Return |result|.
+1. If |result| [=set/is empty=], return an error.
+1. If |result|'s [=set/size=] is greater than |attributionScopeLimit|, return an error.
+1. If |result|'s [=set/size=] is greater than [=max attribution scopes per source=], return an error.
+1. Return |result|.
+
 To <dfn noexport>parse source-registration JSON</dfn> given a [=byte sequence=]
 |json|, a [=suitable origin=] |sourceOrigin|, a [=suitable origin=] |reportingOrigin|, a
 [=source type=] |sourceType|, a [=moment=] |sourceTime|, and a [=boolean=] |fenced|:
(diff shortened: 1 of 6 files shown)


index.bs
```diff
@@ -3165,6 +3229,10 @@ To <dfn>process an attribution source</dfn> given an [=attribution source=] |sou
 1. Set |source|'s [=attribution source/number of event-level reports=] to 0 if
     |source|'s [=attribution source/randomized response=] is null,
     [=attribution source/randomized response=]'s [=list/size=] otherwise.
+1. Let |states| be the [=obtain a set of possible trigger states|number of possible trigger states=] obtained with |randomizedResponseConfig|.
```
linnan-github, line 3232: Can this be moved to the follow up PR as it’s not about parsing.

ts/src/header-validator/validate-json.ts
```diff
@@ -1335,6 +1394,44 @@ function eventTriggerData(
...
+      return false
+    }
+    return true
+  }
+
+  return set(
+    j,
+    ctx,
+    (j) => string(j, ctx).filter(attributionScopeStringLength),
+    {
```
linnan-github, line 1429: minLength should be 1?

linnan-github (changes requested):

ts/src/header-validator/source.test.ts
```diff
@@ -2544,6 +2559,222 @@ const testCases: TestCase[] = [
...
+    name: 'attribution-scopes-too-many',
+    json: `{
+      "destination": "https://a.test",
+      "attribution_scope_limit": 1,
+      "attribution_scopes": ["1","2","3","4","5","6","7","8","9","10","11","12","13","14","15","16","17","18","19","20","21"]
+    }`,
+    expectedErrors: [
+      {
+        path: ['attribution_scopes'],
+        msg: 'length must be in the range [0, 20]',
```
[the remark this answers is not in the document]
linnan-github: It looks like the length is checked in two places and different error msg may be emitted depending on the error. That seems a little confusing, but probably fine? Cc @apasel422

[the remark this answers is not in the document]
feifeiji89: > Shouldn’t this be in the range [1, 1]?

The check against attribution_scope_limit is covered in test : `attribution-scopes-size-exceeds-attribution-scope-limit`

I updated the error message to be more accurate.

ts/src/header-validator/validate-json.ts
```diff
@@ -1132,12 +1169,34 @@ function source(j: Json, ctx: SourceContext): Maybe<Source> {
           'destination_limit_priority',
           withDefault(int64, 0n)
         ),
+        attributionScopeLimit: field(
+          'attribution_scope_limit',
+          withDefault(positiveUint32, null)
+        ),
+        attributionScopes: field(
```
linnan-github, line 1176: Wdyt about passing the parsed `attributionScopeLimit` to validate `attribution_scopes` and `max_event_states`?

ts/src/header-validator/validate-json.ts
```diff
@@ -759,6 +767,32 @@ function channelCapacity(s: Source, ctx: SourceContext): void {
...
+        return false
+      }
+      return true
+    }
+    return isInRange(
+      s.attributionScopes.size,
+      ctx,
+      1,
+      s.attributionScopeLimit,
+      `attribution_scope_limit must be greater than or equal to the number of attribution scopes ${s.attributionScopeLimit}`
```
apasel422, line 791:
```suggestion
      `attribution_scope_limit must be greater than or equal to the number of attribution scopes (${s.attributionScopeLimit})`
```

apasel422: But I'm also confused about the wording here, since it looks like we're validating the size of the `attribution_scopes` field, not the value of the `attribution_scope_limit` field. In other words, I'd expect this to be phrased as something like `number of attribution scopes (<number>) must be less than or equal to attribution_scope_limit (<number>)`.

index.bs
```diff
@@ -2687,6 +2709,32 @@ non-negative integer |defaultBudget|, and an [=aggregatable debug reporting conf
...
+1. Let |values| be |map|["<code>[=source-registration JSON key/attribution_scopes=]</code>"].
+1. If |values| is not a [=list=], return an error.
+1. [=list/iterate|For each=] |value| of |values|:
+    1. If |value| is not a [=string=], return an error.
+    1. If |value|'s [=string/length=] is greater than [=max length of attribution scope for source=], return an error.
+    1. [=set/Append=] |value| to |result|.
+1. If |attributionScopeLimit| is null:
+    1. If |result| is not [=set/is empty|empty=], return an error.
+    1. Return |result|.
+1. If |result| [=set/is empty=], return an error.
```
apasel422, line 2733: Could you clarify why it's an error for the scope set to be empty?

apasel422 (changes requested):

ts/src/header-validator/validate-json.ts
feifeiji89: that's my initial thought as well. could you advise how do we change the field scope to attribution_scopes inside another scope? I was having trouble make it work somehow.

apasel422: You won't be able to *change* the scope within a `scope` call, only push a new one, so you'd have to check this outside the outer `ctx.scope` call.

index.bs
feifeiji89: @linnan-github As Nan pointed in a separate thread, to prevent adtechs to select 2 buckets with scopes and no scopes when `attribution_scope_limit = 1`, which essentially gives adtech 2 scopes. so we've decided not to support empty scopes when the limit is specificied. cc'ed you the internal tracking bug.

ts/src/header-validator/validate-json.ts
feifeiji89: thanks, as Nan suggested, I was able to remove this validation method after passing attributionScopeLimit value into `attributionScopes` and `maxEventStates`.

ts/src/header-validator/validate-json.ts
feifeiji89: Done. thanks, it looks much cleaner

ts/src/header-validator/validate-json.ts
```diff
@@ -1052,94 +1060,129 @@ export type Source = CommonDebug &
...
-            triggerSpecsFromTriggerData(j, ctx, triggerSpecsDeps),
-          ...(ctx.parseFullFlex
-            ? {
-                trigger_specs: (j) => triggerSpecs(j, ctx, triggerSpecsDeps),
-              }
-            : {}),
-        },
-        defaultTriggerSpecsVal
-      )(j, ctx)
+  return (
```
linnan-github, line 1069: Is outermost `()` necessary?

ts/src/header-validator/validate-json.ts
```diff
@@ -1335,6 +1378,62 @@ function eventTriggerData(
...
+  return number(j, ctx)
+    .filter(isInteger, ctx)
+    .filter(isInRange, ctx, 1, ctx.vsv.maxTriggerStateCardinality)
+    .filter((n) => {
+      if (
+        attributionScopeLimit.value === null &&
+        n !== constants.defaultMaxEventStates
+      ) {
+        ctx.error(
+          'non-default max_event_states when attribution_scope_limit is not set'
```
linnan-github, line 1401: `max_event_states` in the message is probably redundant given it's the context.

index.bs
apasel422: Please add a non-normative note explaining that below this algorithm, as it is not obvious.

ts/src/header-validator/validate-json.ts
```diff
@@ -1052,94 +1060,129 @@ export type Source = CommonDebug &
...
+            'attribution_scopes',
+            withDefault(
+              (j) =>
+                attributionScopes(
+                  ctx,
+                  j,
+                  attributionScopeLimitVal.value !== null ? 1 : 0,
+                  Math.min(
+                    constants.maxAttributionScopesPerSource,
+                    attributionScopeLimitVal.value || 0
```
apasel422, line 1163:
```suggestion
                    attributionScopeLimitVal.value ?? 0
```

ts/src/header-validator/source.test.ts
```diff
@@ -2544,6 +2559,223 @@ const testCases: TestCase[] = [
...
+  {
+    name: 'missing-attribution-scope-limit-max-event-states',
+    json: `{
+      "destination": "https://a.test",
+      "max_event_states": 5
+    }`,
+    expectedErrors: [
+      {
+        path: ['max_event_states'],
+        msg: 'non-default max_event_states when attribution_scope_limit is not set',
```
apasel422, line 2652: Can we rephrase this to better indicate what the error actually is? It's unclear if the `when` means that this *is* the current situation, or that that *should be* the situation.

ts/src/header-validator/source.test.ts
```diff
@@ -2544,6 +2559,223 @@ const testCases: TestCase[] = [
...
+  {
+    name: 'missing-attribution-scope-limit-attribution-scopes',
+    json: `{
+      "destination": "https://a.test",
+      "attribution_scopes": ["1", "2"]
+    }`,
+    expectedErrors: [
+      {
+        path: ['attribution_scopes'],
+        msg: 'length must be in the range [0, 0]',
```
apasel422, line 2639: We may want to add an explanation to this message, as it isn't obvious why the list has to be empty. In particular, we may want to hint that the `attribution_scope_limit` field should be added to avoid this.

ts/src/header-validator/source.test.ts
```diff
@@ -2544,6 +2559,223 @@ const testCases: TestCase[] = [
...
+    name: 'attribution-scopes-size-exceeds-attribution-scope-limit',
+    json: `{
+      "destination": "https://a.test",
+      "attribution_scope_limit": 1,
+      "attribution_scopes": ["1", "2"]
+    }`,
+    expectedErrors: [
+      {
+        path: ['attribution_scopes'],
+        msg: 'length must be in the range [1, 1]',
```
apasel422, line 2708: We may want to indicate that the second `1` in this message is derived from the scope limit field.

ts/src/header-validator/source.test.ts
```diff
@@ -2544,6 +2559,223 @@ const testCases: TestCase[] = [
...
+      },
+    ],
+  },
+  {
+    name: 'attribution-scope-not-string',
+    json: `{
+      "destination": "https://a.test",
+      "attribution_scope_limit": 2,
+      "attribution_scopes": [1, 2]
```
apasel422, line 2717:
```suggestion
      "attribution_scopes": [1]
```
For simplicity.

ts/src/header-validator/trigger.test.ts
```diff
@@ -1626,6 +1628,40 @@ const testCases: jsontest.TestCase<Trigger>[] = [
     ],
   },
 
+  // Attribution Scope
+  {
+    name: 'attribution-scope-not-string',
+    json: `{"attribution_scopes": [1, 2]}`,
```
apasel422, line 1634:
```suggestion
    json: `{"attribution_scopes": [1]}`,
```
For simplicity.

apasel422 (changes requested):

ts/src/header-validator/validate-json.ts
feifeiji89: npm run pretty added this :)

index.bs
```diff
@@ -2730,6 +2730,9 @@ To <dfn>parse attribution scopes for source</dfn> from a [=map=] |map| and a pos
 1. If |attributionScopeLimit| is null:
     1. If |result| is not [=set/is empty|empty=], return an error.
     1. Return |result|.
+
+Note: Empty attribution scopes are not allowed to prevent the selection of both buckets with and without scopes when |attributionScopeLimit| is set to 1, which would effectively result in two scopes.
```
apasel422, line 2734: `bucket` seems to be a novel term here, and is easy to confuse with the same term from aggregatable attribution. Do we mean `sources` instead?

apasel422 (changes requested):

index.bs
feifeiji89: Done.

ts/src/header-validator/source.test.ts
feifeiji89: refactored the code a bit to make error message more specific

index.bs
```diff
@@ -2687,6 +2709,36 @@ non-negative integer |defaultBudget|, and an [=aggregatable debug reporting conf
...
+    1. If |result| is not [=set/is empty|empty=], return an error.
+    1. Return |result|.
+1. If |result| [=set/is empty=], return an error.
+1. If |result|'s [=set/size=] is greater than |attributionScopeLimit|, return an error.
+1. If |result|'s [=set/size=] is greater than [=max attribution scopes per source=], return an error.
+1. Return |result|.
+
+Note: Empty attribution scopes are not allowed if |attributionScopeLimit| is set,
+to prevent the selection of both buckets with and without scopes 
+if |attributionScopeLimit| is set to 1, which would effectively result in two scopes.
```
linnan-github, line 2740: I don't think the issue only applies to `|attributionScopeLimit|` being 1. Instead, it applies to any value as that effectively increases the number of scopes by 1.

index.bs
feifeiji89: Good point. Done.

ts/src/header-validator/validate-json.ts
```diff
@@ -1335,6 +1373,78 @@ function eventTriggerData(
...
+      )
+      return false
+    }
+    return true
+  }
+
+  return set(j, ctx, (j) =>
+    string(j, ctx).filter(attributionScopeStringLength)
+  ).filter((scopes) => {
+    if (attributionScopeLimit === undefined) {
```
linnan-github, line 1422: When would this happen?

feifeiji89: at trigger parsing time.

index.bs
feifeiji89: lmk if it looks better now.

ts/src/header-validator/validate-json.ts
```diff
@@ -1335,6 +1373,78 @@ function eventTriggerData(
...
+): Maybe<number> {
+  return number(j, ctx)
+    .filter(isInteger, ctx)
+    .filter(isInRange, ctx, 1, ctx.vsv.maxTriggerStateCardinality)
+    .filter((n) => {
+      if (
+        attributionScopeLimit.value === null &&
+        n !== constants.defaultMaxEventStates
+      ) {
+        ctx.error('non-default if attribution_scope_limit is not set')
```
linnan-github, line 1395: Maybe `non-default when attribution_scope_limit is not set`?

feifeiji89: good point. Done.

linnan-github: Please update the message to “must be default…” to avoid confusion.

feifeiji89: Done.

linnan-github (approved):

index.bs
```diff
@@ -2687,6 +2709,36 @@ non-negative integer |defaultBudget|, and an [=aggregatable debug reporting conf
...
+1. If |attributionScopeLimit| is null:
+    1. If |result| is not [=set/is empty|empty=], return an error.
+    1. Return |result|.
+1. If |result| [=set/is empty=], return an error.
+1. If |result|'s [=set/size=] is greater than |attributionScopeLimit|, return an error.
+1. If |result|'s [=set/size=] is greater than [=max attribution scopes per source=], return an error.
+1. Return |result|.
+
+Note: Empty attribution scopes are not allowed if |attributionScopeLimit| is set,
+to prevent the selection of both sources with and without scopes which would effectively
```
apasel422, line 2739:
```suggestion
to prevent the selection of both sources with and without scopes, which would effectively
```

apasel422 (approved):

apasel422 (approved):

PR #1364, opened 2024-07-09
Information gain computation logic for Attribution Scope.

feifeiji89:
The current attribution logic in the Attribution Reporting API may not be ideal for use-cases where an ad-tech needs more fine grain control over the attribution granularity (i.e. campaign, product, conversion ID, etc.) before a source is chosen for attribution. Currently available features such as top-level filters are not fully sufficient for this use-case because they happen after a source has been selected (i.e. after attribution). We can optionally support this use-case by allowing callers of the API to specify predefined attribution scopes that will be considered for filtering before attributing a source, in order to more efficiently extract utility out of the API.

This PR covers the information gain computation logic:
Given event report configuration `(c, b, w)` definition:
- the maximum number `c` of conversions per impression
- the number `b` of bits per conversion
- the number `w` of reporting windows

To accommodate our default configuration navigation=(3,3,3) and event=(1,1,1), we propose introduction of a new information gain limit of 11.55 for attribution scope attack vector (for navigation sources) specifically because the upper bound doesn’t include source noise considerations.

***
<a href="https://pr-preview.s3.amazonaws.com/feifeiji89/attribution-reporting-api/pull/1364.html" title="Last updated on Aug 15, 2024, 4:25 PM UTC (cdd2ffe)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/attribution-reporting-api/1364/1dbcd91...feifeiji89:cdd2ffe.html" title="Last updated on Aug 15, 2024, 4:25 PM UTC (cdd2ffe)">Diff</a>

PR #1215, opened 2024-03-25
Pre-Attribution Filtering: Attribution Scope

feifeiji89:
The current attribution logic in the Attribution Reporting API may not be ideal for use-cases where an ad-tech needs more fine grain control over the attribution granularity (i.e. campaign, product, conversion ID, etc.) before a source is chosen for attribution. Currently available features such as top-level filters are not fully sufficient for this use-case because they happen after a source has been selected (i.e. after attribution). We can optionally support this use-case by allowing callers of the API to specify predefined attribution scopes that will be considered for filtering before attributing a source, in order to more efficiently extract utility out of the API.

***
<a href="https://pr-preview.s3.amazonaws.com/feifeiji89/attribution-reporting-api/pull/1215.html" title="Last updated on Jun 26, 2024, 7:32 PM UTC (7af91de)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/attribution-reporting-api/1215/bacfb84...feifeiji89:7af91de.html" title="Last updated on Jun 26, 2024, 7:32 PM UTC (7af91de)">Diff</a>

PR #1365, opened 2024-07-09
Attribution and source processing logic for Attribution scope.

feifeiji89:
The current attribution logic in the Attribution Reporting API may not be ideal for use-cases where an ad-tech needs more fine grain control over the attribution granularity (i.e. campaign, product, conversion ID, etc.) before a source is chosen for attribution. Currently available features such as top-level filters are not fully sufficient for this use-case because they happen after a source has been selected (i.e. after attribution). We can optionally support this use-case by allowing callers of the API to specify predefined attribution scopes that will be considered for filtering before attributing a source, in order to more efficiently extract utility out of the API.

This PR covers the attribution and source processing logic.

***
<a href="https://pr-preview.s3.amazonaws.com/feifeiji89/attribution-reporting-api/pull/1365.html" title="Last updated on Aug 13, 2024, 6:53 PM UTC (ec6d054)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/attribution-reporting-api/1365/d22686c...feifeiji89:ec6d054.html" title="Last updated on Aug 13, 2024, 6:53 PM UTC (ec6d054)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -3116,6 +3116,8 @@ and a [=boolean=] |isNoised|:
         : "<code>[=source debug data type/source-reporting-origin-per-site-limit=]</code>"
         :: [=map/Set=] |body|["`limit`"] to the user agent's [=max source reporting origins per source reporting site=],
             [=serialize an integer|serialized=].
+        : "<code>[=source debug data type/source-max-event-states-limit=]</code>"
+        :: [=map/Set=] |body|["`limit`"] to the |source|'s [=attribution source/attribution scopes=]'s [=attribution scopes/max event states=].
 
         </dl>
 1. [=Assert=]: |dataTypeToReport| is not null.
--- a/index.bs
+++ b/index.bs
@@ -3193,7 +3193,7 @@ To <dfn>remove associated event-level reports and rate-limit records</dfn> given
     1. If |report|'s [=event-level report/source identifier=] is not equal to |sourceId|, [=iteration/continue=].
     1. If |report|'s [=event-level report/trigger time=] is less than |minTriggerTime|, [=iteration/continue=].
     1. [=set/Remove=] |report| from the [=event-level report cache=].
-    1. [=list/Remove=] the [=attribution rate-limit records=] |entry| from the [=attribution rate-limit cache=] if |entry|'s [=attribution rate-limit record/entity ID=] is equal to |report|'s [=event-level report/report ID=].
+    1. [=list/Remove=] all [=attribution rate-limit records=] |entry| from the [=attribution rate-limit cache=] where |entry|'s [=attribution rate-limit record/entity ID=] is equal to |report|'s [=event-level report/report ID=].
 
 To <dfn>remove sources with unselected attribution scopes for destination</dfn> given a [=site=] |destination| and an [=attribution source=] |pendingSource|:
 1. Let |scopeRecords| be a new [=list=].
--- a/index.bs
+++ b/index.bs
@@ -328,6 +328,8 @@ add the property
 Issue: Use/propagate [=navigation params/navigationSourceEligible=] to the
 [=navigation request=]'s [=request/Attribution Reporting eligibility=].
 
+Issue: Enforce [attribution-scope privacy limits](#attribution-scope).
+
 # Network monkeypatches # {#network-monkeypatches}
 
 <pre class="idl">
@@ -1162,6 +1164,7 @@ Possible values are:
 <li>"<dfn><code>source-destination-limit-replaced</code></dfn>"
 <li>"<dfn><code>source-destination-per-day-rate-limit</code></dfn>"
 <li>"<dfn><code>source-destination-rate-limit</code></dfn>"
+<li>"<dfn><code>source-max-event-states-limit</code></dfn>"
 <li>"<dfn><code>source-noised</code></dfn>"
 <li>"<dfn><code>source-reporting-origin-limit</code></dfn>"
 <li>"<dfn><code>source-reporting-origin-per-site-limit</code></dfn>"
@@ -3168,6 +3171,73 @@ To <dfn>delete expired sources</dfn> given a [=moment=] |now|:
     1. If |source|'s [=attribution source/expiry time=] is less than |now|,
         [=set/remove=] |source| from the [=attribution source cache=].
 
+To <dfn>find sources with common destinations and reporting origin</dfn> given an [=attribution source=] |pendingSource|:
+1. Let |matchingSources| be a new [=list=].
+1. [=set/iterate|For each=] |source| of the user agent's [=attribution source cache=]:
+    1. Let |commonDestinations| be the [=set/intersection=] of |source|'s [=attribution source/attribution destinations=] and |pendingSource|'s [=attribution source/attribution destinations=].
+    1. If |commonDestinations| [=set/is empty=], [=iteration/continue=].
+    1. If |source|'s [=attribution source/reporting origin=] and |pendingSource|'s [=attribution source/reporting origin=] are not [=same origin=], [=iteration/continue=].
+    1. [=list/Append=] |source| to |matchingSources|.
+1. Return |matchingSources|.
+
+To <dfn>check if source has incompatible attribution scope fields</dfn> given an [=attribution source=] |source| and an [=attribution source=] |pendingSource|:
+1. If |pendingSource|'s [=attribution source/attribution scopes=] is null:
+    1. Return true if |source|'s [=attribution source/attribution scopes=] is null, false otherwise.
+1. If |source|'s [=attribution source/attribution scopes=] is null, return true.
+1. If |source|'s [=attribution source/attribution scopes=]'s [=attribution scopes/max event states=] is not equal to |pendingSource|'s [=attribution source/attribution scopes=]'s [=attribution scopes/max event states=], return true.
+1. If |source|'s [=attribution source/attribution scopes=]'s [=attribution scopes/limit=] is less than |pendingSource|'s [=attribution source/attribution scopes=]'s [=attribution scopes/limit=], return true.
+1. Return false.
+
+To <dfn>remove associated event-level reports and rate-limit records</dfn> given an [=attribution source/source identifier=] |sourceId| and a [=moment=] |minTriggerTime|:
+1. [=set/iterate|For each=] [=event-level report=] |report| of the [=event-level report cache=]:
+    1. If |report|'s [=event-level report/source identifier=] is not equal to |sourceId|, [=iteration/continue=].
+    1. If |report|'s [=event-level report/trigger time=] is less than |minTriggerTime|, [=iteration/continue=].
+    1. [=set/Remove=] |report| from the [=event-level report cache=].
+    1. [=list/Remove=] the [=attribution rate-limit records=] |entry| from the [=attribution rate-limit cache=] if |entry|'s [=attribution rate-limit record/entity ID=] is equal to |report|'s [=event-level report/report ID=].
+
+To <dfn>remove sources with unselected attribution scopes for destination</dfn> given a [=site=] |destination| and an [=attribution source=] |pendingSource|:
+1. Let |scopeRecords| be a new [=list=].
+1. Let |scopes| be |pendingSource|'s [=attribution source/attribution scopes=]'s [=attribution scopes/values=].
+1. [=set/iterate|For each=] |scope| in |scopes|:
+    1. [=list/Append=] the [=tuple=] (|scope|, |pendingSource|) to |scopeRecords|.
+1. [=set/iterate|For each=] |source| of the [=attribution source cache=]:
+    1. If |source|'s [=attribution source/reporting origin=] and |pendingSource|'s [=attribution source/reporting origin=] are not [=same origin=], [=iteration/continue=].
+    1. If |source|'s [=attribution source/attribution destinations=] does not [=set/contain=] |destination|, [=iteration/continue=].
+    1. If |source|'s [=attribution source/attribution scopes=]'s is null, [=iteration/continue=].
+    1. [=set/iterate|For each=] |scope| in |source|'s [=attribution source/attribution scopes=]'s [=attribution scopes/values=]:
+        1. [=list/Append=] the [=tuple=] (|scope|, |source|) to |scopeRecords|.
+1. [=list/sort in ascending order|Sort=] |scopeRecords| in ascending order with |a| being less than |b| if any of the following are true:
+      * |a|[1]'s [=attribution source/source time=] is greater than |b|[1]'s [=attribution source/source time=].
+      * |a|[1]'s [=attribution source/source time=] is equal to |b|[1]'s [=attribution source/source time=] and |a|[0] is greater than |b|[0].
+1. Let |selectedScopes| be |scopes|.
+1. Let |sourcesToRemove| be a new [=set=].
+1. [=list/iterate|For each=] |record| of |scopeRecords|:
+    1. If |selectedScopes|'s [=set/size=] is less than |pendingSource|'s [=attribution source/attribution scopes=]'s [=attribution scopes/limit=], [=set/append=] |record|[0] to |selectedScopes|.
+    1. Otherwise, if |selectedScopes| does not [=set/contain=] |record|[0], [=set/append=] |record|[1] to |sourcesToRemove|.
+1. [=set/iterate|For each=] |source| of the |sourcesToRemove|:
+    1. [=Remove associated event-level reports and rate-limit records=] with |source|'s [=attribution source/source identifier=] and |pendingSource|'s [=attribution source/source time=].
+    1. [=set/Remove=] |source| from the [=attribution source cache=].
+
+To <dfn>remove sources with unselected attribution scopes</dfn> given an [=attribution source=] |pendingSource|:
+1. If |pendingSource|'s [=attribution source/attribution scopes=] is null, return.
+1. [=Assert=]: |pendingSource|'s [=attribution source/attribution destinations=] is [=list/sort in ascending order|sorted=] in
+    ascending order, with |a| being less than |b| if |a|,
+    <a href="https://html.spec.whatwg.org/multipage/origin.html#serialization-of-a-site">serialized</a>,
+    is less than |b|, <a href="https://html.spec.whatwg.org/multipage/origin.html#serialization-of-a-site">serialized</a>.
+1. [=set/iterate|For each=] |destination| in |pendingSource|'s [=attribution source/attribution destinations=]:
+    1. [=Remove sources with unselected attribution scopes for destination=] with |destination| and |pendingSource|.
+
+To <dfn>remove or update sources with incompatible attribution scope fields</dfn> given an [=attribution source=] |pendingSource|:
+1. Let |matchingSources| be the result of running [=find sources with common destinations and reporting origin=] with |pendingSource|.
+1. [=list/iterate|For each=] |source| of |matchingSources|:
+    1. If the result of running [=check if source has incompatible attribution scope fields=] with |source| and |pendingSource| is false, [iteration/continue=].
+    1. If |pendingSource|'s [=attribution source/attribution scopes=] is null and |source|'s [=attribution source/attribution scopes=] is not null:
+        1. Set |source|'s [=attribution source/attribution scopes=] to null.
+    1. Otherwise:
+        1. [=Remove associated event-level reports and rate-limit records=] with |source|'s [=attribution source/source identifier=] and |pendingSource|'s [=attribution source/source time=].
+        1. [=set/Remove=] |source| from the [=attribution source cache=].
+1. [=Remove sources with unselected attribution scopes=] with |pendingSource|.
+
 To <dfn>process an attribution source</dfn> given an [=attribution source=] |source|:
 
 1. [=Delete expired sources=] with |source|'s [=attribution source/source time=].
@@ -3196,6 +3266,10 @@ To <dfn>process an attribution source</dfn> given an [=attribution source=] |sou
 1. Set |source|'s [=attribution source/number of event-level reports=] to 0 if
     |source|'s [=attribution source/randomized response=] is null,
     [=attribution source/randomized response=]'s [=list/size=] otherwise.
+1. Let |states| be the [=obtain a set of possible trigger states|number of possible trigger states=] obtained with |randomizedResponseConfig|.
+1. If |sourceType| is "<code>[=source type/event=]</code>" and |states| is greater than |source|'s [=attribution source/attribution scopes=]'s [=attribution scopes/max event states=]:
+    1. Run [=obtain and deliver debug reports on source registration=] with "<code>[=source debug data type/source-max-event-states-limit=]</code>" and |source|.
+    1. Return.
 1. Let |pendingSourcesForSourceOrigin| be the [=set=] of all
     [=attribution sources=] |pendingSource| of the [=attribution source cache=] where |pendingSource|'s
     [=attribution source/source origin=] and |source|'s
@@ -3226,6 +3300,7 @@ To <dfn>process an attribution source</dfn> given an [=attribution source=] |sou
     [=set/append=] "<code>[=source debug data type/source-destination-limit-replaced=]</code>"
     to |debugDataTypes|.
 1. Run [=delete sources for unexpired destination limit=] with |sourcesToDeleteForDestinationLimit| and |source|'s [=attribution source/source time=].
+1. [=Remove or update sources with incompatible attribution scope fields=] with |source|.
 1. Let |isNoised| be true if |source|'s [=attribution source/randomized response=]
     is not null, otherwise false.
 1. If |destinationRateLimitResult| is "<code>[=destination rate-limit result/hit global limit=]</code>":
@@ -4176,6 +4251,11 @@ and an optional [=attribution source=]
 1. Run [=obtain and deliver an aggregatable debug report on trigger registration=]
     with |dataSet|, |trigger|, and |sourceToAttribute|.
 
+To <dfn>check if an [=attribution source=] and [=attribution trigger=] have matching attribution scopes</dfn> given an [=attribution source=] |source| and an [=attribution trigger=] |trigger|:
+
+1. If |trigger|'s [=attribution trigger/attribution scopes=] [=set/is empty=], return true.
+1. Return whether the [=set/intersection=] of |source|'s [=attribution source/attribution scopes=]'s [=attribution scopes/values=] and |trigger|'s [=attribution trigger/attribution scopes=] is not [=set/is empty|empty=].
+
 To <dfn>find matching sources</dfn> given an [=attribution trigger=] |trigger|:
 
 1. Let |matchingSources| be a new [=list=].
@@ -4189,7 +4269,12 @@ To <dfn>find matching sources</dfn> given an [=attribution trigger=] |trigger|:
       * |a|'s [=attribution source/priority=] is less than |b|'s [=attribution source/priority=].
       * |a|'s [=attribution source/priority=] is equal to |b|'s [=attribution source/priority=] and |a|'s
          [=attribution source/source time=] is less than |b|'s [=attribution source/source time=].
+      * the result of running [=check if an attribution source and attribution trigger have matching attribution scopes=] with |a| and |trigger| is false
+        and the result of running [=check if an attribution source and attribution trigger have matching attribution scopes=] with |b| and |trigger| is true
+1. If |matchingSources| is not [=list/is empty|empty=] and the result of running [=check if an attribution source and attribution trigger have matching attribution scopes=] with |matchingSources|[0] and |trigger| is false, return a new [=list=].
 1. Return |matchingSources|.
+
+Note: Sources with matching attribution scopes from the trigger will be prioritized. If no source matches attribution scopes from the trigger, an empty list will be returned.
 
 To <dfn>check if an [=attribution trigger=] contains aggregatable data</dfn> given an [=attribution trigger=] |trigger|,
 run the following steps:
@@ -5281,3 +5366,13 @@ Possible mitigations:
     presence from the reporting origin. Compared to the previous mitigation, the
     proxy server could itself handle the [=event-level report/trigger priority=]
     functionality, at the cost of increased complexity in the proxy.
+
+## Attribution scope ## {#attribution-scope}
+
+*This section is non-normative.*
+
+It is possible for an adversary to register multiple [=source type/navigation=] sources in response to a single navigation, and use these multiple sources, each with a different non-empty [=attribution source/attribution scopes=] value, to gain additional information about a user based on which attribution scope is chosen. To prevent this abuse the number of unique attribution scope sets per reporting origin per navigation needs to be limited.
+
+Proposed mitigation:
+
+Limit to 1 unique non-empty attribution scope set per reporting origin per navigation. Extraneous sources will be dropped.
--- a/index.bs
+++ b/index.bs
@@ -3269,7 +3269,7 @@ To <dfn>process an attribution source</dfn> given an [=attribution source=] |sou
     |source|'s [=attribution source/randomized response=] is null,
     [=attribution source/randomized response=]'s [=list/size=] otherwise.
 1. Let |states| be the [=obtain a set of possible trigger states|number of possible trigger states=] obtained with |randomizedResponseConfig|.
-1. If |sourceType| is "<code>[=source type/event=]</code>" and |states| is greater than |source|'s [=attribution source/attribution scopes=]'s [=attribution scopes/max event states=]:
+1. If |source|'s [=attribution source/attribution scopes=] is not null and |sourceType| is "<code>[=source type/event=]</code>" and |states| is greater than |source|'s [=attribution source/attribution scopes=]'s [=attribution scopes/max event states=]:
     1. Run [=obtain and deliver debug reports on source registration=] with "<code>[=source debug data type/source-max-event-states-limit=]</code>" and |source|.
     1. Return.
 1. Let |pendingSourcesForSourceOrigin| be the [=set=] of all
--- a/index.bs
+++ b/index.bs
@@ -3117,7 +3117,9 @@ and a [=boolean=] |isNoised|:
         :: [=map/Set=] |body|["`limit`"] to the user agent's [=max source reporting origins per source reporting site=],
             [=serialize an integer|serialized=].
         : "<code>[=source debug data type/source-max-event-states-limit=]</code>"
-        :: [=map/Set=] |body|["`limit`"] to the |source|'s [=attribution source/attribution scopes=]'s [=attribution scopes/max event states=].
+        ::
+            1. [=Assert=]: |source|'s [=attribution source/attribution scopes=] is not null.
+            1. [=map/Set=] |body|["`limit`"] to the |source|'s [=attribution source/attribution scopes=]'s [=attribution scopes/max event states=].
 
         </dl>
 1. [=Assert=]: |dataTypeToReport| is not null.
--- a/index.bs
+++ b/index.bs
@@ -3230,8 +3230,8 @@ To <dfn>remove sources with unselected attribution scopes</dfn> given an [=attri
 To <dfn>remove or update sources with incompatible attribution scope fields</dfn> given an [=attribution source=] |pendingSource|:
 1. Let |matchingSources| be the result of running [=find sources with common destinations and reporting origin=] with |pendingSource|.
 1. [=list/iterate|For each=] |source| of |matchingSources|:
-    1. If the result of running [=check if source has incompatible attribution scope fields=] with |source| and |pendingSource| is false, [iteration/continue=].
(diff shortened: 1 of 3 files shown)


index.bs
```diff
@@ -3212,6 +3215,69 @@ To <dfn>delete expired sources</dfn> given a [=moment=] |now|:
...
+1. [=set/iterate|For each=] |source| of the user agent's [=attribution source cache=]:
+    1. Let |commonDestinations| be the [=set/intersection=] of |source|'s [=attribution source/attribution destinations=] and |pendingSource|'s [=attribution source/attribution destinations=].
+    1. If |commonDestinations| [=set/is empty=], [=iteration/continue=].
+    1. If |source|'s [=attribution source/reporting origin=] and |pendingSource|'s [=attribution source/reporting origin=] are not [=same origin=], [=iteration/continue=].
+    1. [=list/Append=] |source| to |matchingSources|.
+1. Return |matchingSources|.
+
+To <dfn>check if source has incompatible attribution scope fields</dfn> given an [=attribution source=] |source| and an [=attribution source=] |pendingSource|:
+1. If |pendingSource|'s [=attribution source/attribution scope limit=] is null:
+    1. [=Assert=]: |source|' [=attribution source/attribution scope limit=] is null.
```
linnan-github, line 3229: The `Assert` here seems a little confusing. Wdyt about reorganizing the check as the follows:

1.  In `remove or update sources with incompatible attribution scope fields`:
```
    1. If the result of running [=check if source has incompatible attribution scope fields=] with |source| and |pendingSource| is false, [iteration/continue=].
    1. If |pendingSource|'s [=attribution source/attribution scope limit=] is null and |source|'s [=attribution source/attribution scope limit=] is not null:
        1. Set |source|'s [=attribution source/attribution scope limit=] to null.
        1. Set |source|'s [=attribution source/attribution scopes=] to a new [=set=].
        1. Set |source|'s [=attribution source/max event states=] to [=default max event states=].
    1. Otherwise:
        1. [=Remove associated event reports and rate-limit records=] with |source|'s [=attribution source/source identifier=] and |pendingSource|'s [=attribution source/source time=].
        1. [=set/Remove=] |source| from the [=attribution source cache=].
```
2. Return  whether `|source|'s [=attribution source/attribution scope limit=] is null` in this algorithm.

index.bs
```diff
@@ -3212,6 +3215,69 @@ To <dfn>delete expired sources</dfn> given a [=moment=] |now|:
...
+1. If |source|'s [=attribution source/max event states=] is not equal to |pendingSource|'s [=attribution source/max event states=], return true.
+1. If |source|'s [=attribution source/attribution scope limit=] is less than |pendingSource|'s [=attribution source/attribution scope limit=], return true.
+1. Return false.
+
+To <dfn>remove associated event reports and rate-limit records</dfn> given an [=attribution source/source identifier=] |sourceId| and a [=moment=] |minTriggerTime|:
+1. [=set/iterate|For each=] [=event-level report=] |report| of the [=event-level report cache=]:
+    1. If |report|'s [=event-level report/source identifier=] is not equal to |sourceId|, [=iteration/continue=].
+    1. If |report|'s [=event-level report/trigger time=] is less than |minTriggerTime|, [=iteration/continue=].
+    1. [=set/Remove=] |report| from the [=event-level report cache=].
+    1. [=list/Remove=] all elements from [=attribution rate-limit records=] associated with |report|.
```
linnan-github, line 3241: This can remove at most 1 element right?

index.bs
```diff
@@ -3212,6 +3215,69 @@ To <dfn>delete expired sources</dfn> given a [=moment=] |now|:
...
+
+To <dfn>remove sources with attribution scopes not selected for destination</dfn> given a [=site=] |destination| and an [=attribution source=] |pendingSource|:
```
linnan-github, line 3243: Copied from https://github.com/WICG/attribution-reporting-api/pull/1215#discussion_r1644945378.

The current algorithm doesn't guarantee that the attribution scopes of the pending source are always selected.

index.bs
```diff
@@ -3249,6 +3319,7 @@ To <dfn>process an attribution source</dfn> given an [=attribution source=] |sou
     1. Run [=obtain and deliver debug reports on source registration=] with
         « "<code>[=source debug data type/source-storage-limit=]</code>" » and |source|.
     1. Return.
+1. [=Remove or update sources with incompatible attribution scope fields=] with |source|.
```
linnan-github, line 3322: Shouldn't this be checked after `delete sources for unexpired destination limit`?

index.bs
```diff
@@ -5453,3 +5532,13 @@ Possible mitigations:
...
+
+## Attribution scope ## {#attribution-scope}
+
+*This section is non-normative.*
+
+It is possible for an adversary to register multiple [=source type/navigation=] sources in response to a single navigation, and use these multiple sources, each with a different non-empty [=attribution source/attribution scopes=] value, to gain additional information about a user based on which attribution scope is chosen. To prevent this abuse the number of unique attribution scope sets per reporting origin per navigation needs to be limited.
+
+Proposed mitigation:
+
+Limit to 1 unique attribution scope set per reporting origin per navigation. Extraneous sources will be dropped.
```
linnan-github, line 5544:
```suggestion
Limit to 1 unique non-empty attribution scope set per reporting origin per navigation. Extraneous sources will be dropped.
```

linnan-github (changes requested):

index.bs
```diff
@@ -3212,6 +3215,69 @@ To <dfn>delete expired sources</dfn> given a [=moment=] |now|:
...
+1. [=list/iterate|For each=] |record| of |attributionScopeRecords|:
+    1. If |selectedAttributionScopes|'s [=set/size=] is less than |pendingSource|'s [=attribution source/attribution scope limit=], [=set/append=] |record|[0] to |selectedAttributionScopes|.
+    1. Otherwise, if |selectedAttributionScopes| does not [=set/contain=] |record|[0], [=set/append=] |record|[1] to |sourcesToRemove|.
+1. [=set/iterate|For each=] |source| of the |sourcesToRemove|:
+    1. [=Remove associated event reports and rate-limit records=] with |source|'s [=attribution source/source identifier=] and |pendingSource|'s [=attribution source/source time=].
+    1. [=set/Remove=] |source| from the [=attribution source cache=].
+
+To <dfn>remove sources with attribution scopes not selected</dfn> given an [=attribution source=] |pendingSource|:
+1. If |pendingSource|'s [=attribution source/attribution scope limit=] is null, return.
+1. [=set/iterate|For each=] |destination| in |pendingSource|'s [=attribution source/attribution destinations=]:
```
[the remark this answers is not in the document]
feifeiji89: Yes, that's the desired behavior. The order we process the destinations can cause slightly different sources being deactivated. Added `Assert` statement to make sure the destinations are sorted.

index.bs
feifeiji89: Done. changed wording.

index.bs
feifeiji89: Good point. Done.

index.bs
feifeiji89: Done.

index.bs
```diff
@@ -1141,6 +1143,7 @@ Possible values are:
 <li>"<dfn><code>source-destination-limit-replaced</code></dfn>"
 <li>"<dfn><code>source-destination-per-day-rate-limit</code></dfn>"
 <li>"<dfn><code>source-destination-rate-limit</code></dfn>"
+<li>"<dfn><code>source-max-event-states-limit</code></dfn>"
```
apasel422, line 1146: This needs to be added to https://github.com/WICG/attribution-reporting-api/blob/590c74ec584a22268f6c621a1e3d162486421a04/ts/src/constants.ts#L64.

index.bs
```diff
@@ -3153,6 +3156,73 @@ To <dfn>delete expired sources</dfn> given a [=moment=] |now|:
...
+1. [=set/iterate|For each=] |source| of the user agent's [=attribution source cache=]:
+    1. Let |commonDestinations| be the [=set/intersection=] of |source|'s [=attribution source/attribution destinations=] and |pendingSource|'s [=attribution source/attribution destinations=].
+    1. If |commonDestinations| [=set/is empty=], [=iteration/continue=].
+    1. If |source|'s [=attribution source/reporting origin=] and |pendingSource|'s [=attribution source/reporting origin=] are not [=same origin=], [=iteration/continue=].
+    1. [=list/Append=] |source| to |matchingSources|.
+1. Return |matchingSources|.
+
+To <dfn>check if source has incompatible attribution scope fields</dfn> given an [=attribution source=] |source| and an [=attribution source=] |pendingSource|:
+1. If |pendingSource|'s [=attribution source/attribution scope limit=] is null:
+    1. Return |source|'s [=attribution source/attribution scope limit=] is null.
```
apasel422, line 3170:
```suggestion
    1. Return true if |source|'s [=attribution source/attribution scope limit=] is null, false otherwise.
```

index.bs
```diff
@@ -3153,6 +3156,73 @@ To <dfn>delete expired sources</dfn> given a [=moment=] |now|:
...
+1. If |source|'s [=attribution source/attribution scope limit=] is null, return true.
+1. If |source|'s [=attribution source/max event states=] is not equal to |pendingSource|'s [=attribution source/max event states=], return true.
+1. If |source|'s [=attribution source/attribution scope limit=] is less than |pendingSource|'s [=attribution source/attribution scope limit=], return true.
+1. Return false.
+
+To <dfn>remove associated event reports and rate-limit records</dfn> given an [=attribution source/source identifier=] |sourceId| and a [=moment=] |minTriggerTime|:
```
apasel422, line 3176:
```suggestion
To <dfn>remove associated event-level reports and rate-limit records</dfn> given an [=attribution source/source identifier=] |sourceId| and a [=moment=] |minTriggerTime|:
```

index.bs
```diff
@@ -3153,6 +3156,73 @@ To <dfn>delete expired sources</dfn> given a [=moment=] |now|:
...
+1. [=set/iterate|For each=] [=event-level report=] |report| of the [=event-level report cache=]:
+    1. If |report|'s [=event-level report/source identifier=] is not equal to |sourceId|, [=iteration/continue=].
+    1. If |report|'s [=event-level report/trigger time=] is less than |minTriggerTime|, [=iteration/continue=].
+    1. [=set/Remove=] |report| from the [=event-level report cache=].
+    1. [=list/Remove=] the element from [=attribution rate-limit records=] associated with |report|.
```
apasel422, line 3181: This is a bit vague. Please describe exactly which elements are associated with `|report|`.

index.bs
```diff
@@ -3153,6 +3156,73 @@ To <dfn>delete expired sources</dfn> given a [=moment=] |now|:
...
+1. [=list/sort in ascending order|Sort=] |attributionScopeRecords| in ascending order with |a| being less than |b| if any of the following are true:
+      * |a|[1]'s [=attribution source/source time=] is greater than |b|[1]'s [=attribution source/source time=].
+      * |a|[1]'s [=attribution source/source time=] is equal to |b|[1]'s [=attribution source/source time=] and |a|[0] is greater than |b|[0].
+1. Let |selectedAttributionScopes| be |pendingSource|'s [=attribution source/attribution scopes=].
+1. Let |sourcesToRemove| be a new [=set=].
+1. [=list/iterate|For each=] |record| of |attributionScopeRecords|:
+    1. If |selectedAttributionScopes|'s [=set/size=] is less than |pendingSource|'s [=attribution source/attribution scope limit=], [=set/append=] |record|[0] to |selectedAttributionScopes|.
+    1. Otherwise, if |selectedAttributionScopes| does not [=set/contain=] |record|[0], [=set/append=] |record|[1] to |sourcesToRemove|.
+1. [=set/iterate|For each=] |source| of the |sourcesToRemove|:
+    1. [=Remove associated event reports and rate-limit records=] with |source|'s [=attribution source/source identifier=] and |pendingSource|'s [=attribution source/source time=].
```
apasel422, line 3201:
```suggestion
    1. [=Remove associated event-level reports and rate-limit records=] with |source|'s [=attribution source/source identifier=] and |pendingSource|'s [=attribution source/source time=].
```

index.bs
```diff
@@ -3153,6 +3156,73 @@ To <dfn>delete expired sources</dfn> given a [=moment=] |now|:
...
+    1. [=set/Remove=] |source| from the [=attribution source cache=].
+
+To <dfn>remove sources with attribution scopes not selected</dfn> given an [=attribution source=] |pendingSource|:
```
apasel422, line 3204: The name of this algorithm is a bit hard to parse. Maybe `remove sources with unselected attribution scopes`?

index.bs
```diff
@@ -3153,6 +3156,73 @@ To <dfn>delete expired sources</dfn> given a [=moment=] |now|:
...
+To <dfn>remove or update sources with incompatible attribution scope fields</dfn> given an [=attribution source=] |pendingSource|:
+1. Let |matchingSources| be the result of running [=find sources with common destinations and reporting origin=] with |pendingSource|.
+1. [=list/iterate|For each=] |source| of |matchingSources|:
+    1. If the result of running [=check if source has incompatible attribution scope fields=] with |source| and |pendingSource| is false, [iteration/continue=].
+    1. If |pendingSource|'s [=attribution source/attribution scope limit=] is null and |source|'s [=attribution source/attribution scope limit=] is not null:
+        1. Set |source|'s [=attribution source/attribution scope limit=] to null.
+        1. Set |source|'s [=attribution source/attribution scopes=] to a new [=set=].
+        1. Set |source|'s [=attribution source/max event states=] to [=default max event states=].
+    1. Otherwise:
+        1. [=Remove associated event reports and rate-limit records=] with |source|'s [=attribution source/source identifier=] and |pendingSource|'s [=attribution source/source time=].
```
apasel422, line 3222:
```suggestion
        1. [=Remove associated event-level reports and rate-limit records=] with |source|'s [=attribution source/source identifier=] and |pendingSource|'s [=attribution source/source time=].
```

index.bs
```diff
@@ -4161,6 +4236,11 @@ and an optional [=attribution source=]
 1. Run [=obtain and deliver an aggregatable debug report on trigger registration=]
     with |dataSet|, |trigger|, and |sourceToAttribute|.
 
+To <dfn>check if an [=attribution source=] and [=attribution trigger=] have matching attribution scopes</dfn> given an [=attribution source=] |source| and [=attribution trigger=] |trigger|:
```
apasel422, line 4239:
```suggestion
To <dfn>check if an [=attribution source=] and [=attribution trigger=] have matching attribution scopes</dfn> given an [=attribution source=] |source| and an [=attribution trigger=] |trigger|:
```

index.bs
```diff
@@ -4174,6 +4254,9 @@ To <dfn>find matching sources</dfn> given an [=attribution trigger=] |trigger|:
       * |a|'s [=attribution source/priority=] is less than |b|'s [=attribution source/priority=].
       * |a|'s [=attribution source/priority=] is equal to |b|'s [=attribution source/priority=] and |a|'s
          [=attribution source/source time=] is less than |b|'s [=attribution source/source time=].
+      * the result of running [=check if an attribution source and attribution trigger have matching attribution scopes=] with |a| and |trigger| is false
```
apasel422, line 4257: Could you clarify, perhaps in a non-normative note, what the intention of this list item is with regard to how matching sources are ordered?

index.bs
```diff
@@ -3153,6 +3156,73 @@ To <dfn>delete expired sources</dfn> given a [=moment=] |now|:
...
+1. Return false.
+
+To <dfn>remove associated event reports and rate-limit records</dfn> given an [=attribution source/source identifier=] |sourceId| and a [=moment=] |minTriggerTime|:
+1. [=set/iterate|For each=] [=event-level report=] |report| of the [=event-level report cache=]:
+    1. If |report|'s [=event-level report/source identifier=] is not equal to |sourceId|, [=iteration/continue=].
+    1. If |report|'s [=event-level report/trigger time=] is less than |minTriggerTime|, [=iteration/continue=].
+    1. [=set/Remove=] |report| from the [=event-level report cache=].
+    1. [=list/Remove=] the element from [=attribution rate-limit records=] associated with |report|.
+
+To <dfn>remove sources with attribution scopes not selected for destination</dfn> given a [=site=] |destination| and an [=attribution source=] |pendingSource|:
```
apasel422, line 3183: The name of this algorithm is a bit hard to parse. Maybe `remove sources with unselected attribution scopes for destination`?

apasel422 (changes requested):

index.bs
```diff
@@ -3181,6 +3251,10 @@ To <dfn>process an attribution source</dfn> given an [=attribution source=] |sou
 1. Set |source|'s [=attribution source/number of event-level reports=] to 0 if
     |source|'s [=attribution source/randomized response=] is null,
     [=attribution source/randomized response=]'s [=list/size=] otherwise.
+1. Let |states| be the [=obtain a set of possible trigger states|number of possible trigger states=] obtained with |randomizedResponseConfig|.
+1. If |sourceType| is "<code>[=source type/event=]</code>" and |states| is greater than |source|'s [=attribution source/max event states=]:
+    1. Run [=obtain and deliver debug reports on source registration=] with "<code>[=source debug data type/source-max-event-states-limit=]</code>" and |source|.
+    1. Return.
```
ThomasQuesadilla, line 3257: Considering that the number of states is calculated in `computing the channel capacity of a source` above as well, would it make more sense to move this to be before calculating channel capacity and pass `|states|` into the computation? That way we prioritize `source-max-event-states-limit` over the channel capacity limit like we do for `source-trigger-state-cardinality-limit`

index.bs
```diff
@@ -3181,6 +3251,10 @@ To <dfn>process an attribution source</dfn> given an [=attribution source=] |sou
 1. Set |source|'s [=attribution source/number of event-level reports=] to 0 if
     |source|'s [=attribution source/randomized response=] is null,
     [=attribution source/randomized response=]'s [=list/size=] otherwise.
+1. Let |states| be the [=obtain a set of possible trigger states|number of possible trigger states=] obtained with |randomizedResponseConfig|.
+1. If |sourceType| is "<code>[=source type/event=]</code>" and |states| is greater than |source|'s [=attribution source/max event states=]:
+    1. Run [=obtain and deliver debug reports on source registration=] with "<code>[=source debug data type/source-max-event-states-limit=]</code>" and |source|.
```
ThomasQuesadilla, line 3256: Do we care about reporting the max event states limit when issuing the `source-max-event-states-limit` debug report for an event type source exceeding it? I figure they'll already have it in hand but I'm not sure if we wanted to be consistent with the `source-trigger-state-cardinality-limit` debug report, which [does include the trigger state cardinality limit](https://wicg.github.io/attribution-reporting-api/#ref-for-source-debug-data-type-source-trigger-state-cardinality-limit).

apasel422: Thinking about this some more, I'm not sure that this new debug report type should exist, as it's effectively a form of parsing error resulting in an internal inconsistency between the source registration's actual number of states and what it has chosen to be the maximum number of states. Maybe we can omit it for now and consider re-adding it in the future once we ponder the best design.

linnan-github: Agreed that it's effectively a form of parsing error, but we check the limit here to reduce the complexity in parsing due to obtaining number of possible trigger states. I think it would still be beneficial to add this new debug report type so that each source registration would continue to create a verbose debug report.

This limit is determined by both the ad-tech and the user agent's default value, and adding the limit to the report    SGTM.

apasel422: > we check the limit here to reduce the complexity in parsing due to obtaining number of possible trigger states

I don't feel strongly, but regardless of where the check is performed, we have the freedom to report the problem using whatever debug type we want.

Compare this with validation of the `event_level_epsilon` field.

index.bs
```diff
@@ -3181,6 +3251,10 @@ To <dfn>process an attribution source</dfn> given an [=attribution source=] |sou
 1. Set |source|'s [=attribution source/number of event-level reports=] to 0 if
     |source|'s [=attribution source/randomized response=] is null,
     [=attribution source/randomized response=]'s [=list/size=] otherwise.
+1. Let |states| be the [=obtain a set of possible trigger states|number of possible trigger states=] obtained with |randomizedResponseConfig|.
+1. If |sourceType| is "<code>[=source type/event=]</code>" and |states| is greater than |source|'s [=attribution source/max event states=]:
```
apasel422, line 3255: Shouldn't we only be performing this check if the scope limit isn't null? Otherwise, we're going to start rejecting registrations with > 3 states that aren't using scopes at all.

apasel422: https://github.com/WICG/attribution-reporting-api/pull/1386

linnan-github: With this change, allowing max event states to 3 without attribution scopes limit doesn't seem to make sense any more. When there's no attribution scopes, we probably should disallow max event states to be set (stricter), or treat it as unknown fields as it's irrelevant (less strict). And max event states should be default to null when there's no attribution scopes, or 3 when there is.

As a side note, it seems to be much easier to wrap the three fields into a dictionary for clearness, but I assume it's too late to make such a change.

apasel422: If we think it makes sense to put the 3 options into a dictionary, I think we can do so.

feifeiji89: > Shouldn't we only be performing this check if the scope limit isn't null? Otherwise, we're going to start rejecting registrations with > 3 states that aren't using scopes at all.

As Nan pointed out, it might not be the best option to wrap the three fields into a dictionary as it will significantly push the Rb release. I see the benefit of that, but am still on the fence whether adding another layer makes it cleaner.

Thanks for catching and fixing this @apasel422

apasel422: > as it will significantly push the Rb release

Should that really be a concern when we're trying to specify something that will be easy to understand by our users?

https://www.w3.org/TR/design-principles/#priority-of-constituencies

feifeiji89: this is resolved in https://github.com/WICG/attribution-reporting-api/pull/1391

feifeiji89: this is resolved in https://github.com/WICG/attribution-reporting-api/pull/1391

linnan-github:
Please resolve the merge conflicts.

attribution_scopes.md
```diff
@@ -221,8 +221,8 @@ Note that the `attribution_scope_limit` does not have any impact on the privacy
 
 ### Additional Privacy Limits
 
-It is possible for an adversary to register multiple navigation sources during source registration, and use these multiple sources, each with a different attribution scopes value, to gain additional information about a user based on which attribution scope is chosen. To prevent this abuse the number of unique attribution scope sets per reporting origin per navigation needs to be limited.
+It is possible for an adversary to register multiple navigation sources during source registration, and use these multiple sources, each with a different attribution scopes value, to gain additional information about a user based on which attribution scope is chosen. To prevent this abuse the number of unique attribution scope sets per reporting origin per destination per navigation needs to be limited.
```
linnan-github, line 224: As discussed, we don’t add destination for this limit for now.

attribution_scopes.md
```diff
@@ -221,8 +221,8 @@ Note that the `attribution_scope_limit` does not have any impact on the privacy
...
 
 TODO: update [parameters](https://github.com/WICG/attribution-reporting-api/blob/main/params/chromium-params.md) table with new rate limit
 
-Strawman: 1 unique attribution scope set per reporting origin per navigation
+Strawman: 1 unique attribution scope set per reporting origin per destination per navigation
```
linnan-github, line 228: Ditto above

index.bs
feifeiji89: this should be addressed in the information gain PR: https://github.com/WICG/attribution-reporting-api/pull/1364/files

index.bs
linnan-github: Per chat offline, we will use a new debug type. @feifeiji89 can you add the limit to the report?

linnan-github (changes requested):

index.bs
feifeiji89: Done.

index.bs
apasel422: Please address this.

apasel422 (changes requested):
Please rebase and address the merge conflicts.

index.bs
feifeiji89: Done. lmk if that looks better now.

index.bs
```diff
@@ -3168,6 +3171,73 @@ To <dfn>delete expired sources</dfn> given a [=moment=] |now|:
...
+1. If |source|'s [=attribution source/attribution scopes=]'s [=attribution scopes/limit=] is less than |pendingSource|'s [=attribution source/attribution scopes=]'s [=attribution scopes/limit=], return true.
+1. Return false.
+
+To <dfn>remove associated event-level reports and rate-limit records</dfn> given an [=attribution source/source identifier=] |sourceId| and a [=moment=] |minTriggerTime|:
+1. [=set/iterate|For each=] [=event-level report=] |report| of the [=event-level report cache=]:
+    1. If |report|'s [=event-level report/source identifier=] is not equal to |sourceId|, [=iteration/continue=].
+    1. If |report|'s [=event-level report/trigger time=] is less than |minTriggerTime|, [=iteration/continue=].
+    1. [=set/Remove=] |report| from the [=event-level report cache=].
+    1. [=list/Remove=] the [=attribution rate-limit records=] |entry| from the [=attribution rate-limit cache=] if |entry|'s [=attribution rate-limit record/entity ID=] is equal to |report|'s [=event-level report/report ID=].
```
apasel422, line 3196: Should this say `Remove all ... where ...` instead of `Remove the ... if ...` to avoid making any assumption about the contents of the rate limit cache?

apasel422 (changes requested):

index.bs
feifeiji89: good point. Done.

verbose_debugging_reports.md
```diff
@@ -49,6 +49,11 @@ The source was rejected due to the [destination limit][].
 
 Additional fields: `limit`
 
+#### `source-max-event-states-limit`
+
+A source is rejected due to [max event states][].
+
+Additional fields: `limit`
```
apasel422, line 56:
```suggestion
Additional fields: `limit`

```

verbose_debugging_reports.md
```diff
@@ -276,4 +282,4 @@ The trigger was rejected due to an internal error.
 [trigger algorithm]: https://github.com/WICG/attribution-reporting-api/blob/main/EVENT.md#trigger-attribution-algorithm
 [trigger-state cardinality limit]: https://wicg.github.io/attribution-reporting-api/#max-trigger-state-cardinality
 [unattributed reporting origin limit]: https://github.com/WICG/attribution-reporting-api/blob/main/EVENT.md#reporting-origin-limits
-[verbose debugging reports]: https://github.com/WICG/attribution-reporting-api/blob/main/EVENT.md#verbose-debugging-reports
+[verbose debugging reports]: https://github.com/WICG/attribution-reporting-api/blob/main/EVENT.md#verbose-debugging-reports
```
apasel422, line 285: What is being changed here?

verbose_debugging_reports.md
```diff
@@ -49,6 +49,11 @@ The source was rejected due to the [destination limit][].
 
 Additional fields: `limit`
 
+#### `source-max-event-states-limit`
+
+A source is rejected due to [max event states][].
```
apasel422, line 54:
```suggestion
A source is rejected due to the [event state limit][].
```

apasel422 (approved):

index.bs
```diff
@@ -3168,6 +3171,73 @@ To <dfn>delete expired sources</dfn> given a [=moment=] |now|:
...
+    ascending order, with |a| being less than |b| if |a|,
+    <a href="https://html.spec.whatwg.org/multipage/origin.html#serialization-of-a-site">serialized</a>,
+    is less than |b|, <a href="https://html.spec.whatwg.org/multipage/origin.html#serialization-of-a-site">serialized</a>.
+1. [=set/iterate|For each=] |destination| in |pendingSource|'s [=attribution source/attribution destinations=]:
+    1. [=Remove sources with unselected attribution scopes for destination=] with |destination| and |pendingSource|.
+
+To <dfn>remove or update sources with incompatible attribution scope fields</dfn> given an [=attribution source=] |pendingSource|:
+1. Let |matchingSources| be the result of running [=find sources with common destinations and reporting origin=] with |pendingSource|.
+1. [=list/iterate|For each=] |source| of |matchingSources|:
+    1. If the result of running [=check if source has incompatible attribution scope fields=] with |source| and |pendingSource| is false, [iteration/continue=].
```
linnan-github, line 3233:
```suggestion
    1. If the result of running [=check if source has incompatible attribution scope fields=] with |source| and |pendingSource| is false, [=iteration/continue=].
```

index.bs
```diff
@@ -3168,6 +3171,73 @@ To <dfn>delete expired sources</dfn> given a [=moment=] |now|:
...
+1. [=set/iterate|For each=] |source| of the user agent's [=attribution source cache=]:
+    1. Let |commonDestinations| be the [=set/intersection=] of |source|'s [=attribution source/attribution destinations=] and |pendingSource|'s [=attribution source/attribution destinations=].
+    1. If |commonDestinations| [=set/is empty=], [=iteration/continue=].
+    1. If |source|'s [=attribution source/reporting origin=] and |pendingSource|'s [=attribution source/reporting origin=] are not [=same origin=], [=iteration/continue=].
+    1. [=list/Append=] |source| to |matchingSources|.
+1. Return |matchingSources|.
+
+To <dfn>check if source has incompatible attribution scope fields</dfn> given an [=attribution source=] |source| and an [=attribution source=] |pendingSource|:
+1. If |pendingSource|'s [=attribution source/attribution scopes=] is null:
+    1. Return true if |source|'s [=attribution source/attribution scopes=] is null, false otherwise.
```
linnan-github, line 3185: Should the returning result be opposite?

index.bs
```diff
@@ -3168,6 +3171,73 @@ To <dfn>delete expired sources</dfn> given a [=moment=] |now|:
...
+    <a href="https://html.spec.whatwg.org/multipage/origin.html#serialization-of-a-site">serialized</a>,
+    is less than |b|, <a href="https://html.spec.whatwg.org/multipage/origin.html#serialization-of-a-site">serialized</a>.
+1. [=set/iterate|For each=] |destination| in |pendingSource|'s [=attribution source/attribution destinations=]:
+    1. [=Remove sources with unselected attribution scopes for destination=] with |destination| and |pendingSource|.
+
+To <dfn>remove or update sources with incompatible attribution scope fields</dfn> given an [=attribution source=] |pendingSource|:
+1. Let |matchingSources| be the result of running [=find sources with common destinations and reporting origin=] with |pendingSource|.
+1. [=list/iterate|For each=] |source| of |matchingSources|:
+    1. If the result of running [=check if source has incompatible attribution scope fields=] with |source| and |pendingSource| is false, [iteration/continue=].
+    1. If |pendingSource|'s [=attribution source/attribution scopes=] is null and |source|'s [=attribution source/attribution scopes=] is not null:
```
linnan-github, line 3234:
```suggestion
    1. If |pendingSource|'s [=attribution source/attribution scopes=] is null:
```

verbose_debugging_reports.md
feifeiji89: no idea, git error?

index.bs
```diff
@@ -3196,6 +3266,10 @@ To <dfn>process an attribution source</dfn> given an [=attribution source=] |sou
 1. Set |source|'s [=attribution source/number of event-level reports=] to 0 if
     |source|'s [=attribution source/randomized response=] is null,
     [=attribution source/randomized response=]'s [=list/size=] otherwise.
+1. Let |states| be the [=obtain a set of possible trigger states|number of possible trigger states=] obtained with |randomizedResponseConfig|.
```
linnan-github, line 3269: These can be moved earlier, before setting source's randomized response.

linnan-github (changes requested):

index.bs
feifeiji89: Nice catch. Done.

index.bs
```diff
@@ -3266,7 +3267,6 @@ To <dfn>process an attribution source</dfn> given an [=attribution source=] |sou
 1. Set |source|'s [=attribution source/number of event-level reports=] to 0 if
     |source|'s [=attribution source/randomized response=] is null,
     [=attribution source/randomized response=]'s [=list/size=] otherwise.
-1. Let |states| be the [=obtain a set of possible trigger states|number of possible trigger states=] obtained with |randomizedResponseConfig|.
 1. If |sourceType| is "<code>[=source type/event=]</code>" and |states| is greater than |source|'s [=attribution source/attribution scopes=]'s [=attribution scopes/max event states=]:
```
linnan-github, line 3270: This step can be moved above line 3263.

index.bs
johnivdel: I think the confusing bit here is that this is checking a negation, I think typically we'd want the algorithm to be:

"check if source has compatible scope fields", as that is more readable

index.bs
```diff
@@ -3248,6 +3248,7 @@ To <dfn>process an attribution source</dfn> given an [=attribution source=] |sou
     : [=randomized response output configuration/trigger specs=]
     :: |source|'s [=trigger specs=]
 
+1. Let |states| be the [=obtain a set of possible trigger states|number of possible trigger states=] obtained with |randomizedResponseConfig|.
```
linnan-github, line 3251: Optionally, we can pass |states| to channel capacity computation and randomized response algorithm to avoid duplicate computation on the number of possible trigger states. Otherwise, this should be moved close to where it is used.

feifeiji89: Yes, These changes are covered in https://github.com/WICG/attribution-reporting-api/pull/1364/files

We can resolve the merge conflicts after one the PRs is submitted.

linnan-github: Ok, for now, let's just move it down to where it's used.

index.bs
feifeiji89: Done.

index.bs
linnan-github: @feifeiji89 Please add the limit to the report.

index.bs
feifeiji89: Good point. Done.

index.bs
feifeiji89: Done.

index.bs
```diff
@@ -3196,6 +3268,10 @@ To <dfn>process an attribution source</dfn> given an [=attribution source=] |sou
 1. Set |source|'s [=attribution source/number of event-level reports=] to 0 if
     |source|'s [=attribution source/randomized response=] is null,
     [=attribution source/randomized response=]'s [=list/size=] otherwise.
+1. Let |states| be the [=obtain a set of possible trigger states|number of possible trigger states=] obtained with |randomizedResponseConfig|.
+1. If |sourceType| is "<code>[=source type/event=]</code>" and |states| is greater than |source|'s [=attribution source/attribution scopes=]'s [=attribution scopes/max event states=]:
```
linnan-github, line 3272: This should only be checked when scopes is not null?

feifeiji89: Nice catch. Done.

verbose_debugging_reports.md
```diff
@@ -136,12 +142,12 @@ Additionally:
   will also contain a string-typed `limit` field.
 * If the trigger was attributed to a source, then the `body` will also contain
   the following fields:
-   * `source_event_id`: The source registration's `source_event_id`.
-   * `source_site`: The top-level site on which the source registration
-      occurred.
-   * `source_debug_key`: The source registration's `debug_key`, but omitted if
-     the source registration did not contain a valid `debug_key` or
-     [cookie-based debugging][] was prohibited.
+  * `source_event_id`: The source registration's `source_event_id`.
```
linnan-github, line 145: Is this change necessary?

index.bs
```diff
@@ -3113,6 +3116,8 @@ and a [=boolean=] |isNoised|:
         : "<code>[=source debug data type/source-reporting-origin-per-site-limit=]</code>"
         :: [=map/Set=] |body|["`limit`"] to the user agent's [=max source reporting origins per source reporting site=],
             [=serialize an integer|serialized=].
+        : "<code>[=source debug data type/source-max-event-states-limit=]</code>"
+        :: [=map/Set=] |body|["`limit`"] to the |source|'s [=attribution source/attribution scopes=]'s [=attribution scopes/max event states=].
```
linnan-github, line 3120: Please add assertion that scopes is not null before accessing it.

verbose_debugging_reports.md
feifeiji89: Reverted

index.bs
```diff
@@ -3113,6 +3116,10 @@ and a [=boolean=] |isNoised|:
         : "<code>[=source debug data type/source-reporting-origin-per-site-limit=]</code>"
         :: [=map/Set=] |body|["`limit`"] to the user agent's [=max source reporting origins per source reporting site=],
             [=serialize an integer|serialized=].
+        : "<code>[=source debug data type/source-max-event-states-limit=]</code>"
+        ::
+            1. [=Assert=]: |source|'s [=attribution source/attribution scopes=] is not null.
+            1. [=map/Set=] |body|["`limit`"] to the |source|'s [=attribution source/attribution scopes=]'s [=attribution scopes/max event states=].
```
linnan-github, line 3122:
```suggestion
            1. [=map/Set=] |body|["`limit`"] to |source|'s [=attribution source/attribution scopes=]'s [=attribution scopes/max event states=].
```

linnan-github (approved):

PR #1370, opened 2024-07-15
Add command-line tool for validating web source and trigger headers

apasel422:


giladbarkan-github (approved):
Should we add something to documentation with some compile and running example?

apasel422:
> Should we add something to documentation with some compile and running example?

Done.

ts/src/header-validator/main.ts
```diff
@@ -0,0 +1,127 @@
...
+          "Parses an Attribution-Reporting-Register-Source or Attribution-Reporting-Register-Trigger header using Chromium's vendor-specific values.",
+          'Prints errors, warnings, notes, and the effective value (i.e. with defaults populated) to stdout in JSON format.',
+          'Exits with 1 if there were parsing errors, 0 otherwise.',
+        ].join('\n\n'),
+      },
+    ],
+  }
+)
+
+if ((options.file === undefined) === (options.input === undefined)) {
```
agarant, line 83: optional, the logic could be wrapped in a self-executing function.

agarant (approved):

PR #1374, opened 2024-07-16
Expose common validation/serialization interface for each header

apasel422:
This makes it easier to guarantee similar operation between the web validator and the command-line one while eliminating redundant logic.

We separate option-construction for the JSON validators from context creation, which is an internal detail of their operation, which enables further encapsulation.

### DIFF
--- a/ts/src/header-validator/index.ts
+++ b/ts/src/header-validator/index.ts
@@ -1,13 +1,10 @@
 import { SourceType, parseSourceType } from '../source-type'
-import { ValidationResult } from './context'
 import * as vsv from '../vendor-specific-values'
-import { Maybe } from './maybe'
 import { makeLi } from './issue-utils'
-import { validateSource, validateTrigger } from './validate-json'
-import { serializeEligible, validateEligible } from './validate-eligible'
-import { serializeOsRegistration, validateOsRegistration } from './validate-os'
-import { serializeInfo, validateInfo } from './validate-info'
-import { serializeSource, serializeTrigger } from './to-json'
+import * as eligible from './validate-eligible'
+import * as info from './validate-info'
+import * as os from './validate-os'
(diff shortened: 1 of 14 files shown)


ts/src/header-validator/index.ts
```diff
@@ -1,13 +1,10 @@
...
-import { Maybe } from './maybe'
 import { makeLi } from './issue-utils'
-import { validateSource, validateTrigger } from './validate-json'
-import { serializeEligible, validateEligible } from './validate-eligible'
-import { serializeOsRegistration, validateOsRegistration } from './validate-os'
-import { serializeInfo, validateInfo } from './validate-info'
-import { serializeSource, serializeTrigger } from './to-json'
+import * as eligible from './validate-eligible'
+import * as info from './validate-info'
+import * as os from './validate-os'
```
agarant, line 6: Could we have these files implement the Validator type and export a validator directly  ? Seems a bit hacky to rely on import * having the right shape.

apasel422: We could do that, though it would require the creation of an additional object without real benefit, IMO: The module is already an object that conforms to the `Validator` interface; compilation would fail if the shape was incorrect.

agarant: IMO. the benefit of the additional object is clarity. I suppose, that the name of the file ties them together. But having the extra object would make it explicit. But I leave it up to you.

agarant (approved):

ts/src/header-validator/index.ts
apasel422: I will leave it as is for now, but I'll be doing some additional refactoring to split up the source/trigger stuff, since `validate-json.ts` is quite large, so it may make sense to refine this there.

PR #1375, opened 2024-07-17
Add max-trigger state cardinality to parameter list

apasel422:


ThomasQuesadilla (approved):
LGTM, let's remember to update the value link if we move forward with moving the config variable.

PR #1380, opened 2024-07-25
Set empty eligible header when there is no attribution support

### LINKED ISSUE #1368: Consider omitting Attribution Reporting request headers when there is none attribution support
Currently when there is none attribution support (neither web nor os), the `Attribution-Reporting-Eligible` header may still be sent and `Attribution-Reporting-Support` header is also sent as empty. We may consider omitting both to reduce bandwidth and a better indicator that the request is not eligible for Attribution Reporting.

linnan-github:
Fixes #1368

When there is no attribution support, set Attribution-Report-Eligible header to empty string to tell the reporting origin that the request is eligible for nothing. We send the empty headers instead of omit the headers as the absence of that header in some contexts still corresponds to the request being trigger-eligible.

***
<a href="https://pr-preview.s3.amazonaws.com/linnan-github/conversion-measurement-api/pull/1380.html" title="Last updated on Jul 31, 2024, 1:54 AM UTC (3567b17)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/attribution-reporting-api/1380/99e3b1a...linnan-github:3567b17.html" title="Last updated on Jul 31, 2024, 1:54 AM UTC (3567b17)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -389,6 +389,12 @@ and its <dfn lt="eligible key">allowed keys</dfn> are:
 
 </dl>
 
+"<code><dfn>Attribution-Reporting-Support</dfn></code>" is a
+[=structured header/dictionary|Dictionary Structured Header=]
+set on a [=request=] that indicates which registrars, if
+any, the corresponding [=response=] can use. Its values are not specified and
+its allowed keys are the [=registrars=].
+
 To <dfn>obtain a dictionary structured header value</dfn> given a [=list=] of [=strings=] |keys| and
 a [=set=] of [=strings=] |allowedKeys|:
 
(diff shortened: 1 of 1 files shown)


apasel422 (approved):

index.bs
```diff
@@ -440,25 +446,30 @@ To <dfn>set Attribution Reporting headers</dfn> given a [=request=] |request|:
...
+        :: [=list/Append=] "<code>[=eligible key/event-source=]</code>" to |keys|.
+        : "<code>[=eligibility/navigation-source=]</code>"
+        :: [=list/Append=] "<code>[=eligible key/navigation-source=]</code>" to
+            |keys|.
+        : "<code>[=eligibility/trigger=]</code>"
+        :: [=list/Append=] "<code>[=eligible key/trigger=]</code>" to |keys|.
+        : "<code>[=eligibility/event-source-or-trigger=]</code>"
+        :: [=list/Append=] "<code>[=eligible key/event-source=]</code>" and
+            "<code>[=eligible key/trigger=]</code>" to |keys|.
+1. Let |eligibleDict| be the result of [=obtaining a dictionary structured header value=] with
```
johnivdel, line 465: I would prefer a slightly different approach here, where instead of modifying the lines above, we simple insert a step like:

1. If |supportedRegistrars| [=list/is empty=], [=list/empty|clear=] |keys|.

I know the impl might behave differently for such an algorithm, but I think this better shows the intention that if we don't have a registrar we want to clear out any eligibilities.

wdyt?

johnivdel (approved):
LGTM with nit

index.bs
linnan-github: Sounds good, updated, thanks!

PR #1390, opened 2024-08-09
Reorganize verbose debugging reports document

### LINKED ISSUE #1389: Non valid reports of type "trigger-unknown-error"
100% of our “trigger-unknown-error” reports received over the past month are non valid reports (the body messages of this report type do not include the mandatory source attributes in addition to the parameter “limit”), not respecting the [documentation](https://github.com/WICG/attribution-reporting-api/blob/main/verbose_debugging_reports.md#report-data).

- I wanted to see if you are aware of the issue, maybe the documentation needs some update.

apasel422:
- Make it clear which fields are always, sometimes, and never present
- Use consistent terminology and descriptions

Fixes #1389

### DIFF
--- a/verbose_debugging_reports.md
+++ b/verbose_debugging_reports.md
@@ -1,120 +1,12 @@
 Verbose Debugging Reports
 =========================
 
-This document is a collection of [verbose debugging
-reports](https://github.com/WICG/attribution-reporting-api/blob/main/EVENT.md#verbose-debugging-reports)
-that are supported.
+This document describes the format of [verbose debugging reports][].
 
-### Source debugging reports
+The report data is included in the request body as a JSON list of dictionaries.
+Each dictionary has a string field `type` and a dictionary field `body`, e.g.:
 
-Here are the debugging reports supported for [attribution source
-registrations](https://github.com/WICG/attribution-reporting-api/blob/main/EVENT.md#registering-attribution-sources).
-
-#### `source-destination-limit`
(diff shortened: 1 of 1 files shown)


verbose_debugging_reports.md
```diff
@@ -1,175 +1,282 @@
...
 #### `source-noised`
-[Noise](https://github.com/WICG/attribution-reporting-api/blob/main/EVENT.md#data-limits-and-noise) is applied to a source event.
+
+The source was successfully registered, but it will not be attributable by any
+subsequent trigger because [noise][] has been applied.
+
+The `body` may also include a `source_destination_limit` field if the
+[destination limit][] was exceeded.
+
+Issue: It is unclear where in the [specification][] the
```
linnan-github, line 60: It's added when `source-destination-limit-replaced` is included in the data types to report, which happens when the source destination limit is hit and deleting old sources.

apasel422: Would you mind providing a link to the relevant step? IIUC [this](https://wicg.github.io/attribution-reporting-api/#ref-for-source-debug-data-type-source-noised) is the only step that produces a `source-noised` value.

apasel422: Ah. I missed the fact that we are modifying `body` in place by looping over multiple data types. I think that's fairly confusing and could do with some assertions, at the very least, if not a general refactor to make it more clear what's going on.

linnan-github (approved):

PR #1395, opened 2024-08-15
Omit source_registration_time field while excluding source registration time

linnan-github:
This is compatible with currently supported aggregation service versions.

***
<a href="https://pr-preview.s3.amazonaws.com/linnan-github/conversion-measurement-api/pull/1395.html" title="Last updated on Aug 15, 2024, 6:35 PM UTC (0bd7aa8)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/attribution-reporting-api/1395/cd5c617...linnan-github:0bd7aa8.html" title="Last updated on Aug 15, 2024, 6:35 PM UTC (0bd7aa8)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -4622,16 +4622,9 @@ of running the following steps:
 
 1. If |report|'s [=aggregatable report/debug mode=] is <strong>enabled</strong>,
     [=map/set=] |sharedInfo|["`debug_mode`"] to "`enabled`".
-1. If |report| is an [=aggregatable attribution report=]:
-    1. If |report|'s [=aggregatable attribution report/source registration time configuration=] is:
-        <dl class="switch">
-        : "<code>[=aggregatable source registration time configuration/include=]</code>"</dt>
-        :: [=map/Set=] |sharedInfo|["`source_registration_time`"] to the result of [=obtaining rounded source time=]
-            with |report|'s [=aggregatable attribution report/source time=], [=serialize an integer|serialized=].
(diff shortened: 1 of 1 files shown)


apasel422:
> This is compatible with currently supported aggregation service versions.

I believe this to be true, but do we need to follow any kind of [versioning](https://wicg.github.io/attribution-reporting-api/#aggregatable-report-shared-info) requirements for the reports themselves?

linnan-github:
> > This is compatible with currently supported aggregation service versions.
> 
> I believe this to be true, but do we need to follow any kind of [versioning](https://wicg.github.io/attribution-reporting-api/#aggregatable-report-shared-info) requirements for the reports themselves?

The version "1.0" supports this, so there's no change needed.

apasel422:
> The version "1.0" supports this, so there's no change needed.

Is that specified somewhere? If so, please link to it.

linnan-github:
> > The version "1.0" supports this, so there's no change needed.
> 
> Is that specified somewhere? If so, please link to it.

I don't think there's a link for this yet.

apasel422:
> I don't think there's a link for this yet. We also omit the field for aggregatable debug reports, so there should not be an issue.

In that case, I'd suggest waiting to merge this until there's public documentation of the behavior here.

preethiraghavan1:
Aggregation Service assumes this field "source_registration_time" as optional, so omitting this is ok. This has always been the case, so there is no need for any version bump on the report for this.

apasel422 (approved):

PR #1429, opened 2024-09-15
Use bit search for max epsilon

giladbarkan-github:
Bit search can yield higher precision than binary search in constant time and
avoid the use of a tolerance.

### DIFF
--- a/ts/src/flexible-event/privacy.test.ts
+++ b/ts/src/flexible-event/privacy.test.ts
@@ -119,7 +119,7 @@ void test('epsilonToBoundInfoGainAndDp', async (t) => {
           numStates,
           infoGainUpper,
           epsilonUpper,
-          0.00000000000001
+          1e-14
         )
 
         const epsilonByBitSearch = epsilonToBoundInfoGainAndDp(
@@ -129,12 +129,66 @@ void test('epsilonToBoundInfoGainAndDp', async (t) => {
         )
 
         assert(epsilonByBitSearch >= epsilonByBinarySearch)
-
-        if (epsilonByBitSearch > epsilonByBinarySearch) {
-          assert(
-            maxInformationGain(numStates, epsilonByBitSearch) <= infoGainUpper
-          )
+        assert(
+          maxInformationGain(numStates, epsilonByBitSearch) <= infoGainUpper
+        )
+        if (epsilonByBitSearch < epsilonUpper) {
+          assert(maxInformationGain(numStates, epsilonByBitSearch + 1e-15)
+                 > infoGainUpper)
         }
+      })
+    )
+  )
+})
+
+const epsilonSearchTests = [
+  {
+    numStates: 5545,
+    infoGainUpper: 6.5,
+    epsilonUpper: 14,
+    expected: 9.028709123768687,
+  },
+  {
+    numStates: 2106,
+    infoGainUpper: 6.5,
+    epsilonUpper: 14,
+    expected: 8.366900276574821,
+  },
+  {
+    numStates: 16036,
+    infoGainUpper: 6.5,
+    epsilonUpper: 14,
+    expected: 9.829279343808693,
+  },
+  {
+    numStates: 84121,
+    infoGainUpper: 11.5,
+    epsilonUpper: 14,
+    expected: 12.45087042924698,
+  },
+  {
+    numStates: 24895,
+    infoGainUpper: 11.5,
+    epsilonUpper: 14,
+    expected: 11.723490703852157,
+  },
+  {
+    numStates: 3648,
+    infoGainUpper: 11.5,
+    epsilonUpper: 14,
+    expected: 12.233993328032184,
+  },
+]
+
+void test('epsilonSearch', async (t) => {
+  await Promise.all(
+    epsilonSearchTests.map((tc) =>
+      t.test(`${tc.numStates}, ${tc.infoGainUpper}, ${tc.epsilonUpper}`, () => {
+        assert.deepStrictEqual(tc.expected, epsilonToBoundInfoGainAndDp(
+          tc.numStates,
+          tc.infoGainUpper,
+          tc.epsilonUpper
+        ))
       })
     )
   )
--- a/ts/src/flexible-event/privacy.test.ts
+++ b/ts/src/flexible-event/privacy.test.ts
@@ -10,7 +10,7 @@ import {
   flipProbabilityDp,
   maxInformationGain,
   epsilonToBoundInfoGainAndDp,
-  epsilonToBoundInfoGainAndDpBinarySearch
+  epsilonToBoundInfoGainAndDpBinarySearch,
 } from './privacy'
 
 const flipProbabilityTests = [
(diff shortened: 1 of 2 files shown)


ts/src/flexible-event/privacy.ts
```diff
@@ -214,3 +214,43 @@ function epsilonToBoundInfoGainAndDp(
...
+    if (epsilon > epsilonUpperBound) {
+      candidate[i] = 0
+      continue
+    }
+
+    const infoGain = maxInformationGain(numStates, epsilon)
+
+    if (infoGain > infoGainUpperBound) {
+      candidate[i] = 0
+    } else if (epsilon == epsilonUpperBound) {
```
apasel422, line 242:
```suggestion
    } else if (epsilon === epsilonUpperBound) {
```

apasel422 (changes requested):

ts/src/flexible-event/privacy.ts
```diff
@@ -214,3 +214,43 @@ function epsilonToBoundInfoGainAndDp(
...
+
+// Returns the effective epsilon needed to satisfy an information gain bound
+// given a number of output states in the q-ary symmetric channel.
+export function epsilonToBoundInfoGainAndDp(
+  numStates: number,
+  infoGainUpperBound: number,
+  epsilonUpperBound: number
+): number {
+  // Calculate epsilon to double precision.
+  const candidate: number[] = new Array<number>(64).fill(0)
```
apasel422, line 226: Instead of using an array of 64 numbers, can we use typed arrays operating on a shared buffer that lets us perform bitwise ops?

```js
const buf = new ArrayBuffer(8)
const floatView = new Float64Array(buf)
const intView = new BigUint64Array(buf)
for (let byte = 0; byte < 8; ++byte) {
  for (let bit = 0n ; bit < 8n; ++bit) {
    intView[byte] |= 1n << bit
    const epsilon = floatView[0]!
    ...
    intView[byte] &= ~(1n << bit)
```

apasel422: We may need to use [`DataView`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/DataView) to guarantee endianness, since I assume we need big-endian for the search to hold.

apasel422:
```js
  const buf = new ArrayBuffer(8)
  const dataView = new DataView(buf)

  for (let bit = 0n; bit < 64n; ++bit) {
    dataView.setBigUint64(0, dataView.getBigUint64(0) | (1n << bit))

    const epsilon = dataView.getFloat64(0)
    if (epsilon > epsilonUpperBound) {
      dataView.setBigUint64(0, dataView.getBigUint64(0) & ~(1n << bit))
      continue
    }

    const infoGain = maxInformationGain(numStates, epsilon)

    if (infoGain > infoGainUpperBound) {
      dataView.setBigUint64(0, dataView.getBigUint64(0) & ~(1n << bit))
    } else if (epsilon === epsilonUpperBound) {
      return epsilon
    }
  }

  return dataView.getFloat64(0)
```

giladbarkan-github: Ha! I forgot JavaScript could offer this lower level. Will look at it tomorrow. Thanks!

giladbarkan-github: This is very helpful, thanks again!

ts/src/flexible-event/privacy.ts
giladbarkan-github: Done.

ts/src/flexible-event/privacy.test.ts
```diff
@@ -101,6 +103,43 @@ void test('maxInformationGain', async (t) => {
...
+  const infoGainUppers = [11.5, 6.5]
+  const epsilonUpper = 14
+  const numStatesRange = 100000
+  const numTests = 500
+
+  await Promise.all(
+    [...Array(numTests).keys()].map((i) =>
+      t.test(`${i}`, () => {
+        const numStates = Math.ceil(Math.random() * numStatesRange)
+        const infoGainUpper = infoGainUppers[Math.round(Math.random())] || 11.5
```
apasel422, line 116: Is the `|| 1.5` needed here? IIUC `Math.random()` returns a number in `[0, 1)`, which after rounding will be either 0 or 1, both of which are valid indices into `infoGainUppers`, meaning that the right-hand side of the `||` should never be taken.

If it's only there to satisfy TypeScript, we should instead use an assertion to indicate that we know that the value is not `undefined`:

```js
infoGainUppers[Math.round(Math.random())]!
```
If for some reason we do need to handle the case in which it's `undefined`, let's use [`??`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing) instead of `||`.

apasel422 (changes requested):
Does Android use this algorithm in its implementation? If so, could we add some analogous deterministic test cases to make it easier to verify and reproduce the behavior here?

giladbarkan-github:
> Does Android use this algorithm in its implementation? If so, could we add some analogous deterministic test cases to make it easier to verify and reproduce the behavior here?

What do you think about a test that simply validates that adding anything in the double range to epsilon would result in too high info gain (still using random tests)? This would prove correctness.

ts/src/flexible-event/privacy.test.ts
giladbarkan-github: Yes, it was a compilation error. `!` looks good, thanks!

apasel422:
> > Does Android use this algorithm in its implementation? If so, could we add some analogous deterministic test cases to make it easier to verify and reproduce the behavior here?
> 
> What do you think about adding a validation that adding anything in the double range to epsilon would result in too high info gain (still using random tests)? This would prove correctness.

Property-based testing is good, but I still think some deterministic test cases would be beneficial as well as they are more easily ported between implementations.

ts/src/flexible-event/privacy.test.ts
giladbarkan-github: Done.

giladbarkan-github:
> > > Does Android use this algorithm in its implementation? If so, could we add some analogous deterministic test cases to make it easier to verify and reproduce the behavior here?
> > 
> > 
> > What do you think about adding a validation that adding anything in the double range to epsilon would result in too high info gain (still using random tests)? This would prove correctness.
> 
> Property-based testing is good, but I still think some deterministic test cases would be beneficial as well as they are more easily ported between implementations.

Android does not use an epsilon search presently. I'll add the extra validation, log some of its results and add those as deterministic.

csharrison:
The bit search looks fine but why keep the binary search verison too?

ts/src/flexible-event/privacy.ts
```diff
@@ -214,3 +214,34 @@ function epsilonToBoundInfoGainAndDp(
     return epsilon
   }
 }
+
+// Returns the effective epsilon needed to satisfy an information gain bound
+// given a number of output states in the q-ary symmetric channel.
+export function epsilonToBoundInfoGainAndDp(
```
csharrison, line 220: Do you mind commenting briefly what this algorithm is doing at a high level. There's a lot of bit twiddling going on

giladbarkan-github: Done!

giladbarkan-github:
> The bit search looks fine but why keep the binary search verison too?

I first thought it can help validate but what Andrew termed the "property" assertion seems similarly good.

giladbarkan-github:
> > The bit search looks fine but why keep the binary search verison too?
> 
> I first thought it can help validate but what Andrew termed the "property" assertion seems similarly good.

Removed.

ts/src/flexible-event/privacy.test.ts
```diff
@@ -101,6 +102,90 @@ void test('maxInformationGain', async (t) => {
...
 })
 
+void test('epsilonToBoundInfoGainAndDp', async (t) => {
+  const infoGainUppers = [11.5, 6.5]
+  const epsilonUpper = 14
+  const numStatesRange = 100000
+  const numTests = 500
+
+  await Promise.all(
+    [...Array(numTests).keys()].map((i) =>
```
apasel422, line 112:
```suggestion
    Array(numTests).map((, i) =>
```

apasel422 (approved):

ts/src/flexible-event/privacy.test.ts
giladbarkan-github: Done.

giladbarkan-github: I reverted this because just using the map without filling the array did not seem to work. Either current code or, for example `...fill(0).map` worked.

apasel422: I forgot that [the `Array(length)` constructor](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/Array#arraylength) is strange in that regard.

There's no need to use the `[...]` syntax with `keys()`, though, since that creates an intermediate array only to immediately create a new one from it. Just do:

```js
Array(numTests).fill(0).map((, i) => t.test(...)
```

giladbarkan-github: Done.

apasel422 (approved):

csharrison (approved):

PR #1436, opened 2024-09-17
Clean up epsilonToBoundInfoGainAndDp

apasel422:
- Avoid unnecessary bit operations resulting in bigint construction
- Avoid unnecessary array construction in tests


giladbarkan-github (approved):
Nice!

PR #1457, opened 2024-10-28
Explainer change for separate rate-limits for embedded site

linnan-github:


jolynyao:
LGTM thanks @linnan-github!

EVENT.md
```diff
@@ -980,6 +980,10 @@ were reported.
 
 In order to achieve the privacy goals listed above the API has various rate limits, which can be found [here](https://github.com/WICG/attribution-reporting-api/blob/main/params/chromium-params.md) for Chromium.
 
+Note: Browsers may choose to allow separate rate limits for the embedded site
```
apasel422, line 983: Is `embedded site` standard terminology? If so, let's link to a definition of it. Otherwise, let's explain further. Same for the spec changes.

index.bs
```diff
@@ -1832,6 +1832,8 @@ To <dfn>check if cookie-based debugging is allowed</dfn> given a
 To obtain the <dfn export for=node>context origin</dfn> of a [=node=] |node|, return |node|'s [=node navigable=]'s
 [=navigable/top-level traversable=]'s [=navigable/active document=]'s [=origin=].
 
+Note: The user agent MAY return the embedded origin for specific use cases.
```
apasel422, line 1835: I don't think this belongs in a note. We can instead phrase it as:

```
1. Optionally, return the embedded origin if it exists.
2. Return |node|'s ...
```

And then explain the reasoning or what an `embedded origin` is in the note.

apasel422 (changes requested):

EVENT.md
apasel422: Also, IIUC the spec changes aren't limited to rate limits. Rather, they effectively replace the context origin in *all uses* across the entire spec.

For sources, this includes ["max pending sources per source origin"](https://wicg.github.io/attribution-reporting-api/#max-pending-sources-per-source-origin) and ["max destinations covered by unexpired sources"](https://wicg.github.io/attribution-reporting-api/#max-destinations-covered-by-unexpired-sources), which aren't explicitly considered "rate limits."

It also includes [the source site in verbose debug report bodies](https://wicg.github.io/attribution-reporting-api/#ref-for-attribution-source-source-site%E2%91%A0%E2%91%A2).

For aggregatable reports, it affects the [effective attribution destination](https://wicg.github.io/attribution-reporting-api/#aggregatable-report-effective-attribution-destination) and therefore ["max aggregatable attribution reports per attribution destination"](https://wicg.github.io/attribution-reporting-api/#max-aggregatable-attribution-reports-per-attribution-destination).

IMO we should make it explicit that the context origin is entirely replaced, not just for rate-limit purposes.

apasel422, line 983: Is there a reason the explainer uses `embedded site` but the spec uses `embedded origin`?

linnan-github: I used “rate limits” to cover those limits as well. The destination limit is also discussed in this section, so I think it’s fine. I can mention storage limit explicitly if you feel strongly.

linnan-github: That’s just because the explainer mentions “top-level site” in that section, whereas the spec change is where we get the origin.

apasel422: The aforementioned limits aren't rate limits, as they do not expire after a specific window, and they are mentioned in different parts of the spec and explainer.

linnan-github: Updated to mention storage limit, rate-limits, and destination limits explicitly.

verbose_debugging_reports.md
```diff
@@ -33,7 +33,7 @@ The `body` will contain the following fields for all source debugging reports:
   This will be a string if there was one such site, or a list of strings if
   there were multiple.
 * `source_event_id`: The source registration's `source_event_id`.
-* `source_site`: The top-level site on which the source registration occurred.
+* `source_site`: The top-level site, or the embedded site for specific use cases, on which the source registration occurred.
```
arpanah, line 36: add (for example, [AMP](https://amp.dev) pages rendered by an AMP
cache/viewer provider) here as well

index.bs
```diff
@@ -1869,8 +1869,14 @@ To <dfn>check if cookie-based debugging is allowed</dfn> given a
...
 
-To obtain the <dfn export for=node>context origin</dfn> of a [=node=] |node|, return |node|'s [=node navigable=]'s
-[=navigable/top-level traversable=]'s [=navigable/active document=]'s [=origin=].
+To obtain the <dfn export for=node>context origin</dfn> of a [=node=] |node|:
+
+1. Optionally, return the embedded origin if it exists.
+1. Return |node|'s [=node navigable=]'s [=navigable/top-level traversable=]'s
+    [=navigable/active document=]'s [=origin=].
+
+Note: The user agent MAY return the embedded origin to allow separate limits
```
arpanah, line 1878: nit: replace capital MAY

index.bs
```diff
@@ -1875,8 +1875,14 @@ To <dfn>check if cookie-based debugging is allowed</dfn> given a
 
 <h3 algorithm id="obtaining-context-origin">Obtaining context origin</h3>
 
-To obtain the <dfn export for=node>context origin</dfn> of a [=node=] |node|, return |node|'s [=node navigable=]'s
-[=navigable/top-level traversable=]'s [=navigable/active document=]'s [=origin=].
+To obtain the <dfn export for=node>context origin</dfn> of a [=node=] |node|:
+
+1. Optionally, return the embedded origin if it exists.
```
apasel422, line 1880: It'd be helpful for people reading the spec to have some definition of `embedded origin` here, either in a note or in a link.

apasel422, line 1880: Does `optionally` here mean that different implementations can choose whether they care about embedded origins and still be compliant with the specification?

apasel422 (changes requested):

index.bs
linnan-github: Yes, that is the intention to leave it defined by the user agent.

apasel422: I think we should clarify that. Otherwise, it could be interpreted as "some of the time, do this," when it really means that the implementation can choose whether it supports the notion of embedded origins, but if it does, this algorithm should be consistent.

linnan-github: It could also apply to the case "some of the time, do this". For example, if we only want to do this for sources but not triggers.

apasel422: Right, but the point is that the algorithm should be consistent under particular static circumstances. For example, it is not in the spirit of this specification for the implementation to generate a random number to decide whether to return the embedded origin or actual origin.

index.bs
```diff
@@ -1875,8 +1875,15 @@ To <dfn>check if cookie-based debugging is allowed</dfn> given a
...
-[=navigable/top-level traversable=]'s [=navigable/active document=]'s [=origin=].
+To obtain the <dfn export for=node>context origin</dfn> of a [=node=] |node|:
+
+1. If applicable for [=implementation-defined=] use cases, return the embedded origin.
+1. Otherwise, return |node|'s [=node navigable=]'s [=navigable/top-level traversable=]'s
+    [=navigable/active document=]'s [=origin=].
+
+Note: The user agent may return the embedded origin to allow separate limits
+for specific use cases. Embedded origin could be the effective top-level origin,
+e.g. the origin of the website displayed in the AMP viewer.
```
apasel422, line 1886: I still don't think someone reading this is going to understand what an embedded origin is, as it's not a standard term as far as I know. An example would definitely help, but we should also avoid referring to AMP in the spec if possible. Maybe we can give a neutral example like comparing `https://foo.example/` vs `https://foo.example/bar.example`?

index.bs
```diff
@@ -1875,8 +1875,16 @@ To <dfn>check if cookie-based debugging is allowed</dfn> given a
...
-To obtain the <dfn export for=node>context origin</dfn> of a [=node=] |node|, return |node|'s [=node navigable=]'s
-[=navigable/top-level traversable=]'s [=navigable/active document=]'s [=origin=].
+To obtain the <dfn export for=node>context origin</dfn> of a [=node=] |node|:
+
+1. If applicable for [=implementation-defined=] use cases, return the embedded origin.
+1. Otherwise, return |node|'s [=node navigable=]'s [=navigable/top-level traversable=]'s
+    [=navigable/active document=]'s [=origin=].
+
+Note: The user agent may return the embedded origin to allow separate limits
+for specific use cases. Embedded origin could be the effective top-level origin.
```
apasel422, line 1885:
```suggestion
for specific use cases. The embedded origin could be the effective top-level origin.
```

index.bs
```diff
@@ -1875,8 +1875,16 @@ To <dfn>check if cookie-based debugging is allowed</dfn> given a
...
+For example, https://foo.example/bar.example is embedded within https://foo.example,
+where https://foo.example is the top-level origin and https://bar.example is the embedded origin.
```
apasel422, line 1887:
```suggestion
For example, a URL like `https://foo.example/bar.example` could be interpreted as embedding the origin `https://bar.example`.
```

apasel422 (changes requested):

PR #1477, opened 2024-12-10
Use more precise warning when public suffix is used as a destination

apasel422:
For example, previously https://com would be reported as the warning "URL components other than site (https://null) will be ignored".


linnan-github (commented):
Actually if https://com is allowed by Chromium, should the header validator align?

apasel422:
> Actually if https://com is allowed by Chromium, should the header validator align?

Updated to instead provide a better warning, rather than an error.

linnan-github (approved):

PR #1478, opened 2024-12-11
Update No Duplicate Rule to signify that it is Shared Id that can't appear in more than one job, not the reports.

preethiraghavan1:
With contribution filtering implementation done, the filtering ID is included in the Shared ID generation. Since the data is disjoint between filtering IDs, a report can appear in multiple batches as long as their Shared IDs are not the same across them. This warrants an update to the No Duplicate rule because it is not the reports that shouldn't be duplicated, but the underlying data represented by Shared ID, that shouldn't be.

### DIFF
--- a/AGGREGATION_SERVICE_TEE.md
+++ b/AGGREGATION_SERVICE_TEE.md
@@ -96,6 +96,7 @@ throughout this proposal.
 * _Coordinator:_ an entity responsible for key management and aggregatable report
   accounting. The coordinator maintains a list of hashes of approved aggregation
   service configurations and configures access to decryption keys.
+* _Shared ID:_ A unique identifier assigned to a group of reports in combination with [filtering IDs](https://github.com/patcg-individual-drafts/private-aggregation-api/blob/main/flexible_filtering.md#proposal-filtering-id-in-the-encrypted-payload) to prevents overlap between batches of reports. This eliminates the need to track individual reports and allows for efficient privacy budget management at the group level. 
 
 ## Aggregation workflow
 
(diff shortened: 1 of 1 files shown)


AGGREGATION_SERVICE_TEE.md
```diff
@@ -96,6 +96,7 @@ throughout this proposal.
 * _Coordinator:_ an entity responsible for key management and aggregatable report
   accounting. The coordinator maintains a list of hashes of approved aggregation
   service configurations and configures access to decryption keys.
+* _Shared ID:_ A unique identifier assigned to a group of reports in combination with [filtering IDs](https://github.com/patcg-individual-drafts/private-aggregation-api/blob/main/flexible_filtering.md#proposal-filtering-id-in-the-encrypted-payload) to prevents overlap between batches of reports. This eliminates the need to track individual reports and allows for efficient privacy budget management at the group level.
```
apasel422, line 99:
```suggestion
* _Shared ID:_ A unique identifier assigned to a group of reports in combination with [filtering IDs](https://github.com/patcg-individual-drafts/private-aggregation-api/blob/main/flexible_filtering.md#proposal-filtering-id-in-the-encrypted-payload) to prevent overlap between batches of reports. This eliminates the need to track individual reports and allows for efficient privacy budget management at the group level. 
```

apasel422 (changes requested):

linnan-github (approved):

AGGREGATION_SERVICE_TEE.md
```diff
@@ -243,7 +244,7 @@ single aggregation batch (as duplicates) or in multiple batches. Because
 of this, the aggregation service enforces a "no duplicates" rule:
 
 * No aggregatable report can appear more than once within a batch. 
-* No aggregatable report can appear in more than one batch or contribute
+* No Shared ID can appear in more than one batch or contribute
```
arpanah, line 247: What does 'Or contribute' mean in this context?

arpanah (approved):

petro353:
> With contribution filtering implementation done, the filtering ID is included in the Shared ID generation. Since the data is disjoint between filtering IDs, a report can appear in multiple batches as long as their Shared IDs are not the same across them. This warrants an update to the No Duplicate rule because it is not the reports that shouldn't be duplicated, but the underlying data represented by Shared ID, that shouldn't be.

